PR #829, opened 2024-04-16
Move license script into scripts directory

Mike-CZ:


Mike-CZ:
> the script updates all files in its directory and subdirectories. If the script is executed in the `script` directory, it will miss almost all files in the repo.
> 
> it cannot be executed as `./script/add_license_header.sh` because it will not find the `license_header.txt`

True, have overlooked that. Added path resolution so it can be called from anywhere. Also noticed, that the condition is failing and the header is beeing re-added. Will fix in following PR.

kjezek (approved):

HerbertJordan (approved):

PR #831, opened 2024-04-18
empties errors for re-used stateDB

### LINKED ISSUE #821: The test TestDatabase_BulkLoadProducesBlocks in the Carmen package is flaky
A flaky test in the `carmen` package caused a [stress test](https://jenkins.fantom.network/job/Carmen/job/Stress-Tests/40/console) on Jenkins to fail.

To reproduce the issue, run
```
$ go test ./carmen -count 10 --timeout 1000000s
```

The error produced looks like this:
```
--- FAIL: TestDatabase_BulkLoadProducesBlocks (0.23s)
    database_test.go:412: unexpected error during query evaluation: failed to compute hash: injectedErr
        failed to compute hash: injectedErr
        failed to compute hash: injectedErr
        failed to compute hash: injectedErr
--- FAIL: TestDatabase_CloseDB_Unfinished_Queries (0.18s)
    database_test.go:810: cannot close query: failed to compute hash: injectedErr
--- FAIL: TestDatabase_AddBlock_QueryBlock_Parallel (0.91s)
    database_test.go:1008: cannot query block: failed to compute hash: injectedErr
signal: killed
FAIL	github.com/Fantom-foundation/Carmen/go/carmen	168.419s
FAIL
```

kjezek:
This PR fixes #821 

The key part is that it clears the list of errors of re-used `stateDB`. Before this fix, errors accumulated in one test propagated to other tests as the `stateDB` instances are re-used from a pool.

Furthermore, a few places were extended with closing the database in tests (it did not relate to the fixed bug).

### DIFF
--- a/go/state/state_db.go
+++ b/go/state/state_db.go
@@ -1330,6 +1330,7 @@ func (s *stateDB) resetState(state State) {
 	s.ResetBlockContext()
 	s.storedDataCache.Clear()
 	s.reincarnation = map[common.Address]uint64{}
+	s.errors = s.errors[0:0]
 	s.state = state
 }
 
--- a/go/state/state_db_test.go
+++ b/go/state/state_db_test.go
@@ -3269,6 +3269,94 @@ func TestStateDB_StateDBCanNotEndABlockIfCommitIsNotAllowed(t *testing.T) {
 	}
 }
 
+func TestStateDB_NewCopy_MustBeEmpty(t *testing.T) {
+	injectedErr := fmt.Errorf("injectedErr")
+	ctrl := gomock.NewController(t)
+	state := NewMockState(ctrl)
+	gomock.InOrder(
+		state.EXPECT().Exists(gomock.Any()),
+		state.EXPECT().Exists(gomock.Any()).Return(true, nil),
+		state.EXPECT().Exists(gomock.Any()).Return(false, injectedErr),
+		state.EXPECT().Exists(gomock.Any()),
+	)
+	state.EXPECT().Check().Times(2)
+
+	checkEmpty := func(db NonCommittableStateDB, expectEmpty bool) {
+		inst := db.(*nonCommittableStateDB)
+		if got, want := len(inst.accounts) == 0, expectEmpty; got != want {
+			t.Errorf("empty check fails: got %v", got)
+		}
+		if got, want := len(inst.nonces) == 0, expectEmpty; got != want {
+			t.Errorf("empty check fails: got %v", got)
+		}
+		if got, want := inst.data.Size() == 0, expectEmpty; got != want {
+			t.Errorf("empty check fails: got %v", got)
+		}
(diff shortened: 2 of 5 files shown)


HerbertJordan (approved):
Great find, thanks for fixing that!

As I understand, only the change in `go/state/state_db.go` is the actual fix. I would suggest to put this into its own PR and do the cleanups in the tests in an independent PR. If we would encounter problems in the future, this makes it easier to understand the changes in these PRs.

kjezek:
ok, I put all other changes to #832 as they are related to that PR

HerbertJordan (dismissed):

go/state/state_db.go
```diff
@@ -1330,6 +1330,7 @@ func (s *stateDB) resetState(state State) {
 	s.ResetBlockContext()
 	s.storedDataCache.Clear()
 	s.reincarnation = map[common.Address]uint64{}
+	s.errors = s.errors[0:0]
```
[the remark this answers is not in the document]
kjezek: ok, I added a test. I found out that there is no test that verifies the re-used `stateDB` is reset, so I tested it fully

HerbertJordan (approved):

go/state/state_db_test.go
```diff
@@ -3269,6 +3269,94 @@ func TestStateDB_StateDBCanNotEndABlockIfCommitIsNotAllowed(t *testing.T) {
...
+	checkEmpty(db, false)
+
+	if err := db.Check(); !errors.Is(err, injectedErr) {
+		t.Errorf("db should fail")
+	}
+
+	db.EndTransaction()
+
+	// re-cycle the same state
+	db.(*nonCommittableStateDB).resetState(state)
```
petr-hanzl, line 3352: How about we create a private factory func that returns a pointer to the object `nonCommittableStateDB` instead of the interface `NonCommittableStateDB` and use it here to avoid casting? Assuming we call the private factory inside the public one.

go/state/state_db_test.go
```diff
@@ -3269,6 +3269,94 @@ func TestStateDB_StateDBCanNotEndABlockIfCommitIsNotAllowed(t *testing.T) {
...
+	db.SetCode(address1, make([]byte, 10))
+	db.SetNonce(address1, 0x2)
+	db.CreateAccount(common.Address{0xC}) // triggers injected error
+	db.AddAddressToAccessList(common.Address{0xD})
+
+	checkEmpty(db, false)
+
+	if err := db.Check(); !errors.Is(err, injectedErr) {
+		t.Errorf("db should fail")
+	}
```
petr-hanzl, line 3347: I think you should check the error and the error type in a separate condition.

petr-hanzl (approved):

go/state/state_db_test.go
HerbertJordan: Fixed

PR #832, opened 2024-04-18
fix memory leak in tests

### LINKED ISSUE #833: Out-of-Memory when running tests of facade many times
Running 
```
 go test ./carmen -count 100000 -memprofile mem_10000_new.prof   --parallel 1  --timeout 24h
```

causes out of memory. The number of loops to run the test may be smaller, the error occurs quickly:

Memory profile shows only that the node cache is dominating each test: 
image

Attempt was made just to reduce the cache size: [PR draft](https://github.com/Fantom-foundation/Carmen/pull/832)
but it did not solve the issue.

kjezek:
this PR removes memory leak in tests caused by failing `Close()` of database in tests. 

furthermore it reduces size of caches. 

It fixes #833

### DIFF
--- a/go/carmen/carmen_test.go
+++ b/go/carmen/carmen_test.go
@@ -32,8 +32,14 @@ var testNonArchiveConfig = Configuration{
 	Archive: Archive(state.NoArchive),
 }
 
+var testProperties = Properties{
+	LiveDBCache:  "2000",
+	ArchiveCache: "2000",
+	StorageCache: "100",
+}
+
 func TestCarmen_DatabaseLiveCycle(t *testing.T) {
-	db, err := OpenDatabase(t.TempDir(), testConfig, nil)
+	db, err := openTestDatabase(t)
 	if err != nil {
 		t.Fatalf("failed to open database: %v", err)
 	}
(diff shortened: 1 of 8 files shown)


go/carmen/carmen_test.go
```diff
@@ -32,8 +32,14 @@ var testNonArchiveConfig = Configuration{
...
 
+var testProperties = Properties{
+	LiveDBCache:  "2000",
+	ArchiveCache: "2000",
+	StorageCache: "100",
+}
+
 func TestCarmen_DatabaseLiveCycle(t *testing.T) {
-	db, err := OpenDatabase(t.TempDir(), testConfig, nil)
+	db, err := OpenDatabase(t.TempDir(), testConfig, testProperties)
```
HerbertJordan, line 42: nit: provide a `OpenTestDatabase(t)` function?

go/carmen/property.go
```diff
@@ -27,6 +27,8 @@ const (
 	// ArchiveCache is a configuration property defining an approximate upper
 	// limit for the in-memory node-cache size of the Archive in bytes.
 	ArchiveCache = Property("ArchiveCache")
+	// StorageCache is the size of cache used for account storages.
```
HerbertJordan, line 30: Not sure an external user understands this parameter based on this line of documentation. Please improve.

HerbertJordan (dismissed):
Thanks for improving the memory footprint of our tests!

go/carmen/property.go
kjezek: ok, extended

go/carmen/carmen_test.go
kjezek: ok, yes, introduced this method

HerbertJordan (approved):

PR #835, opened 2024-04-22
Fix license script and add missing headers

Mike-CZ:
This PR fixes the license generator. Also adds missing headers for BUILD files.

### DIFF
--- a/.github/workflows/cpp-linter.yml
+++ b/.github/workflows/cpp-linter.yml
@@ -1,7 +1,7 @@
 # Copyright (c) 2024 Fantom Foundation
 # 
 # Use of this software is governed by the Business Source License included
-# in the LICENSE.TXT file and at fantom.foundation/bsl11.
+# in the LICENSE file and at fantom.foundation/bsl11.
 # 
 # Change Date: 2028-4-16
 # 
--- a/.github/workflows/cpp-linter.yml
+++ b/.github/workflows/cpp-linter.yml
@@ -6,7 +6,7 @@
 # Change Date: 2028-4-16
 # 
 # On the date above, in accordance with the Business Source License, use
-# of this software will be governed by the GNU Lesser General Public Licence v3
+# of this software will be governed by the GNU Lesser General Public Licence v3.
 
 name: cpp-linter
 
--- a/.github/workflows/cpp-tests.yml
+++ b/.github/workflows/cpp-tests.yml
@@ -6,7 +6,7 @@
 # Change Date: 2028-4-16
 # 
 # On the date above, in accordance with the Business Source License, use
-# of this software will be governed by the GNU Lesser General Public Licence v3
+# of this software will be governed by the GNU Lesser General Public Licence v3.
 
 name: cpp-tests
 
--- a/.github/workflows/cpp-tests.yml
+++ b/.github/workflows/cpp-tests.yml
@@ -1,7 +1,7 @@
 # Copyright (c) 2024 Fantom Foundation
 # 
 # Use of this software is governed by the Business Source License included
-# in the LICENSE.TXT file and at fantom.foundation/bsl11.
+# in the LICENSE file and at fantom.foundation/bsl11.
 # 
 # Change Date: 2028-4-16
 # 
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -1,7 +1,7 @@
 # Copyright (c) 2024 Fantom Foundation
 # 
 # Use of this software is governed by the Business Source License included
-# in the LICENSE.TXT file and at fantom.foundation/bsl11.
+# in the LICENSE file and at fantom.foundation/bsl11.
 # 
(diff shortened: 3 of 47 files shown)


kjezek (changes requested):
I think this change only ignores new lines in the header. When the license header is modified, it will not update the files, right? If the change is a new formatting, it will be ignored. If the text is changed it will add an extra header to each file, right?

Furthermore, the cpp formatter required the part "..., use of ... " to be at the same line.  Hence I think the added headers will break build once enabled.

Mike-CZ:
> I think this change only ignores new lines in the header. When the license header is modified, it will not update the files, right? If the change is a new formatting, it will be ignored. If the text is changed it will add an extra header to each file, right?
> 
> Furthermore, the cpp formatter required the part "..., use of ... " to be at the same line. Hence I think the added headers will break build once enabled.

This change fixes the condition in the generator script, that is always false and the header is being re-added although it is already present. Also fixes some inconsistencies in files, where the header does not match the one being added. That's all it does.

kjezek:
> > I think this change only ignores new lines in the header. When the license header is modified, it will not update the files, right? If the change is a new formatting, it will be ignored. If the text is changed it will add an extra header to each file, right?
> > Furthermore, the cpp formatter required the part "..., use of ... " to be at the same line. Hence I think the added headers will break build once enabled.
> 
> This change fixes the condition in the generator script, that is always false and the header is being re-added although it is already present. Also fixes some inconsistencies in files, where the header does not match the one being added. That's all it does. I will create separate PR with feature, that allows "updating" of license headers.
> 
> Yes, it will break the build, we need to unify this mess. But I can't include fixing cpp build into this PR, since it has nothing to do with it.

Right, but I think the condition was false because formatting of the header file did not match actual headers. This change basically ignores empty lines and new lines - if I am not wrong, it creates one long line of the header, right?  

I think we should work towards a more clever script that can also update an existing header, and use it to unify all headers that were added manually for the first iteration. 

Also I would split into two PRs, one that modifies the script and one which updates headers.

Mike-CZ:
> > > I think this change only ignores new lines in the header. When the license header is modified, it will not update the files, right? If the change is a new formatting, it will be ignored. If the text is changed it will add an extra header to each file, right?
> > > Furthermore, the cpp formatter required the part "..., use of ... " to be at the same line. Hence I think the added headers will break build once enabled.
> > 
> > 
> > This change fixes the condition in the generator script, that is always false and the header is being re-added although it is already present. Also fixes some inconsistencies in files, where the header does not match the one being added. That's all it does. I will create separate PR with feature, that allows "updating" of license headers.
> > Yes, it will break the build, we need to unify this mess. But I can't include fixing cpp build into this PR, since it has nothing to do with it.
> 
> Right, but I think the condition was false because formatting of the header file did not match actual headers. This change basically ignores empty lines and new lines - if I am not wrong, it creates one long line of the header, right?
> 
> I think we should work towards a more clever script that can also update an existing header, and use it to unify all headers that were added manually for the first iteration.
> 
> Also I would split into two PRs, one that modifies the script and one which updates headers.

The problem was, that it was comparing a string, where "\n" was not evaluated and was present as two plain chars. So it always failed, because the file contains actual new line symbol. So that's why I added evaluation via $(echo -e ...). So it is not ignoring new lines, it is exact match.

I planned to do it as next iteration, but I might do it directly.

kjezek:
> > > > I think this change only ignores new lines in the header. When the license header is modified, it will not update the files, right? If the change is a new formatting, it will be ignored. If the text is changed it will add an extra header to each file, right?
> > > > Furthermore, the cpp formatter required the part "..., use of ... " to be at the same line. Hence I think the added headers will break build once enabled.
> > > 
> > > 
> > > This change fixes the condition in the generator script, that is always false and the header is being re-added although it is already present. Also fixes some inconsistencies in files, where the header does not match the one being added. That's all it does. I will create separate PR with feature, that allows "updating" of license headers.
> > > Yes, it will break the build, we need to unify this mess. But I can't include fixing cpp build into this PR, since it has nothing to do with it.
> > 
> > 
> > Right, but I think the condition was false because formatting of the header file did not match actual headers. This change basically ignores empty lines and new lines - if I am not wrong, it creates one long line of the header, right?
> > I think we should work towards a more clever script that can also update an existing header, and use it to unify all headers that were added manually for the first iteration.
> > Also I would split into two PRs, one that modifies the script and one which updates headers.
> 
> The problem was, that it was comparing a string, where "\n" was not evaluated and was present as two plain chars. So it always failed, because the file contains actual new line symbol. So that's why I added evaluation via $(echo -e ...). So it is not ignoring new lines, it is exact match.
> 
> I planned to do it as next iteration, but I might do it directly.

Thanks for explanation. I would still suggest to split this PR into one that fixes the script and another one that modifies the headers.

If you want to work on the script in more PRs it is fine. I just think that perhaps this first update of the script is rather small, but it is ok to have it in more PRs if you prefer.

PR #836, opened 2024-04-23
Fix multiple flushing of codes

hkalina:
Everytime when MptState is being Flushed, the codes file is written to the disk, even when no code have been modified.

This affects event the time of Carmen closing, as the flush is called here several times:

image

This PR adds a codeDirty flag, which avoids these unnecessary overrides.

This also adds proper codeMutex locking into the Flush and GetMemoryFootprint.
We may want to use ReadWrite lock in the future, however until now, we just supposed there is no concurrency of Flush/GetMemoryFootprint with code writes.

### DIFF
--- a/go/database/mpt/state.go
+++ b/go/database/mpt/state.go
@@ -20,6 +20,7 @@ import (
 	"github.com/Fantom-foundation/Carmen/go/state"
 	"hash"
 	"io"
+	"maps"
 	"os"
 	"sync"
 	"unsafe"
@@ -384,7 +385,10 @@ func (s *MptState) Visit(visitor NodeVisitor) error {
 }
 
 func (s *MptState) GetCodes() (map[common.Hash][]byte, error) {
-	return s.code, nil
+	s.codeMutex.Lock()
+	res := maps.Clone(s.code)
+	s.codeMutex.Unlock()
+	return res, nil
 }
 
 // Flush codes and state trie
--- a/go/database/mpt/state.go
+++ b/go/database/mpt/state.go
@@ -136,6 +136,7 @@ type MptState struct {
 	lock      common.LockFile
 	trie      *LiveTrie
 	code      map[common.Hash][]byte
+	codeDirty bool
 	codeMutex sync.Mutex
 	codefile  string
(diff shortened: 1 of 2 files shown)


go/database/mpt/state.go
HerbertJordan, line 1: There is also a `GetCodes` function that should be replaced by
```
func (s *MptState) GetCodes() (map[common.Hash][]byte, error) {
	s.codeMutex.Lock()
	res := maps.Clone(s.code)
	s.codeMutex.Unlock()
	return res, nil
}
```

HerbertJordan (changes requested):
Thanks for adding this, looks good to me, but I can only approve this with proper unit test coverage.

At the least, there needs to be a test that tests that after this change indeed codes are only flushed once.

go/database/mpt/state.go
hkalina: Are you sure? This may significantly slow-down the genesis export, which seems to be the only usage of this.
Would not be a comment warning that the provided map must not be accessed concurrently with the state writes sufficient?

HerbertJordan: I do not think this has a high impact on the export time. The codes are not a huge contributor to the overall runtime. Also, the clone operation is a shallow-clone, only copying the map structure but not the codes in it.

go/database/mpt/state_test.go
```diff
@@ -529,6 +529,37 @@ func TestState_ForestErrorIsReportedInFlushAndClose(t *testing.T) {
...
+	}
+
+	stat2, err := os.Stat(dir + "/codes.dat")
+	if err != nil {
+		t.Errorf("failed to get modtime of codes file after second flush")
+	}
+	if stat1.ModTime() != stat2.ModTime() {
+		t.Errorf("codes written even when not dirty")
+	}
+}
```
HerbertJordan, line 561: Can you add a third flush after an actual modification to include coverage for the "mark-dirty" mechanism?

HerbertJordan (changes requested):
Looks good, just one small additional test-step request.

HerbertJordan (approved):
Thanks!

PR #837, opened 2024-04-23
Mark nodes clean when they are flushed

kjezek:
this PR marks nodes clean when they are flushed. This causes that repeated call to` Flush()` does not store nodes again and again. 

This change should boost performance as unnecessary I/O is avoided.


HerbertJordan (approved):
Very nice, also the test, thanks!

PR #838, opened 2024-04-23
Benchmark dirty trie

petr-hanzl:
This PR adds benchmark to test closing a Trie with high number of dirty nodes.

### DIFF
--- a/go/database/mpt/live_trie_test.go
+++ b/go/database/mpt/live_trie_test.go
@@ -968,8 +968,8 @@ func BenchmarkLiveTrie_WithHighNumberOfDirtyNodes_Flush(b *testing.B) {
 	// largest nodes have ~512 bytes - we need to create a cache with 18 GB cache,
 	// hence we need ~ 18*2^30/2^9 = 18*2^21 ~ 38 million
 	cacheSize := 38_000_000
-	// number of accounts need to be approximately 1/4 smaller because we store leafs and branches
-	numAccounts := cacheSize - 10_500_000
+	// number of accounts need to be ~1/3 smaller because we store leafs and branches
+	numAccounts := cacheSize - (cacheSize / 3)
 	dir := b.TempDir()
 	state, err := OpenGoFileState(dir, config, cacheSize)
 	if err != nil {
--- a/go/database/mpt/live_trie_test.go
+++ b/go/database/mpt/live_trie_test.go
@@ -960,3 +960,34 @@ func BenchmarkValueInsertionInFileTrie(b *testing.B) {
 		})
 	}
 }
+
+func BenchmarkLiveTrie_WithHighNumberOfDirtyNodes_Flush(b *testing.B) {
+	// running this test for all configurations at once takes too much time
+	// hence we use only the latest one
+	config := S5ArchiveConfig
+	// largest nodes have ~512 bytes - we need to create a cache with 18 GB cache,
+	// hence we need ~ 18*2^30/2^9 = 18*2^21 ~ 38 million
+	cacheSize := 38_000_000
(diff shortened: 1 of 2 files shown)


go/database/mpt/live_trie_test.go
```diff
@@ -960,3 +960,34 @@ func BenchmarkValueInsertionInFileTrie(b *testing.B) {
...
+	// running this test for all configurations at once takes too much time
+	// hence we use only the latest one
+	config := S5ArchiveConfig
+	// largest nodes have ~512 bytes - we need to create a cache with 18 GB cache,
+	// hence we need ~ 18*2^30/2^9 = 18*2^21 ~ 38 million
+	cacheSize := 38_000_000
+	// number of accounts need to be ~1/3 smaller because we store leafs and branches
+	numAccounts := cacheSize - (cacheSize / 3)
+	dir := b.TempDir()
+	state, err := OpenGoFileState(dir, config, cacheSize)
```
HerbertJordan, line 974: A benchmark targeting a state should not be in `live_trie_test.go` but in `state_test.go`

go/database/mpt/live_trie_test.go
```diff
@@ -960,3 +960,34 @@ func BenchmarkValueInsertionInFileTrie(b *testing.B) {
 		})
 	}
 }
+
+func BenchmarkLiveTrie_WithHighNumberOfDirtyNodes_Flush(b *testing.B) {
+	// running this test for all configurations at once takes too much time
+	// hence we use only the latest one
+	config := S5ArchiveConfig
+	// largest nodes have ~512 bytes - we need to create a cache with 18 GB cache,
```
HerbertJordan, line 968: I checked yesterday, while investigating memory issues reported by Jan: the largest nodes actually have 944 bytes.

go/database/mpt/live_trie_test.go
```diff
@@ -960,3 +960,34 @@ func BenchmarkValueInsertionInFileTrie(b *testing.B) {
...
+	// hence we need ~ 18*2^30/2^9 = 18*2^21 ~ 38 million
+	cacheSize := 38_000_000
```
HerbertJordan, line 970: As discussed offline, the problem size should be considerably reduced to make this a reasonable benchmark.

go/database/mpt/live_trie_test.go
petr-hanzl: It make sense. Grafana showed memory consumption of <30GiB. Thanks for the heads up!

go/database/mpt/live_trie_test.go
petr-hanzl: Unfortunately, I wasnt able to push this test under 30s.

go/database/mpt/state_test.go
```diff
@@ -625,3 +625,44 @@ func TestState_parseCodes_ReadFailures(t *testing.T) {
...
+			if err != nil {
+				b.Fatalf("failed to create account: %v", err)
+			}
+			err = state.SetCode(addr, []byte{byte(j)})
+			if err != nil {
+				b.Fatalf("failed to set code: %v", err)
+			}
+		}
+
+		if err = state.Flush(); err != nil {
```
kjezek, line 659: should not we call only `Flush`() in the `b.N` for-loop?  At least for the bug with slow `Flush`(), I guess we are interested in performance of `Flush` only

HerbertJordan:
This PR got superseded by #842

go/database/mpt/state_test.go
HerbertJordan: Indeed, those operations in addition to the flush are very costly, dominating the execution time. The problem got fixed in #842 .

PR #839, opened 2024-04-24
Extend license script to allow modifications

Mike-CZ:
This PR enables the license script to regenerate already present license headers with a new version stored in `license_header.txt` file.


kjezek (commented):
Is the documentation at the beginning of the file up to date? 

It still says" # However, the script does not allow at the moment for updating
# the license header."

i.e. I am confused what the script does at the moment

scripts/license/add_license_header.sh
```diff
@@ -51,28 +51,60 @@ extend_license_header() {
...
+#   comment prefix, e.g.: //, #,
 add_license_to_files() {
     local file_extension="$1"
-    local license_header="$2"
+    local prefix="$2"
+    local license_header="$(extend_license_header "$prefix")"
 
     # Get a list of all files in the project directory
-    local all_files=($(find $root_dir -type f -name "*$file_extension"))
+    mapfile -t -d $'\0' all_files < <(find "$root_dir" -type f -name "*$file_extension" -print0)
```
kjezek, line 61: what is `mapfile`? Is it available in bash, or does it need to install something extra - would be better not to need to install.

Mike-CZ: It is in bash since 4.4 version. It should be instead of the old style. See [this](https://github.com/koalaman/shellcheck/wiki/SC2207)

Mike-CZ: Switched it back because of macOS compatibility

kjezek (changes requested):
I run the script and it corrupted many files, they cannot compile

kjezek:
does it work for you? One of the issue is for instance this:

image

while original code was:

image

It seems like it replaces `'\n'` with actual new line

Mike-CZ:
Could you try it now?

kjezek (changes requested):
There is one more problem, the script does not recognise C++ headers.

image

Mike-CZ:
> There is one more problem, the script does not recognise C++ headers.
> 
> image

Yes, that's what I've discussed with you on Slack. We will have to manually delete the long-form C++ comments and re-generate them.

kjezek:
Could the script ignore the comment sign? It would solve the problem I guess.

Mike-CZ:
It's not single sign. It is a sequence starting with /* and ending */. That would bring too much complexity into the script. It is much easier just to unify it.

Mike-CZ:
Added very simple skip for C++ comments.

PR #840, opened 2024-04-24
tests how standard slice grows

kjezek:
This PR verifies that huge slices of hash roots are growing sub linearly.  

It is to avoid situation where this slice consumes already GiBs of memory as a consequence of storing minions of roots, and suddenly doubles memory consumption caused by expanded capacity of this slice. 

It basically tests only how current golang behaves, but should serve as a validation that further versions or configurations do not change.

### DIFF
--- a/go/database/mpt/archive_trie_test.go
+++ b/go/database/mpt/archive_trie_test.go
@@ -513,6 +513,52 @@ func TestArchiveTrie_Add_UpdateFailsHashing(t *testing.T) {
 	}
 }
 
+func TestArchive_RootsGrowSubLinearly(t *testing.T) {
+	for _, config := range allMptConfigs {
+		t.Run(config.Name, func(t *testing.T) {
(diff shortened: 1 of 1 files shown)


go/database/mpt/archive_trie_test.go
```diff
@@ -513,6 +513,52 @@ func TestArchiveTrie_Add_UpdateFailsHashing(t *testing.T) {
...
+			defer func() {
+				if err := archive.Close(); err != nil {
+					t.Fatalf("cannot close archive: %v", err)
+				}
+			}()
+
+			// Golang slices grow by the factor of 2 when they are small,
+			// while they grow slower when they become huge.
+			// The slice 'archive.roots' contains millions of elements
+			// printed as GiBs in memory.
```
HerbertJordan, line 532: nit: why `printed`?

HerbertJordan (dismissed):
Looks good, thanks for adding this.

HerbertJordan (approved):

PR #841, opened 2024-04-25
Update estimation of memory usage per node in cache

HerbertJordan:
This PR replaces the out-dated and hard-coded node size estimate of 512 byte per node with a slightly more sophisticated estimate. This should help making the cache-size configuration parameter more reliable.


kjezek (approved):
thanks, looks good

kjezek (dismissed):

petr-hanzl (approved):

PR #847, opened 2024-05-06
Add regenerated unified license headers

Mike-CZ:


cpp/third_party/crc32c/crc32c_config.h
HerbertJordan, line 1: i think we should not add a license header to third-party files like those, in particular if they already contain a license header.

cpp/third_party/leveldb/port.h
HerbertJordan, line 1: Same here and in other files in the `third_party` directory.

HerbertJordan (changes requested):

cpp/third_party/crc32c/crc32c_config.h
Mike-CZ: Will add option to ignore certain paths

Issue #845, opened 2024-05-03
GetArchiveBlockHeight() sometimes returns incorrect data


petr-hanzl:
`GetArchiveBlockHeight()` is flaky and sometimes returns incorrect data when running 
`./build/aida-vm-sdb substate --db-impl carmen --db-variant go-file --carmen-schema 5 --db-tmp /var/data/tmp --aida-db /var/data/aida-db --track-progress --workers=40  --archive --archive-variant s5 --archive-query-rate 2000 0 1000`
it returns that Archive is already on block `X` but when `GetArchiveState(X)` is called, it returns error `block does not exist`.

This only happens sometimes at the beginning of the run so you might need to run the command multiple times to trigger this bug.

kjezek:
fixed by #848

PR #848, opened 2024-05-06
removes check of last block from history queries

kjezek:
this PR removes check for the `lastBlock` from the archive query of the Carmen facade.

this check used to cause error: https://github.com/Fantom-foundation/Carmen/issues/845

the problem was that while the head state block is being committed, the block may be already available in the archive, before the `lastBlock` counter is incremented.  It was fixed the way that this check is removed as it  serves rather for sequencing new head state blocks. 

Another approach would be to capture this counter in the same lock as the one managing the head state, though it would unnecessarily block running parallel archive queries. I.e. such an approach was not taken.

### DIFF
--- a/go/carmen/database.go
+++ b/go/carmen/database.go
@@ -201,10 +201,6 @@ func (db *database) GetHistoricContext(block uint64) (HistoricBlockContext, erro
 		return nil, errDbClosed
 	}
 
-	if db.lastBlock < int64(block) {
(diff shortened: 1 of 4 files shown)


petr-hanzl (approved):
Checked with Aida. Works as intended! Thanks for the quick fix!

HerbertJordan (approved):
Nice find, thanks for fixing it!

PR #850, opened 2024-05-09
implements extraction of witness proofs

kjezek:
This PR implements extraction of Witness proof from database.  It contains API + implementation that allows for generating the witness proof for an address and a list of storage keys. 

It iterates the MPT from its root down the account according to the input address. Until the path exists, it keeps storing encountered nodes in the witness proof. The nodes are stored in the witness proof in pairs of their RLP encoding and hash of the encoding. 

If the input contains in addition a list of keys, the storage root of the respective account is iterated for each key from the account down to the storage Value. The nodes are returned RLP and hash encoded the same way. 

The witness proof is returned all the time even if the key or address could not be found. In this case the proof contains only nodes that could be reached.  Hence such a partial proof can be used by a client to prove  that the address/key is empty.

### DIFF
--- a/go/database/mpt/proof.go
+++ b/go/database/mpt/proof.go
@@ -12,6 +12,7 @@ package mpt
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"github.com/Fantom-foundation/Carmen/go/common"
 	"github.com/Fantom-foundation/Carmen/go/common/tribool"
@@ -44,7 +45,34 @@ type WitnessProof struct {
 // and possibly storage slots of the same account under the input storage keys.
 // This method may return an error when it occurs in the underlying database.
 func CreateWitnessProof(nodeSource NodeSource, root *NodeReference, address common.Address, keys ...common.Key) (WitnessProof, error) {
-	panic("not implemented")
+	proof := proofDb{}
+	visitor := &proofExtractionVisitor{
+		nodeSource: nodeSource,
+		proof:      proof,
+	}
+
+	var innerError error
+
+	_, err := VisitPathToAccount(nodeSource, root, address, MakeVisitor(func(node Node, info NodeInfo) VisitResponse {
+		if res := visitor.Visit(node, info); res == VisitResponseAbort {
+			return VisitResponseAbort
+		}
+		// if account reached, prove storage keys and terminate.
+		if account, ok := node.(*AccountNode); ok {
+			for _, key := range keys {
+				_, err := VisitPathToStorage(nodeSource, &account.storage, key, visitor)
+				if err != nil || visitor.err != nil {
+					innerError = errors.Join(innerError, visitor.err, err)
+					return VisitResponseAbort
+				}
+			}
+			return VisitResponseAbort
+		}
+
+		return VisitResponseContinue
+	}))
+
+	return WitnessProof{proof}, errors.Join(innerError, visitor.err, err)
 }
 
 // Add merges the input witness proof into the current witness proof.
@@ -181,6 +209,33 @@ func (p WitnessProof) AllAddressesEmpty(root common.Hash, from, to common.Addres
 // Equals returns true if the two witness proofs are equal.
 func (p WitnessProof) Equals(other WitnessProof) bool {
 	return maps.EqualFunc(p.proofDb, other.proofDb, rlpEncodedNodeEquals)
+}
+
+// proofExtractionVisitor is a visitor that visits MPT nodes and creates a witness proof.
+// It hashes and encodes the nodes and stores them into the proof database.
+type proofExtractionVisitor struct {
+	proof      proofDb
+	nodeSource NodeSource
+	err        error
+}
+
+// Visit computes RLP and hash of the visited node and puts it to the proof.
+func (p *proofExtractionVisitor) Visit(node Node, info NodeInfo) VisitResponse {
+	if info.embedded {
+		return VisitResponseAbort
+	}
+
+	data := make([]byte, 0, 1024)
+	rlp, err := encodeToRlp(node, p.nodeSource, data)
+	if err != nil {
+		p.err = err
+		return VisitResponseAbort
+	}
+	hash := common.Keccak256(rlp)
+
+	p.proof[hash] = rlp
+
+	return VisitResponseContinue
 }
 
 // String returns a string representation of the witness proof.
--- a/go/database/mpt/proof.go
+++ b/go/database/mpt/proof.go
@@ -221,7 +221,7 @@ type proofExtractionVisitor struct {
 
 // Visit computes RLP and hash of the visited node and puts it to the proof.
 func (p *proofExtractionVisitor) Visit(node Node, info NodeInfo) VisitResponse {
-	if info.embedded {
+	if info.Embedded.True() {
 		return VisitResponseAbort
 	}
 
--- a/go/database/mpt/proof_test.go
+++ b/go/database/mpt/proof_test.go
@@ -12,6 +12,7 @@ package mpt
 
 import (
 	"encoding/hex"
+	"errors"
 	"fmt"
 	"github.com/Fantom-foundation/Carmen/go/common"
 	"github.com/Fantom-foundation/Carmen/go/database/mpt/shared"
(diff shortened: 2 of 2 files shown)


go/database/mpt/proof.go
```diff
@@ -0,0 +1,197 @@
...
+
+		return VisitResponseContinue
+	}))
+
+	if err != nil || storageVisitError != nil || accountVisitor.err != nil {
+		return StorageProofs{}, errors.Join(storageVisitError, accountVisitor.err, err)
+	}
+
+	if !exists {
+		return StorageProofs{storage: make(map[common.Key]StorageProof)}, nil
```
petr-hanzl, line 193: Is creating the map for an empty proof necessary?

kjezek: it is for a more user friendly interface. If the proof is empty and user calls:  `proof.GetStorageProof(key)` and the map was not initialised, it would panic. This way it just returns an empty proof.

kjezek: hm, actually I checked that it does not panic when the map is not initialised. Is it a feature of go that getting a key from a map, which is nil does not fail?

petr-hanzl: I believe nil map does not panic unless you assign yo it.

petr-hanzl: Right. Check this test: 
```
func TestEmptyMap_Panic(t *testing.T) {
	var m map[string]string
	// no panic
	_ = m["a"]

	// panics
	m["a"] = "b"
}
```

It fails with `panic: assignment to entry in nil map`

go/database/mpt/proof.go
```diff
@@ -0,0 +1,197 @@
...
+		hasher:     p.hasher,
+		nodeSource: p.nodeSource,
+	}
+	exists, err := VisitPathToAccount(p.nodeSource, p.root, address, visitor)
+	if err != nil {
+		return AccountProof{}, err
+	}
+	if !exists {
+		return AccountProof{}, visitor.err
+	}
```
HerbertJordan, line 138: A proof that something does not exist is equally valuable. It proofs, for instance, that the balance, nonce, and code is empty.

go/database/mpt/proof.go
HerbertJordan: Yes, reading from `nil` maps is perfectly fine and should be utilized, as in Go default values should be valid states wherever possible.

HerbertJordan, line 193: If the account does not exist, a proof for its lack of existence should be returned.

go/database/mpt/proof.go
```diff
@@ -0,0 +1,197 @@
...
+// It is used to verify the proof by checking that the nodes of the MPT are present in this proof.
+// The proof is created by iterating the MPT from the root down to the leaf node following
+// the address or a storage key.
+// Each node encountered on the path is hashed and stored in the proof.
+type proof []ProofElement
+
+// nodeVisitor is a visitor that visits MPT nodes and creates a witness proof.
+// It uses the hasher and node source to hash and encode the nodes.
+// It computes RLP and hash of nodes that compose the proof.
+type nodeVisitor struct {
```
HerbertJordan, line 47: The name `nodeVisitor` is very generic and it is visible in the entire MPT package. Can you rename it to something like `proofExtractionVisitor`?

go/database/mpt/proof.go
```diff
@@ -0,0 +1,197 @@
...
+
+// GetStorageProof returns the storage proof for the input storage key.
+func (p *StorageProofs) GetStorageProof(key common.Key) StorageProof {
+	return p.storage[key]
+}
+
+// CreateAccountProof creates a witness proof for the account at the given address.
+// It iterates the MPT from the root down to the leaf node following the address.
+// Each node encountered on the path is hashed and stored in a proof.
+func (p *ProofSource) CreateAccountProof(address common.Address) (AccountProof, error) {
```
HerbertJordan, line 127: The implementation of this function could call `CreateStorageProof` with an empty key list to avoid redundant code. Wdyt?

go/database/mpt/proof_test.go
```diff
@@ -0,0 +1,462 @@
...
+		return h
+	})
+	nodeSource.EXPECT().hashAddress(gomock.Any()).AnyTimes().DoAndReturn(func(address common.Address) common.Hash {
+		var h common.Hash
+		copy(h[:], address[:])
+		return h
+	})
+}
+
+func mockNodeSource(ctrl *gomock.Controller, nodes []Node) NodeSource {
```
HerbertJordan, line 295: This name is quite generic and visible in the entire MPT package. Can you think of a more specific name describing what it is providing?

go/database/mpt/proof.go
```diff
@@ -0,0 +1,197 @@
...
+// to the leaf node while checking at every step that the encountered node is present in
+// the proof.
+type ProofSource struct {
+	nodeSource NodeSource
+	hasher     hasher
+	root       *NodeReference
+}
+
+// NewProofSource creates a new proof source with the input node source and root node.
+func NewProofSource(root *NodeReference, nodeSource NodeSource) *ProofSource {
```
HerbertJordan, line 93: Should the `ProofSource` be made an implementation detail and the function
```
func CreateStorageProofs(
        source NodeSource,
        root *NodeReference,
        address common.Address, 
        keys ...common.Key,
) (StorageProofs, error) { ... }
```
become a free function? Same for `CreateAccountProof`?

I do not think there is a utility in having the ProofSource as an independent entity. Wdyt?

go/database/mpt/proof_test.go
```diff
@@ -0,0 +1,462 @@
...
+
+	if got, want := len(proof), len(nodes); got != want {
+		t.Fatalf("unexpected proof length: got %d, want %d", got, want)
+	}
+}
+
+func TestCreateAccountProof_CannotGenerateProof(t *testing.T) {
+	ctrl := gomock.NewController(t)
+
+	nodes := []Node{&EmptyNode{}}
```
HerbertJordan, line 51: This should create a valid proof -- certifying that the account is not present.

go/database/mpt/proof_test.go
```diff
@@ -0,0 +1,462 @@
...
+		rlp, err := hasher.encode(nodes[i], nodeSource, []byte{})
+		if err != nil {
+			t.Fatalf("failed to encode node: %v", err)
+		}
+		if got, want := element.Rlp(), rlp; !bytes.Equal(got, want) {
+			t.Errorf("unexpected RLP: got %x, want %x", got, want)
+		}
+		if got, want := element.Hash(), common.Keccak256(rlp); got != want {
+			t.Errorf("unexpected hash: got %x, want %x", got, want)
+		}
```
HerbertJordan, line 40: The Proof types should eventually have a "IsValid" function which is checking this. Implementing this first, would simplify those tests.

go/database/mpt/proof_test.go
```diff
@@ -0,0 +1,462 @@
...
+			}
+		}
+
+		if got, want := len(proof.GetStorageProof(key)), len(storageNodes[i]); got != want {
+			t.Errorf("unexpected proof length: got %d, want %d", got, want)
+		}
+	}
+}
+
+func TestCreateStorageProofs_CannotGenerateProofs(t *testing.T) {
```
HerbertJordan, line 124: "CannotGenerateProofs" ... in what situation should it not be able to generate proofs?

Ideally test names have the format "Test<Subject>_<Condition><ExpectedOutcome>". Here, for instance, it could be something like
```
TestCreateStorageProofs_NonExistingStorageSlotsLeadToEmptyProofs"
```

go/database/mpt/proof_test.go
```diff
@@ -0,0 +1,462 @@
...
+		call := nodeSource.EXPECT().getViewAccess(gomock.Any()).Return(shared.MakeShared(node).GetViewHandle(), nil)
+		calls = append(calls, call)
+	}
+
+	gomock.InOrder(calls...)
+
+	return nodeSource
+}
+
+func TestCreateProofs_From_Database(t *testing.T) {
```
HerbertJordan, line 311: I think this test could be simplified by exploiting the hash and proof infrastructure. 

If you first implement proof validation utilities, this test would become much simpler since you would not have to collect the actual nodes. You could, instead, trust the validity test of the proof.

go/database/mpt/proof.go
kjezek: how should such a proof look like? Should it contain the nodes that could be reached until it reached a dead end?  Plus there will be a method `isValid()` that would say if the proof lead to an account or not, so that the client does not have to parse it to figure out?

go/database/mpt/proof.go
```diff
@@ -0,0 +1,197 @@
...
+// It contains the witness proof of the account and a list of storage proofs for the storage slots.
+// The witness proof is created by iterating the MPT from the root node
+// down to an account node following the input address.
+// Each node encountered on the path is hashed and stored in a proof.
+// Furthermore, the storage proofs are created by iterating the MPT from the account node
+// down to the leaf node following the storage key.
+// The proof contains one account and possibly multiple storage proofs.
+type StorageProofs struct {
+	accountProof AccountProof
+	storage      map[common.Key]StorageProof
```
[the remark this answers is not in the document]
kjezek: ok, got it. I had it actually in a previous version, but then change it to retain order of nodes in the proof. I guess we do not have to retain the order right? It is encoded in nodes and their links to kids

go/database/mpt/proof.go
kjezek: ok, removed

go/database/mpt/proof.go
```diff
@@ -0,0 +1,197 @@
...
+import (
+	"bytes"
+	"errors"
+	"github.com/Fantom-foundation/Carmen/go/common"
+)
+
+//go:generate mockgen -source proof.go -destination proof_mocks.go -package witness
+
+// RlpNode is an RLP encoded MPT node.
+type RlpNode []byte
```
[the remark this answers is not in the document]
kjezek: ok, used `RlpEncodedNode`

go/database/mpt/proof.go
```diff
@@ -0,0 +1,197 @@
...
+// ProofSource is a utility to create witness proofs for accounts and storage slots.
+// It has access to nodes of the MPT and a root.
+// The witness proof is created by iterating the MPT from the current root
+// down to a leaf node following the input address or  a storage key.
+// Each node encountered on the path is hashed and stored in a proof.
+// This proof represents the witness proof of the input address and storage key.
+// This proof is furthermore verified by iterating it as an MPT structure from the root down
+// to the leaf node while checking at every step that the encountered node is present in
+// the proof.
+type ProofSource struct {
```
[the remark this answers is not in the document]
kjezek: removed completly

go/database/mpt/proof.go
kjezek: done

go/database/mpt/proof.go
kjezek: done

go/database/mpt/nodes_mocks.go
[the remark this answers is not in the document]
kjezek: no, the script adds/modifies the header in mocks as well

go/database/mpt/proof.go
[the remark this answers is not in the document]
kjezek: can you expand more on this? I do not know if it is better to return []byte or string. 

I guess we should return strings later from the `carmen` API

go/database/mpt/proof_test.go
kjezek: now the proof is returned all the time.

go/database/mpt/proof_test.go
kjezek: reworked from scratch

go/database/mpt/proof.go
```diff
@@ -0,0 +1,109 @@
...
+type proofDb map[common.Hash]rlpEncodedNode
+
+// WitnessProof represents a witness proof.
+// It contains a database of MPT nodes and their hashes.
+// The proof may be verified by iterating the proof database from the
+// root hash to a terminal entry according to a Path
+// representing either an account address or a storage key.
+// If the proof contains an account or a storage slot for the input address or key,
+// the proof is representing this account or storage.
+type WitnessProof proofDb
```
petr-hanzl, line 34: Is there are reason why you create an extra type of `proofDb`? Why dont you just create 
```
type WitnessProof map[common.Hash]rlpEncodedNode
```
and avoid creating two types that are same.

kjezek: right, perhaps it is not necessary, can be simplified.

go/database/mpt/proof.go
```diff
@@ -0,0 +1,115 @@
...
+
+	var innerError error
+
+	found, err := VisitPathToAccount(nodeSource, root, address, MakeVisitor(func(node Node, info NodeInfo) VisitResponse {
+		if res := accountVisitor.Visit(node, info); res == VisitResponseAbort {
+			return VisitResponseAbort
+		}
+		// if account reached, prove storage keys and terminate.
+		switch account := node.(type) {
+		case *AccountNode:
```
HerbertJordan, line 62: nit:
```suggestion
		if account, ok := node.(*AccountNode); ok {
```

go/database/mpt/proof.go
```diff
@@ -0,0 +1,115 @@
...
+					proof:      proof,
+				}
+				found, err := VisitPathToStorage(nodeSource, &account.storage, key, storageVisitor)
+				if err != nil || storageVisitor.err != nil {
+					innerError = errors.Join(innerError, storageVisitor.err, err)
+					return VisitResponseAbort
+				}
+				if !found {
+					innerError = errors.Join(innerError, fmt.Errorf("storage key %s not found", key))
+				}
```
HerbertJordan, line 75: Not finding the key is not an error. It means that the slot's value is zero, which is fine. A proof for a zero-valued slot is valid and useful.

go/database/mpt/proof.go
```diff
@@ -0,0 +1,115 @@
...
+			}
+			return VisitResponseAbort
+		}
+
+		return VisitResponseContinue
+	}))
+
+	if !found {
+		err = errors.Join(err, fmt.Errorf("account %s not found", address))
+	}
```
HerbertJordan, line 85: Not finding the account is not an error. It means the account is empty, which is fine. A proof for an empty account is valid and useful.

go/database/mpt/proof.go
```diff
@@ -0,0 +1,115 @@
...
+			return VisitResponseAbort
+		}
+		// if account reached, prove storage keys and terminate.
+		switch account := node.(type) {
+		case *AccountNode:
+			for _, key := range keys {
+				storageVisitor := &proofExtractionVisitor{
+					nodeSource: nodeSource,
+					proof:      proof,
+				}
```
HerbertJordan, line 67: Looks like you could use the same instance as the `accountVisitor` and `storageVisitor`. No need to differentiate them.

go/database/mpt/proof_test.go
```diff
@@ -0,0 +1,211 @@
...
+				},
+			},
+		},
+	})
+
+	proof, err := CreateWitnessProof(ctxt, &root, address, key1, key2)
+	if err != nil {
+		t.Fatalf("failed to create proof: %v", err)
+	}
+
```
HerbertJordan, line 56: nit: check `proof.IsValid()` and `proof.GetBalance(...)` or `proof.GetStorage(...)` ... maybe here or at another place.

HerbertJordan (changes requested):

go/database/mpt/proof.go
kjezek: I see, ok I will basically ignore not-found

go/database/mpt/proof.go
kjezek: I see, ok I will basically ignore not-found

go/database/mpt/proof.go
kjezek: oh good point, it simplifies the code.

go/database/mpt/proof_test.go
kjezek: ok, but they are not implemented yet, perhaps I may postpone this PR when the methods are ready.

petr-hanzl (approved):
lgtm. Thanks

HerbertJordan (approved):
Looks good to me.

PR #856, opened 2024-05-14
Add license headers CI check

Mike-CZ:
This PR adds check into jenkins pipeline, that all files have correct license header.


HerbertJordan (approved):
I am not familiar with shell scripts, I trust you on that. Please have @kjezek approve this as well before merging.

Thanks for adding this!

PR #859, opened 2024-05-16
Add transient storage

### LINKED ISSUE #858: Support transient storage operations in StateDB
To support the latest revision of the EVM bytecode specification, Carmen's deprecated [`VmStateDB`](https://github.com/Fantom-foundation/Carmen/blob/8b853559b0bb89dfe792f637e889a866341f961e/go/state/state_db.go#L31) interface needs to be extended by two additional functions, to mach the definition of Ethereum's [`StateDB`](https://github.com/ethereum/go-ethereum/blob/d2f00cb54edc4486314c25d9e6c5b739009c2201/core/vm/interface.go#L31) interface:
- [GetTransientState](https://github.com/ethereum/go-ethereum/blob/d2f00cb54edc4486314c25d9e6c5b739009c2201/core/vm/interface.go#L56) to load transient state and
- [SetTransientState](https://github.com/ethereum/go-ethereum/blob/d2f00cb54edc4486314c25d9e6c5b739009c2201/core/vm/interface.go#L57) to update values stored in the transient state

This should involve the following tasks:
- extend the [`VmStateDB`](https://github.com/Fantom-foundation/Carmen/blob/8b853559b0bb89dfe792f637e889a866341f961e/go/state/state_db.go#L31) interface
- add a `*common.FastMap[slotId, common.Value]` transientData map to the [`stateDb`](https://github.com/Fantom-foundation/Carmen/blob/8b853559b0bb89dfe792f637e889a866341f961e/go/state/state_db.go#L177) struct
- reset this map as part of [`resetTransactionContext`](https://github.com/Fantom-foundation/Carmen/blob/8b853559b0bb89dfe792f637e889a866341f961e/go/state/state_db.go#L1308)
- add an implementation of [GetTransientState](https://github.com/ethereum/go-ethereum/blob/d2f00cb54edc4486314c25d9e6c5b739009c2201/core/vm/interface.go#L56) to the [`stateDb`](https://github.com/Fantom-foundation/Carmen/blob/8b853559b0bb89dfe792f637e889a866341f961e/go/state/state_db.go#L177); you can use [`GetState`](https://github.com/Fantom-foundation/Carmen/blob/8b853559b0bb89dfe792f637e889a866341f961e/go/state/state_db.go#L778) as a template
- add an implementation of [SetTransientState](https://github.com/ethereum/go-ethereum/blob/d2f00cb54edc4486314c25d9e6c5b739009c2201/core/vm/interface.go#L57) to the [`stateDb`](https://github.com/Fantom-foundation/Carmen/blob/8b853559b0bb89dfe792f637e889a866341f961e/go/state/state_db.go#L177); you can use [`SetState`](https://github.com/Fantom-foundation/Carmen/blob/8b853559b0bb89dfe792f637e889a866341f961e/go/state/state_db.go#L788) as a template
- fix all compilation issues, e.g. by regenerating mocks 
- add comprehensive unit testing for the added functions and applied modifications

petr-hanzl:
This PR adds `TransientStorage` to `stateDB` with `Get/SetTransientStorage` to the `VMStateDB` interface.
Fixes #858

### DIFF
--- a/go/state/state_db.go
+++ b/go/state/state_db.go
@@ -837,13 +837,12 @@ func (s *stateDB) GetTransientState(addr common.Address, key common.Key) common.
 
 func (s *stateDB) SetTransientState(addr common.Address, key common.Key, value common.Value) {
 	sid := slotId{addr, key}
-	if entry, exists := s.transientStorage.Get(sid); exists {
-		if entry == value {
+	if currentValue, exists := s.transientStorage.Get(sid); exists {
+		if currentValue == value {
 			return
 		}
 		// Save previous value for rollbacks
-		oldValue := entry
-		entry = value
+		oldValue := currentValue
 		s.undo = append(s.undo, func() {
 			s.transientStorage.Put(sid, oldValue)
 		})
--- a/go/state/state_db.go
+++ b/go/state/state_db.go
@@ -836,16 +836,14 @@ func (s *stateDB) GetTransientState(addr common.Address, key common.Key) common.
 }
 
 func (s *stateDB) SetTransientState(addr common.Address, key common.Key, value common.Value) {
-	var oldValue common.Value
 	sid := slotId{addr, key}
-	if currentValue, exists := s.transientStorage.Get(sid); exists {
-		if currentValue == value {
-			return
-		}
-		// Save previous value for rollbacks
-		oldValue = currentValue
+	currentValue, _ := s.transientStorage.Get(sid)
+	if currentValue == value {
+		return
 	}
 
+	// Save previous value for rollbacks
+	oldValue := currentValue
 	s.undo = append(s.undo, func() {
 		s.transientStorage.Put(sid, oldValue)
 	})
--- a/go/state/state_db.go
+++ b/go/state/state_db.go
@@ -194,8 +194,8 @@ type stateDB struct {
 	// A transaction local cache of storage values to avoid double-fetches and support rollbacks.
 	data *common.FastMap[slotId, *slotValue]
 
-	// Transient storage
-	transientData *common.FastMap[slotId, common.Value]
+	// Transient storage is a temporary storage that gets deleted after each transaction.
+	transientStorage *common.FastMap[slotId, common.Value]
 
 	// A transaction local cache of contract codes and their properties.
 	codes map[common.Address]*codeValue
@@ -208,9 +208,6 @@ type stateDB struct {
 
 	// A list of operations undoing modifications applied on the inner state if a snapshot revert needs to be performed.
 	undo []func()
-
-	// A list of operations undoing modifications applied on the inner state if a snapshot revert needs to be performed.
-	transientUndo []func()
 
 	// The refund accumulated in the current transaction.
 	refund uint64
@@ -444,7 +441,7 @@ func createStateDBWith(state State, storedDataCacheCapacity int, canApplyChanges
 		balances:          map[common.Address]*balanceValue{},
 		nonces:            map[common.Address]*nonceValue{},
 		data:              common.NewFastMap[slotId, *slotValue](slotHasher{}),
-		transientData:     common.NewFastMap[slotId, common.Value](slotHasher{}),
+		transientStorage:  common.NewFastMap[slotId, common.Value](slotHasher{}),
 		storedDataCache:   common.NewLruCache[slotId, storedDataCacheValue](storedDataCacheCapacity),
 		reincarnation:     map[common.Address]uint64{},
 		codes:             map[common.Address]*codeValue{},
@@ -454,7 +451,6 @@ func createStateDBWith(state State, storedDataCacheCapacity int, canApplyChanges
 		writtenSlots:      map[*slotValue]bool{},
 		accountsToDelete:  make([]common.Address, 0, 100),
 		undo:              make([]func(), 0, 100),
-		transientUndo:     make([]func(), 0, 100),
 		clearedAccounts:   make(map[common.Address]accountClearingState),
 		emptyCandidates:   make([]common.Address, 0, 100),
 		canApplyChanges:   canApplyChanges,
@@ -529,17 +525,6 @@ func (s *stateDB) CreateAccount(addr common.Address) {
 			value.committed = common.Value{}
 			value.committedKnown = true
 			value.current = common.Value{}
-		}
-	})
-
-	// Reset transient storage.
-	s.transientData.ForEach(func(slot slotId, value common.Value) {
-		if slot.addr == addr {
-			// Support rollback of account creation.
-			s.transientUndo = append(s.transientUndo, func() {
-				s.transientData.Put(slot, value)
-			})
-			s.transientData.Remove(slot)
 		}
 	})
 
@@ -846,30 +831,26 @@ func (s *stateDB) SetState(addr common.Address, key common.Key, value common.Val
 
 func (s *stateDB) GetTransientState(addr common.Address, key common.Key) common.Value {
 	sid := slotId{addr, key}
-	val, _ := s.transientData.Get(sid)
+	val, _ := s.transientStorage.Get(sid)
 	return val
 }
 
 func (s *stateDB) SetTransientState(addr common.Address, key common.Key, value common.Value) {
-	if s.createAccountIfNotExists(addr) {
-		// The account was implicitly created and may have to be removed at the end of the block.
-		s.emptyCandidates = append(s.emptyCandidates, addr)
-	}
 	sid := slotId{addr, key}
-	if entry, exists := s.transientData.Get(sid); exists {
+	if entry, exists := s.transientStorage.Get(sid); exists {
 		if entry == value {
 			return
 		}
 		oldValue := entry
 		entry = value
-		s.transientUndo = append(s.transientUndo, func() {
-			s.transientData.Put(sid, oldValue)
+		s.undo = append(s.undo, func() {
+			s.transientStorage.Put(sid, oldValue)
 		})
 
 	} else {
-		s.transientData.Put(sid, value)
-		s.transientUndo = append(s.transientUndo, func() {
-			s.transientData.Remove(sid)
+		s.transientStorage.Put(sid, value)
+		s.undo = append(s.undo, func() {
+			s.transientStorage.Remove(sid)
 		})
 	}
 }
@@ -1361,9 +1342,9 @@ func (s *stateDB) GetArchiveBlockHeight() (uint64, bool, error) {
 func (s *stateDB) resetTransactionContext() {
 	s.refund = 0
 	s.ClearAccessList()
-	s.transientData.Clear()
+	s.transientStorage.Clear()
 	s.undo = s.undo[0:0]
-	s.transientUndo = s.transientUndo[0:0]
+	s.undo = s.undo[0:0]
 	s.emptyCandidates = s.emptyCandidates[0:0]
 	s.logs = s.logs[0:0]
 }
--- a/go/state/state_db.go
+++ b/go/state/state_db.go
@@ -52,6 +52,8 @@ type VmStateDB interface {
 	GetCommittedState(common.Address, common.Key) common.Value
 	GetState(common.Address, common.Key) common.Value
 	SetState(common.Address, common.Key, common.Value)
+	GetTransientState(common.Address, common.Key) common.Value
+	SetTransientState(common.Address, common.Key, common.Value)
 
 	// Code management.
 	GetCode(common.Address) []byte
@@ -192,6 +194,9 @@ type stateDB struct {
 	// A transaction local cache of storage values to avoid double-fetches and support rollbacks.
 	data *common.FastMap[slotId, *slotValue]
 
+	// Transient storage
+	transientData *common.FastMap[slotId, common.Value]
+
 	// A transaction local cache of contract codes and their properties.
 	codes map[common.Address]*codeValue
 
@@ -203,6 +208,9 @@ type stateDB struct {
 
 	// A list of operations undoing modifications applied on the inner state if a snapshot revert needs to be performed.
 	undo []func()
+
+	// A list of operations undoing modifications applied on the inner state if a snapshot revert needs to be performed.
+	transientUndo []func()
 
 	// The refund accumulated in the current transaction.
 	refund uint64
@@ -436,6 +444,7 @@ func createStateDBWith(state State, storedDataCacheCapacity int, canApplyChanges
 		balances:          map[common.Address]*balanceValue{},
 		nonces:            map[common.Address]*nonceValue{},
 		data:              common.NewFastMap[slotId, *slotValue](slotHasher{}),
+		transientData:     common.NewFastMap[slotId, common.Value](slotHasher{}),
 		storedDataCache:   common.NewLruCache[slotId, storedDataCacheValue](storedDataCacheCapacity),
 		reincarnation:     map[common.Address]uint64{},
 		codes:             map[common.Address]*codeValue{},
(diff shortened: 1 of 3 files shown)


go/state/state_db.go
```diff
@@ -192,6 +194,9 @@ type stateDB struct {
 	// A transaction local cache of storage values to avoid double-fetches and support rollbacks.
 	data *common.FastMap[slotId, *slotValue]
 
+	// Transient storage
+	transientData *common.FastMap[slotId, common.Value]
```
kjezek, line 198: I would name the variable `transientStorage`

go/state/state_db.go
```diff
@@ -522,6 +532,17 @@ func (s *stateDB) CreateAccount(addr common.Address) {
 		}
 	})
 
+	// Reset transient storage.
+	s.transientData.ForEach(func(slot slotId, value common.Value) {
```
kjezek, line 536: I think this cannot happen. Transient state must be fresh for each transaction and a transaction cannot: create an account, modify its transient state, and create the same account again

go/state/state_db.go
```diff
@@ -823,6 +844,36 @@ func (s *stateDB) SetState(addr common.Address, key common.Key, value common.Val
...
 }
 
+func (s *stateDB) GetTransientState(addr common.Address, key common.Key) common.Value {
+	sid := slotId{addr, key}
+	val, _ := s.transientData.Get(sid)
+	return val
+}
+
+func (s *stateDB) SetTransientState(addr common.Address, key common.Key, value common.Value) {
+	if s.createAccountIfNotExists(addr) {
```
kjezek, line 854: I think this operation should not create an account.

kjezek (changes requested):
I think the undo process is not correct. It only tracks changes at the moment, but it is never execute.

You should not create a new `transientUndo` list, but rather use the existing undo list. It is already correctly handled when a snapshot is created and reverted.

go/state/state_db.go
kjezek: this storage should be cleared at every transaction, which it seems is no happening at the moment.

go/state/state_db.go
kjezek: Also it wold be nice to define a new private type for this storage like

`type transientStorage common.FastMap[slotId, common.Value]`

go/state/state_db.go
petr-hanzl: So the value should not be stored, if the account does not exist?

kjezek: store the value, but do not create the account.

not sure if it can happen that the value is set for non-existing account, but it is how it works in ethereum
https://github.com/ethereum/go-ethereum/blob/master/core/state/statedb.go#L522-L539

kjezek (changes requested):
semantics of setting values in the storage  is not correct. This test fails:
```

func TestStateDB_NewestValueCanBeObtained(t *testing.T) {
	ctrl := gomock.NewController(t)
	mock := NewMockState(ctrl)
	db := createStateDBWith(mock, defaultStoredDataCacheSize, true)

	db.SetTransientState(address1, key1, val1)
	db.SetTransientState(address1, key1, val2)

	if got, want := db.GetTransientState(address1, key1), val2; got != want {
		t.Errorf("unexpected value, wanted %v, got %v", want, got)
	}
}

```

go/state/state_db_test.go
```diff
@@ -4143,6 +4143,70 @@ func TestSlotIdOrder(t *testing.T) {
...
+	db := createStateDBWith(mock, defaultStoredDataCacheSize, true)
+
+	db.SetTransientState(address1, key1, val1)
+
+	// Call undo
+	for _, f := range db.undo {
+		f()
+	}
+
+	got := db.GetTransientState(address1, key1)
```
kjezek, line 4158: Please use following syntax to test got/wanted value in tests - here and in other tests 
```

if got, want := db.GetTransientState(address1, key1), val2; got != want {
		t.Errorf("unexpected value, wanted %v, got %v", want, got)
	}
```

notice:
* got, want is assigned and checked inline
* error message starts with lower case and there is no perioed at the end of the sentece

go/state/state_db_test.go
```diff
@@ -4143,6 +4143,70 @@ func TestSlotIdOrder(t *testing.T) {
...
 
+func TestStateDB_SetTransientState_NewValue_UndoRemovesValue(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	mock := NewMockState(ctrl)
+	db := createStateDBWith(mock, defaultStoredDataCacheSize, true)
+
+	db.SetTransientState(address1, key1, val1)
+
+	// Call undo
+	for _, f := range db.undo {
```
kjezek, line 4154: here and in other tests do not explicitly execute the undo method. Instead test that proper semantics of when undo is triggered. In particular, create a snapshot, insert values, tests that values are available. Perform rollback and test that previous variables are available. Do this in 2-3 nested calls, i.e. a few snapshots are created, rolled back, etc. Also include Start/End of transaction that empties the storage

go/state/state_db.go
```diff
@@ -1310,6 +1342,8 @@ func (s *stateDB) GetArchiveBlockHeight() (uint64, bool, error) {
 func (s *stateDB) resetTransactionContext() {
 	s.refund = 0
 	s.ClearAccessList()
+	s.transientStorage.Clear()
+	s.undo = s.undo[0:0]
 	s.undo = s.undo[0:0]
```
HerbertJordan, line 1347: no need to clear the undo list twice

go/state/state_db.go
```diff
@@ -823,6 +829,34 @@ func (s *stateDB) SetState(addr common.Address, key common.Key, value common.Val
...
 
+func (s *stateDB) GetTransientState(addr common.Address, key common.Key) common.Value {
+	sid := slotId{addr, key}
+	val, _ := s.transientStorage.Get(sid)
+	return val
+}
+
+func (s *stateDB) SetTransientState(addr common.Address, key common.Key, value common.Value) {
+	sid := slotId{addr, key}
+	if entry, exists := s.transientStorage.Get(sid); exists {
```
kjezek, line 840: `entry` -> `value`

in Maps, `entry` usually refers to the pair of `key-value`, but here you get the `value`, not the whole entry

go/state/state_db.go
```diff
@@ -823,6 +829,34 @@ func (s *stateDB) SetState(addr common.Address, key common.Key, value common.Val
...
+		if entry == value {
+			return
+		}
+		// Save previous value for rollbacks
+		oldValue := entry
+		entry = value
```
kjezek, line 846: this assignment is redundant, it does nothing

go/state/state_db_test.go
```diff
@@ -4143,6 +4143,71 @@ func TestSlotIdOrder(t *testing.T) {
...
 }
 
+func TestStateDB_SetTransientState_RollbackRemovesValue(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	mock := NewMockState(ctrl)
+	db := CreateStateDBUsing(mock)
+
+	s := db.Snapshot()
+	db.SetTransientState(address1, key1, val1)
+
```
kjezek, line 4153: here you should test that the value is present. If `SetTransientState` was implemented as no-op, this test would still pass.

go/state/state_db_test.go
```diff
@@ -4143,6 +4143,71 @@ func TestSlotIdOrder(t *testing.T) {
...
+	db.RevertToSnapshot(s)
+	if got, want := db.GetTransientState(address1, key1), val0; got != want {
```
kjezek, line 4155: you should compare for `var val common.Value`, I guess that `val0` is defined as an empty value, but it is just one of the test values, you should not rely on the fact it is zero.

go/state/state_db_test.go
```diff
@@ -4143,6 +4143,71 @@ func TestSlotIdOrder(t *testing.T) {
...
+	ctrl := gomock.NewController(t)
+	mock := NewMockState(ctrl)
+	db := CreateStateDBUsing(mock)
+
+	db.SetTransientState(address1, key1, val1)
+	s := db.Snapshot()
+	// Save previous value
+	want := db.GetTransientState(address1, key1)
+
+	db.SetTransientState(address1, key1, val2)
```
kjezek, line 4170: same as above, you should test here that the value has really changed before it was reverted

go/state/state_db_test.go
```diff
@@ -4143,6 +4143,71 @@ func TestSlotIdOrder(t *testing.T) {
...
+	}
+}
+
+func TestStateDB_TransientStorage_IsClearedAfterCallingEndTransaction(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	mock := NewMockState(ctrl)
+	db := createStateDBWith(mock, defaultStoredDataCacheSize, true)
+
+	// Simulate a non-existing account.
+	mock.EXPECT().Exists(address1).Return(false, nil)
```
kjezek, line 4184: I think you do not have to create an account, transient storage does not care if the account exits or not

go/state/state_db_test.go
```diff
@@ -4143,6 +4143,71 @@ func TestSlotIdOrder(t *testing.T) {
...
+
+	db.CreateAccount(address1)
+	db.SetTransientState(address1, key1, val1)
+	db.EndTransaction()
```
kjezek, line 4188: the same as above you should check that the value is present in the storage before it gets cleared by `EndTransaction`

kjezek (changes requested):
thanks for this effort, could you please take a bit more care of test?
Have a look at my comments. 
Also could you add more tests:
* deeper nesting of rollback - two levels of nesting should be fine, check that the value is present, and reset at every step
* check that data does not leak across statedb instances when created via `CreateNonCommittableStateDBUsing`

kjezek:
also make sure test coverage is 100%

image

go/state/state_db.go
petr-hanzl: I used the same naming as in [`SetState`](https://github.com/Fantom-foundation/Carmen/blob/dddf65b4e1f3b09c884d10eb97d23399cf07ea74/go/state/state_db.go#L794) but I guess it make sense to use value here more.

go/state/state_db_test.go
petr-hanzl: I added `valEmpty` as global to the testing data, since there are two places already where its checked.

go/state/state_db.go
```diff
@@ -823,6 +829,33 @@ func (s *stateDB) SetState(addr common.Address, key common.Key, value common.Val
...
 
+func (s *stateDB) GetTransientState(addr common.Address, key common.Key) common.Value {
+	sid := slotId{addr, key}
+	val, _ := s.transientStorage.Get(sid)
+	return val
+}
+
+func (s *stateDB) SetTransientState(addr common.Address, key common.Key, value common.Value) {
+	sid := slotId{addr, key}
+	if currentValue, exists := s.transientStorage.Get(sid); exists {
```
kjezek, line 840: since non-existing value is treated as empty value, this code could be simplified just adding previous value in the undo lis.  For a non existing value it would be equal to an empty value, and if rollback happens, it would set the empty value in the map.

go/state/state_db_test.go
```diff
@@ -4143,6 +4144,135 @@ func TestSlotIdOrder(t *testing.T) {
...
+	db.SetTransientState(address1, key1, val1)
+	if got, want := db.GetTransientState(address1, key1), val1; got != want {
+		t.Errorf("unexpected value, wanted %v, got %v", want, got)
+	}
+
+	if got, want := len(db.undo), 1; got != want {
+		t.Errorf("unexpected undo len, wanted: 1, got: %v", len(db.undo))
+	}
+
+	if got, want := reflect.ValueOf(db.undo[0]), reflect.ValueOf(undo); got != want {
```
kjezek, line 4271: perhaps not necessary to test this - it is too unlikely

kjezek (dismissed):
looks good, thanks

go/state/state_db_test.go
petr-hanzl: You are right. This can probably never be false.

go/state/state_db.go
petr-hanzl: Im not sure if I understood your comment correctly. Please check latest commit.

kjezek: the method can be even simpler: 

```
	sid := slotId{addr, key}
	currentValue, _ := s.transientStorage.Get(sid)
	if currentValue == value {
		return
	}

	// Save previous value for rollbacks
	oldValue := currentValue
	s.undo = append(s.undo, func() {
		s.transientStorage.Put(sid, oldValue)
	})

	s.transientStorage.Put(sid, value)

```

kjezek (approved):
The SetTransientState method could be this:
```

func (s *stateDB) SetTransientState(addr common.Address, key common.Key, value common.Value) {
	sid := slotId{addr, key}
	currentValue, _ := s.transientStorage.Get(sid)
	if currentValue == value {
		return
	}

	// Save previous value for rollbacks
	oldValue := currentValue
	s.undo = append(s.undo, func() {
		s.transientStorage.Put(sid, oldValue)
	})

	s.transientStorage.Put(sid, value)
}
```

kjezek (approved):
thanks, looks good. Please do not forget to squash changes

kjezek (changes requested):
sorry for changing my mind. I noticed, and forget before, that the test for `CreateNonCommittableStateDBUsing`
is still missing

kjezek (dismissed):
thanks for adding the missing test

go/state/state_db_test.go
```diff
@@ -4143,6 +4144,177 @@ func TestSlotIdOrder(t *testing.T) {
...
+		t.Errorf("unexpected value, wanted %v, got %v", want, got)
+	}
+
+	db.SetTransientState(address1, key1, val2)
+	if got, want := db.GetTransientState(address1, key1), val2; got != want {
+		t.Errorf("unexpected value, wanted %v, got %v", want, got)
+	}
+}
+
+func TestStateDB_SetTransientState_ValueIsNotSetIfSame(t *testing.T) {
```
HerbertJordan, line 4230: nit: the property referenced by the name of this case (ValueIsNotSetIfSame) is not tested in the code. You would have to check that the undo list did not grow, as it is done in the next test -- so this one is not redundant?

go/state/state_db_test.go
```diff
@@ -4143,6 +4144,177 @@ func TestSlotIdOrder(t *testing.T) {
...
+	db := CreateNonCommittableStateDBUsing(mock)
+
+	db.SetTransientState(address1, key1, val1)
+	if got, want := db.GetTransientState(address1, key1), val1; got != want {
+		t.Errorf("unexpected value, wanted %v, got %v", want, got)
+	}
+
+	db.Release()
+	// TransientStorage must be cleared
+	db2 := CreateNonCommittableStateDBUsing(mock)
```
HerbertJordan, line 4312: It is not guaranteed that `CreateNonCommittableStateDBUsing` is returning the StateDB instance released in line 4310 -- and it is increasingly unlikely when running lots of tests, since the pool of StateDB instances grows.

You can try and call `CreateNonCommittableStateDBUsing` multiple times without releaseing the DB you got until you retrieve the one you had.

Alternatively, the test could also call `resetState` explicitly and check that the transient state got reset.

HerbertJordan (approved):
Looks good, thanks for adding this feature. I added a few small remarks on the tests you might address before merging. Thanks!

go/state/state_db_test.go
petr-hanzl: I guess explicitly calling `resetState` would be the best practice for this test.

go/state/state_db_test.go
petr-hanzl: You are right, this test is redundant - removed.

kjezek (approved):

PR #870, opened 2024-05-22
Add Amount type into Facade Api

Mike-CZ:
This PR adds usage of [uint256](https://pkg.go.dev/github.com/holiman/uint256#Int) package for Balances in Facade Api instead of `big.Int` package.  The replacement for state will be done in following PRs.


HerbertJordan (changes requested):
Looks good, but I think a few utility functions should be added in addition. In particular
- a factory function `NewAmount()` to avoid the need for creating an explicit `*uint256.Int` just to discard it again
- a `String()` function for the `Amount` type

go/carmen/carmen.go
```diff
@@ -401,7 +400,7 @@ type BulkLoad interface {
 
 	// SetBalance sets the balance of the given account address
 	// to the input value.
-	SetBalance(Address, *big.Int)
+	SetBalance(Address, Amount)
```
kjezek, line 403: should it be here and in other methods a pointer to `Amount`? It is done so in Ethereum

go/carmen/query.go
```diff
@@ -12,27 +12,29 @@ package carmen
...
 )
 
 type queryContext struct {
 	state state.State
 	err   error
 }
 
-func (c *queryContext) GetBalance(address Address) *big.Int {
+func (c *queryContext) GetBalance(address Address) Amount {
+	ret := uint256.NewInt(0)
```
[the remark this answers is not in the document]
kjezek: for error cases, I think it should create `Amount{}` not current `Amount(*ret)`
for ok case it should call a factory method, to be added

go/carmen/carmen.go
Mike-CZ: It makes sense in setters to accept pointers

go/carmen/amount.go
```diff
@@ -0,0 +1,115 @@
...
+	result := uint256.Int{}
+	overflow := result.SetFromBig(b)
+	if overflow {
+		panic("Cannot construct U256 from big.Int with more than 256 bits")
+	}
+	return Amount{internal: result}
+}
+
+// Uint64 returns the amount as an uint64.
+func (a Amount) Uint64() uint64 {
```
HerbertJordan, line 84: There should a an accompanying `IsUint64` to test whether the Amount fits into the 64-bit value range.

Alternatively, this function should return an error indicating a potential overflow.

go/carmen/amount.go
```diff
@@ -0,0 +1,115 @@
...
+	return a.internal.IsZero()
+}
+
+// ToBig returns a bigInt version of the amount.
+func (a Amount) ToBig() *big.Int {
+	return a.internal.ToBig()
+}
+
+// Add returns the sum of two amounts.
+func (a Amount) Add(b Amount) Amount {
```
HerbertJordan, line 99: If I read
```
a.Add(b)
```
in a code review I would assume that the value of `b` is added to `a`. But in fact, this operation has no effect right now. The sum is computed and returned, but then discarded.

The operation, as it is defined here, should take both inputs as parameters, producing
```
Add(a,b)
```
However, the problem with this is that Go has no function overloading and the `Add` function name can only be used once.

Potential solutions for this could be:
- add a suffix to the function name, like `AddAmount(a,b)`
- place `Amount` and its operations in an `amount` package and use something like `amount.Add(a,b)`
- do the same as `big` and `uint256` -- make them member functions; in this case you would have to write `Amount{}.Add(a,b)`
- update this `Add` function to have a add-and-assign semantic, like C++'s `+=` operator. In this case, `a.Add(b)` would sum up `a` and `b`, write the result into `a` and return `a`. The down-side is that `a` needs to be a pointer parameter
- implement a generic function `Add(a, b T) T` that checks the types of `a` and `b` to perform the computation

All not great options.  Wdyt?

go/carmen/amount_test.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+func TestAmount_NewAmountFromUint256(t *testing.T) {
+	tests := []struct {
+		in  *uint256.Int
+		out Amount
+	}{
+		{uint256.NewInt(0), NewAmount()},
+		{uint256.NewInt(1), NewAmount(1)},
+		{uint256.NewInt(256), NewAmount(256)},
```
HerbertJordan, line 54: Please add an example that uses more than 64 and 128 bit.

go/carmen/amount_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+func TestAmount_NewAmountFromBytes(t *testing.T) {
+	// test empty amount
+	empty := NewAmountFromBytes()
+	if empty != NewAmount() {
+		t.Errorf("empty amount should be zero")
+	}
+
+	// test amount with single value
+	amount := NewAmountFromBytes([]byte{1, 2, 3, 4}...)
```
HerbertJordan, line 74:
```suggestion
	amount := NewAmountFromBytes(1, 2, 3, 4)
```

HerbertJordan, line 74: nit: a test with more than 8 byte would also test the correct handling of the word order.

go/carmen/amount.go
```diff
@@ -0,0 +1,115 @@
...
+func (a Amount) IsZero() bool {
+	return a.internal.IsZero()
+}
+
+// ToBig returns a bigInt version of the amount.
+func (a Amount) ToBig() *big.Int {
+	return a.internal.ToBig()
+}
+
+// Add returns the sum of two amounts.
```
HerbertJordan, line 98:
```suggestion
// Add returns the sum of two amounts. Overflows are not checked.
```

go/carmen/amount.go
```diff
@@ -0,0 +1,115 @@
...
+func (a Amount) Add(b Amount) Amount {
+	result := Amount{}
+	result.internal.Add(&a.internal, &b.internal)
+	return result
+}
+
+// Sub returns the difference of two amounts.
```
HerbertJordan, line 105:
```suggestion
// Sub returns the difference of two amounts. Underflows are not checked.
```

go/carmen/carmen.go
HerbertJordan: I do not think there is a benefit for passing it as a pointer. It might actually introduce additional memory allocations by Go to move the passed element to the heap.

Keys or values in other API functions are also 32-byte large and passed by value. We have not seen problems with this so far, and I would favor consistency.

HerbertJordan (dismissed):
Thanks for introducing the new Amount type. Looks good to me, just the style of the `Add` and `Sub` operation needs some discussion.

It could be that `Add` and `Sub` are not really needed by this PR. Maybe we leave this for future work.

go/carmen/amount.go
```diff
@@ -0,0 +1,115 @@
...
+		result.SetUint64(0)
+	case 1:
+		result.SetUint64(args[0])
+	case 2:
+		high := uint256.NewInt(args[0])
+		high.Lsh(high, 64) // Shift left by 64 bits
+		low := uint256.NewInt(args[1])
+		result.Set(high.Add(high, low)) // Combine high and low parts
+	default:
+		panic("NewAmount supports up to 1 arguments only")
```
kjezek, line 43:
```suggestion
		panic("NewAmount supports up to 4 arguments")
```

go/carmen/amount.go
```diff
@@ -0,0 +1,115 @@
...
+	return Amount{internal: *value}
+}
+
+// NewAmountFromBytes creates a new Amount instance from up to 32 byte arguments.
+// The arguments are given in the order from most significant to the least
+// significant by padding leading zeros as needed. No argument results in a
+// value of zero.
+func NewAmountFromBytes(bytes ...byte) Amount {
+	if len(bytes) > 32 {
+		panic("Too many arguments")
```
kjezek, line 59: make the message consistent with other constructors

go/carmen/amount.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+	result := Amount{}
+	result.internal.SetBytes(bytes)
+	return result
+}
+
+// NewAmountFromBigInt creates a new Amount instance from a big.Int.
+// The constructor panics if the big.Int is negative or has more than 256 bits.
```
kjezek, line 67: make the doc consistent. Here you mention that the constructor may panic, while it is not mentioned for other constructors

go/carmen/amount.go
```diff
@@ -0,0 +1,115 @@
...
+	return Amount{internal: result}
+}
+
+// NewAmountFromUint256 creates a new amount from an uint256.
+func NewAmountFromUint256(value *uint256.Int) Amount {
+	return Amount{internal: *value}
+}
+
+// NewAmountFromBytes creates a new Amount instance from up to 32 byte arguments.
+// The arguments are given in the order from most significant to the least
```
kjezek, line 54: is not it just a BigEndian? If yes, do not complicate the documentation with your own explanation and refer to this term.

go/carmen/amount.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+// NewAmountFromBigInt creates a new Amount instance from a big.Int.
+// The constructor panics if the big.Int is negative or has more than 256 bits.
+func NewAmountFromBigInt(b *big.Int) Amount {
+	if b == nil {
+		return NewAmount()
+	}
+	if b.Cmp(big.NewInt(0)) == -1 {
+		panic("Cannot construct Amount from negative big.Int")
```
kjezek, line 73:
```suggestion
		panic("cannot construct Amount from negative big.Int")
```

go/carmen/amount.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+	result := uint256.Int{}
+	overflow := result.SetFromBig(b)
+	if overflow {
+		panic("Cannot construct U256 from big.Int with more than 256 bits")
```
kjezek, line 78:
```suggestion
		panic("cannot construct U256 from big.Int with more than 256 bits")
```

go/carmen/amount_test.go
```diff
@@ -0,0 +1,113 @@
...
+	}{
+		{uint256.NewInt(0), NewAmount()},
+		{uint256.NewInt(1), NewAmount(1)},
+		{uint256.NewInt(256), NewAmount(256)},
+	}
+
+	for _, test := range tests {
+		got := NewAmountFromUint256(test.in)
+		want := test.out
+		if want != got {
```
kjezek, line 60: please inline this in the  `if got, want := .., .. ; got != want  {` pattern

go/carmen/carmen.go
kjezek: thanks for explanation

kjezek (dismissed):
thanks for this effort. I have just a few comments to the code style, please see my comments

go/carmen/amount.go
Mike-CZ: Depends whether we want to stick to the no pointers approach. If not, the solution with pointers is the cleanest imo. If we do, I'd prefer to make it as package's function `amount.Add(a,b)`.

HerbertJordan: If it is indeed not needed, please remove it. 
We will figure this out when we have a use case.

HerbertJordan (approved):
Looks good to me!

PR #872, opened 2024-05-22
changes RLP encode methods to standalone, not part of the etHasher

kjezek:
this PR makes RLP encoder functions standalone, not part of the `ethHaser`.  These methods did not use any functionality from the hasher while creating the instance is complicated, it needs to go through MPT configuration.

### DIFF
--- a/go/database/mpt/hasher.go
+++ b/go/database/mpt/hasher.go
@@ -411,7 +411,7 @@ func (h ethHasher) updateHashesInternal(
 				embedded[cur.node.Id()] = true
 			} else {
 				// Encode the node using RLP and compute its hash.
-				data, e := h.encode(node, manager, data)
+				data, e := encode(node, manager, data)
 				if e != nil {
 					cur.handle.Release()
 					err = e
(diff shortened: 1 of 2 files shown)


go/database/mpt/hasher.go
```diff
@@ -467,27 +467,27 @@ func (h ethHasher) getHash(ref *NodeReference, source NodeSource) (common.Hash,
...
-// hashes and embedded flags are updated. If the manager is nil, this operation is a
-// read-only operation accepting the current hashes and embedded flags as the true value
-// even if dirty flags are set. The node and source parameter must not be nil.
-func (h ethHasher) encode(
+// fetched from the given source for deriving the encoding.
+// The node and source parameter must not be nil.
+// The result is stored in the input slice, and the slice is returned as well.
+// The reason for having the slice on the input is that the encoding of nodes
+// is a frequent operation and the slice can be reused to reduce memory allocations.
+func encode(
```
HerbertJordan, line 475: Good idea to make this stand-alone, detached from the hasher. 

However, in this case we share the general MPT package namespace with other functions and should be more specific with the function names. Maybe `encodeToRlp` or `rlpEncode`?

HerbertJordan (approved):
Nice improvement, thanks!

go/database/mpt/hasher.go
kjezek: ok, used `encodeToRlp`

HerbertJordan (approved):

PR #880, opened 2024-05-23
proposes API for witness proofs

kjezek:
This PR proposes an API for witness proof.

It includes features for 
* merging and splitting proofs, 
* extraction of proofs from databse
* proving of account and addresses

This PR creates only stub methods and API, implementation will follow-up

### DIFF
--- a/go/database/mpt/proof.go
+++ b/go/database/mpt/proof.go
@@ -1,0 +1,117 @@
+// Copyright (c) 2024 Fantom Foundation
+//
+// Use of this software is governed by the Business Source License included
+// in the LICENSE file and at fantom.foundation/bsl11.
+//
+// Change Date: 2028-4-16
+//
+// On the date above, in accordance with the Business Source License, use of
+// this software will be governed by the GNU Lesser General Public License v3.
+
+package mpt
+
+import (
+	"github.com/Fantom-foundation/Carmen/go/common"
+	"github.com/Fantom-foundation/Carmen/go/common/tribool"
+)
+
+//go:generate mockgen -source proof.go -destination proof_mocks.go -package witness
+
+// rlpEncodedNode is an RLP encoded MPT node.
+type rlpEncodedNode []byte
+
+// proofDb is a database of RLP encoded MPT nodes and their hashes that represent witness proofs.
+type proofDb map[common.Hash]rlpEncodedNode
+
+// WitnessProof represents a witness proof.
+// It contains a database of MPT nodes and their hashes.
+type WitnessProof struct {
+	proofDb
+}
+
+// CreateWitnessProof creates a witness proof for the input account address
+// and possibly storage slots of the same account under the input storage keys.
+// This method may return an error when it occurs in the underlying database.
+func CreateWitnessProof(nodeSource NodeSource, root *NodeReference, address common.Address, keys ...common.Key) (WitnessProof, error) {
+	panic("not implemented")
+}
+
+// Add merges the input witness proof into the current witness proof.
+func (p WitnessProof) Add(other WitnessProof) {
+	panic("not implemented")
+}
+
+// Extract extracts a sub-proof for a given account and selected storage locations from this proof.
+// It returns a copy that contains only the data necessary for proofing the given address and storage keys.
+// The resulting proof covers proofs for the intersection of the requested properties (account information and slots)
+// and the properties covered by this proof. The second return parameter indicates whether everything that was requested could be covered. If so
+// it is set to true, otherwise it is set to false.
+func (p WitnessProof) Extract(root common.Hash, address common.Address, keys ...common.Key) (WitnessProof, bool) {
+	panic("not implemented")
+}
+
+// IsValid checks that this proof is self-consistent. If the result is true, the proof can be used
+// for extracting verified information. If false, the proof is corrupted and should be discarded.
+func (p WitnessProof) IsValid() bool {
+	panic("not implemented")
+}
+
+// GetAccountInfo extracts an account info from the witness proof for the input root hash and the address.
+// If the witness proof contains an account for the input address, it returns its information.
+// If the proof does not contain an account, it returns false.
+// The method may return an error if the proof is invalid.
+func (p WitnessProof) GetAccountInfo(root common.Hash, address common.Address) (AccountInfo, bool, error) {
+	panic("not implemented")
+}
+
+// GetBalance extracts a balance from the witness proof for the input root hash and the address.
+// If the witness proof contains the requested account for the input address for the given root hash, it returns its balance.
+// If the proof does not cover the requested account, it returns false.
+// The method may return an error if the proof is invalid.
+func (p WitnessProof) GetBalance(root common.Hash, address common.Address) (common.Balance, bool, error) {
+	panic("not implemented")
+}
+
+// GetNonce extracts a nonce from the witness proof for the input root hash and the address.
+// If the witness proof contains the account for the input address, it returns its nonce.
+// If the proof does not contain the account, it returns false.
+// The method may return an error if the proof is invalid.
+func (p WitnessProof) GetNonce(root common.Hash, address common.Address) (common.Nonce, bool, error) {
+	panic("not implemented")
+}
+
+// GetCodeHash extracts a code hash from the witness proof for the input root hash and the address.
+// If the witness proof contains the account for the input address, it returns its code hash.
+// If the proof does not contain the account, it returns false.
+// The method may return an error if the proof is invalid.
+func (p WitnessProof) GetCodeHash(root common.Hash, address common.Address) (common.Hash, bool, error) {
+	panic("not implemented")
+}
+
+// GetState extracts a storage slot from the witness proof for the input root hash, account address and the storage key.
+// If the witness proof contains the input storage slot for the input key, it returns its value.
+// If the proof does not contain the slot, it returns false.
+// The method may return an error if the proof is invalid.
+func (p WitnessProof) GetState(root common.Hash, address common.Address, key common.Key) (common.Value, bool, error) {
+	panic("not implemented")
+}
+
+// AllStatesZero checks that all storage slots are empty for the input root hash,
+// account address and the storage key range. If the witness proof contains all empty slots
+// for the input key range, it returns true. If there is at least one non-empty slot,
+// it returns false. If the proof is not complete, it returns unknown. An incomplete proof
+// is a proof where the input address or key terminates in a node that is not a correct
+// value node, or an empty node.
+func (p WitnessProof) AllStatesZero(root common.Hash, address common.Address, from, to common.Key) (tribool.Tribool, error) {
+	panic("not implemented")
+}
+
+// AllAddressesEmpty checks that all accounts are empty for the input root hash and the address range.
+// If the witness proof contains all empty accounts for the input address range, it returns true.
+// An empty account is an account that contains a zero balance, nonce, and code hash.
+// If there is at least one non-empty account, it returns false. If the proof is not complete,
+// it returns unknown. An incomplete proof is a proof where the input address terminates in a node
+// that is not a correct account node.
+func (p WitnessProof) AllAddressesEmpty(root common.Hash, from, to common.Address) (tribool.Tribool, error) {
+	panic("not implemented")
+}
--- a/go/database/mpt/nodes.go
+++ b/go/database/mpt/nodes.go
@@ -213,7 +213,7 @@ type Node interface {
 	Visit(source NodeSource, thisRef *NodeReference, depth int, visitor NodeVisitor) (abort bool, err error)
 }
 
-// NodeSource is a interface for any object capable of resolving NodeIds into
+// NodeSource is an interface for any object capable of resolving NodeIds into
 // Nodes. It is intended to be implemented by a Node-governing component
 // handling the life-cycle of nodes and loading/storing nodes to persistent
 // storage. It also serves as a central source for trie configuration flags.


go/database/mpt/proof.go
```diff
@@ -0,0 +1,116 @@
...
+// WitnessProof represents a witness proof.
+// It contains a database of MPT nodes and their hashes.
+// The proof may be verified by iterating the proof database from the
+// root hash to a terminal entry according to a Path
+// representing either an account address or a storage key.
+// The proof is valid if it can be iterated to an expected terminal entry.
+type WitnessProof proofDb
+
+// Merge merges the input witness proof into the current witness proof.
+func (p *WitnessProof) Merge(other *WitnessProof) {
```
HerbertJordan, line 24: Since the underlying type of a `WitnessProof` is a `map` it is implicitly a reference type. So you could get away without the pointer in `*WitnessProof` and just use `WitnessProof` for the `this` and other arguments, as well as results.

This would eliminate the situation where a proof my by `nil`.

HerbertJordan, line 24: Maybe rename this to `Add`?
When reading
```
a.Add(b)
```
in code I would expect `a` to be modified and `b` to remain as it is.

For something called `Merge` I would expect a use case to look like
```
c := Merge(a,b)
```
where `c` contains the content of `a` and `b` and the arguments are not modified.

go/database/mpt/proof.go
```diff
@@ -0,0 +1,116 @@
...
+// It returns a copy that contains only proof for the storage key.
+// If the key is not found in the current proof, it returns an empty proof.
+func (p *WitnessProof) ExtractStorageProof(nodeSource NodeSource, storageRoot common.Hash, address common.Address, key common.Key) *WitnessProof {
+	return &WitnessProof{}
+}
+
+// ProveAccount verifies the witness proof against the input root hash and address.
+// If the witness proof represents the input account for the input address, it returns the account node.
+// If the proof is invalid or does not represent the account, it returns nil.
+func (p *WitnessProof) ProveAccount(nodeSource NodeSource, root common.Hash, address common.Address) *AccountNode {
```
HerbertJordan, line 47: The resulting `AccountNode` has only private fields that can not be checked by users of this function for the value of the Balance or Nonce. 

You could return an [AccountInfo](https://github.com/Fantom-foundation/Carmen/blob/842587ea653b1fdbe8d871b40a3a9d127b8a7ffa/go/database/mpt/account.go#L19) instead.

go/database/mpt/proof.go
```diff
@@ -0,0 +1,116 @@
...
+// ExtractStorageProof extracts a sub-proof from the current witness proof.
+// It returns a copy that contains only proof for the storage key.
+// If the key is not found in the current proof, it returns an empty proof.
+func (p *WitnessProof) ExtractStorageProof(nodeSource NodeSource, storageRoot common.Hash, address common.Address, key common.Key) *WitnessProof {
+	return &WitnessProof{}
+}
+
+// ProveAccount verifies the witness proof against the input root hash and address.
+// If the witness proof represents the input account for the input address, it returns the account node.
+// If the proof is invalid or does not represent the account, it returns nil.
```
HerbertJordan, line 46: It might be useful for the user to differentiate whether the proof is invalid or just does not include the requested account.

go/database/mpt/proof.go
```diff
@@ -0,0 +1,116 @@
...
+func (p *WitnessProof) Merge(other *WitnessProof) {
+	for k, v := range *other {
+		(*p)[k] = v
+	}
+}
+
+// ExtractAccountProof extracts a sub-proof from the current witness proof.
+// It returns a copy that contains only proof for the input address and storage keys.
+// If the input address or keys are not found in the current proof, it returns an empty proof.
+func (p *WitnessProof) ExtractAccountProof(nodeSource NodeSource, root common.Hash, address common.Address, keys ...common.Key) *WitnessProof {
```
HerbertJordan, line 33: This operation (and the other member functions below) do not need a `NodeSource` argument. All information should already be contained in the proof.

go/database/mpt/proof.go
```diff
@@ -0,0 +1,116 @@
...
+	return &WitnessProof{}
+}
+
+// ExtractStorageProof extracts a sub-proof from the current witness proof.
+// It returns a copy that contains only proof for the storage key.
+// If the key is not found in the current proof, it returns an empty proof.
+func (p *WitnessProof) ExtractStorageProof(nodeSource NodeSource, storageRoot common.Hash, address common.Address, key common.Key) *WitnessProof {
```
HerbertJordan, line 40: This one seems to be a special case of the function above for a single key.

If the intention is to produce a proof that only covers the path from the AccountNode to the Value node storing the key, the resulting proof would be kinda useless since the root hash of the storage trie is not available / accessible. You would not be able to verify the proofs for anything in such a fragment.

I would remove this function and only offer the one above.

go/database/mpt/proof.go
```diff
@@ -0,0 +1,116 @@
...
+// Merge merges the input witness proof into the current witness proof.
+func (p *WitnessProof) Merge(other *WitnessProof) {
+	for k, v := range *other {
+		(*p)[k] = v
+	}
+}
+
+// ExtractAccountProof extracts a sub-proof from the current witness proof.
+// It returns a copy that contains only proof for the input address and storage keys.
+// If the input address or keys are not found in the current proof, it returns an empty proof.
```
HerbertJordan, line 32: If some keys are not provable, will the resulting proof cover those or will it be anyway empty?

go/database/mpt/proof.go
```diff
@@ -0,0 +1,116 @@
...
+	if node == nil {
+		return nil
+	}
+	return node.(*AccountNode)
+}
+
+// ProveStorage verifies the witness proof against the input storage root hash, and storage key.
+// If the witness proof represents the input storage slot for the input key, it returns the value node.
+// If the proof is invalid or does not represent the storage slot, it returns nil.
+func (p *WitnessProof) ProveStorage(nodeSource NodeSource, storageRoot common.Hash, key common.Key) *ValueNode {
```
HerbertJordan, line 59: I guess it is sufficient to return a `Value` instead of a `ValueNode` here. The `value` in the `ValueNode` is private.

go/database/mpt/proof.go
```diff
@@ -0,0 +1,116 @@
...
+	if err != nil {
+		p.err = err
+		return VisitResponseAbort
+	}
+	hash := common.Keccak256(rlp)
+
+	(*p.proof)[hash] = rlp
+
+	return VisitResponseContinue
+}
```
HerbertJordan, line 116: How about leaning more into the concept of a `WitnessProof` as a `DataBase`, as described in the comments above. The user interaction could look somehow like this:
```

	proof := CreateProof( ... )

	valid := proof.IsValid()
	info, err := proof.GetAccountInfo(root, addr)
	balance, err := proof.GetBalance(root, addr)
	nonce, err := proof.GetNonce(root, addr)
	hash, err := proof.GetCodeHash(root, addr)
	value, err := proof.GetState(root, addr, key)

	empty, err := proof.AllStatesZero(root, addr, from, to)
	empty, err := proof.AllAddressesEmpty(root, from, to)
```
where
 - `IsValid` checks that the proof has all valid entries; this means, RLP encodings are valid encodings and their hashes match the key in the map
 - `GetXXX` tries to walk from the given root hash to the requested property; if the path could be completed, the requested property is returned; if nodes are missing, an error is returned (the proof does not cover this property)
 - `AllStatesZero` returns true if from the content of the proof it can be shown that the range key `from` to key `to` (inclusive) is all zero; if there is a non-zero value in this range, false is returned; it returns an error if there is not sufficient data in the proof to show either property
 - `AllAccountsEmpty` does the same for a range of accounts

A plus for this is that the proof only needs to be evaluated once before it can be used to extract information multiple times. We could also produce an error in case a user tries to access data without prior validation.

Wdyt?

go/database/mpt/proof.go
kjezek: what would be an invalid proof?
1. hash(rlp) != hash
2. proof leads to a non-empty hash, but this hash is not in the proof
?

go/database/mpt/proof.go
kjezek: I used it only because `AddressToNibblePath` needs the `NodeSource`. But perhaps this is a wrong design at first place. It checks via the NodeSource if the path must be hashed and uses the cached hasher.

Should we just assume "S5" scheme here, ignore caching and call `common.Keccak256(address)` ?

go/database/mpt/proof.go
kjezek: ok, will remove it. I was not sure if there is a use-case where only key(s) needs to be extracted

go/database/mpt/proof.go
kjezek: I will update the documentation. I think it is not correct at the moment. I think it should work similar to proving account/keys, it will keep adding nodes into the proof as far as it can reach. If the account or a key does not lead to a result, the proof will contain only nodes that could be reach. It can be empty if no node was reached.

go/database/mpt/proof.go
kjezek: right, will change

go/database/mpt/proof.go
```diff
@@ -0,0 +1,116 @@
+package mpt
+
+import (
+	"github.com/Fantom-foundation/Carmen/go/common"
+)
+
+//go:generate mockgen -source proof.go -destination proof_mocks.go -package witness
+
+// RlpEncodedNode is an RLP encoded MPT node.
+type RlpEncodedNode []byte
```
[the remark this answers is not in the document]
kjezek: ok, will change

go/database/mpt/proof.go
kjezek: ok, changed. I meant change it everywhere right?

go/database/mpt/proof.go
kjezek: yes, it sounds good!

shall we return `err` or `bool`?  If the client asks for a missing information, it is not an error state.  Err state would be if the proof is not valid for instance.  We could say in the err where the path could not continue, but not sure user needs / should have this information. 

Perhaps we should even return `(X, bool, err)` - err for invalid proof, bool for missing proof for the input. 
or `(*X, err) `- returning nil for a missing element - but perhaps the version with `bool` is more golang style

kjezek: perhaps `CreateProof( ... )`  should also return `bool` to be discarded if the poof could not be created for an input at all.

go/database/mpt/proof.go
kjezek: but not sure if such a partial proof is useful?

go/database/mpt/proof.go
kjezek: added two Getters at the moment, will add more next

go/database/mpt/proof.go
HerbertJordan: I think such a proof is useful, since it proofs that the key's value is 0.

go/database/mpt/proof.go
HerbertJordan: > perhaps CreateProof( ... ) should also return bool to be discarded if the poof could not be created for an input at all.

`CreateProof` should only fail if there is an IO error. Attempting to proof a key or account that is not the database is a valid use case. In case of a key, it would proof that the value is zero. I case of an account it would proof that the account is empty.

kjezek: right, but I wanted to ask more, if we should return `bool` also to show that the address/key was not found. If we do not return it, the user will not know that the input data was not proven unless he calls Getters to read the value back

HerbertJordan: A few more ideas:

To make sure that the user of the API is considering the potential invalid proof or lack of information, and not just ignores the error accidentally, we could wrap the result into a `Optional` like container. Something like this:
```
type Certified[T any] struct {
	valid bool
	value T
}

func newCertified[T any](value T) Certified[T] {
	return Certified[T]{valid: true, value: value}
}

func newMissing[T any]() Certified[T] {
	return Certified[T]{}
}

func (c Certified[T]) IsValid() bool {
	return c.valid
}

func (c Certified[T]) Value() T {
	if !c.valid {
		panic("value is not valid")
	}
	return c.value
}
```
and then have functions like
```
func (p WitnessProof) GetBalance(root Hash, addr Address) Certified[Balance] { ... }
```
This would make properties only accessible after a check, like this
```
balance := proof.GetBalance()
if !balance.IsValid() {
  // handle missing balance
} else {
  // use balance
  _ = balance.Value()
}
```
The down-side of this is that the error handling pattern is more Go-style conform.

Basically there are three options:
 - use `Value, error` and trust that the user is checking the error
 - use `*Value` and trust that the user checks for `nil`
 - use `Certified[Value]` and trust that the user checks for validity
 
All not good options ... any other ideas?

go/database/mpt/proof.go
kjezek: similar question to the one below: should we return bool to inform that the input was not proven? Or do we just return the proof, and the user is expected to call one of the getters to actually see that the input is not proven?

go/database/mpt/proof.go
HerbertJordan: > shall we return err or bool? If the client asks for a missing information, it is not an error state. Err state would be if the proof is not valid for instance. We could say in the err where the path could not continue, but not sure user needs / should have this information.

I agree, if we go for this, it should be a `bool`, not an `error`. 

About paths that can not be continued:
 - if the path ends in an empty node, this means the value that one looks for is zero => valid proof for the zero value
 - if nodes are missing in the proof and thus no next step is possible, there is not enough information in the proof to know the value or its validity

There are three cases we need to support:
- the proof is valid and the property looking for is covered by the nodes in the proof => return the value
- the proof is valid, but the property is not covered => return something indicating `unknown`
- the proof is not valid => return something indicating an invalid proof

kjezek: right, this lead me to a return type `value, bool, err`

* `value returned, true, nil `- when the value was found
* `value undefined/empty, false, nil` - when the value could not be found
* `value undefined/empty, false, err != nil` - when the proof is not valid, basically it could not be parsed due to wrong RLP, not sure if missing nodes are also an invalid stare, or it is just the same as if an empty node is reached

I would prefer returning this tuple of three elements.  I think this is a natural way in golang. I know the pattern with `Optional` from Java, but I think it is too sophisticated for go. I feel go code is usually written simpler

go/database/mpt/proof.go
HerbertJordan: The only way it can not be proven is if there was an IO error. In this case an error is already expected. No need for a bool.

go/database/mpt/proof.go
HerbertJordan: It just makes client code look like this:
```
value, valid, err := proof.GetBalance(...)
if err != nil {
   // handle issue
} else if !valid {
   // handle issue
}
// do something with value
```
Lots of things to be checked after every read ... even if validity was already checked before. So this might be skipped by doing it once and signalling a invalid proof as an error.

HerbertJordan: > not sure if missing nodes are also an invalid stare, or it is just the same as if an empty node is reached

A node missing that is not supposed to be an empty node (identified by a unique hash) is an invalid -- since incomplete -- proof.

kjezek: > It just makes client code look like this:
> 
> ```
> value, valid, err := proof.GetBalance(...)
> if err != nil {
>    // handle issue
> } else if !valid {
>    // handle issue
> }
> // do something with value
> ```
> 
> Lots of things to be checked after every read ... even if validity was already checked before. So this might be skipped by doing it once and signalling a invalid proof as an error.

Yes, I fully agree with this. Just I think it is how code in golang is written, i.e. perhaps if we do it this way nobody will get surprised.

kjezek: If needed we could provide a facade/wrapper on top it working in the manner you have proposed above, but the low-level access I would keep the tedious way.

go/database/mpt/proof.go
```diff
@@ -0,0 +1,164 @@
...
+// If the input address or keys are not found in the proof, it returns a proof that
+// contains only partial proof consisting only of elements that could be reached
+// from the input address and keys in the current proof.
+// If the proof cannot be fully extracted, this method returns false.
+func (p WitnessProof) ExtractAccount(root common.Hash, address common.Address, keys ...common.Key) (WitnessProof, bool) {
+	return WitnessProof{}, false
+}
+
+// IsValid checks that the proof has all valid entries.
+// It means, RLPs are valid encodings and their hashes match the key in the map.
```
HerbertJordan, line 51: These are implementation details. For the user, it should be something like:
```suggestion
// IsValid checks that this proof is self-consistent. If the result is true, the proof can be used
// for extracting verified information. If false, the proof is corrupted and should be discarded.
```

go/database/mpt/proof.go
```diff
@@ -0,0 +1,164 @@
...
+	path := AddressToNibblePath(address, nil) // TODO hash path without the node source
+	node := p.getNext(root, path)
+	if node == nil {
+		return AccountInfo{}, false, nil
+	}
+	return node.(*AccountNode).info, true, nil
+}
+
+// GetBalance extracts a balance from the witness proof for the input root hash and the address.
+// If the witness proof contains an account for the input address, it returns its balance.
```
HerbertJordan, line 70: A proof may contain multiple accounts (since it may be the result of several "Add" calls). Thus, the documentation should consider this case:
```suggestion
// If the witness proof contains the requested account for the input address for the given root hash, it returns its balance.
```

go/database/mpt/proof.go
```diff
@@ -0,0 +1,164 @@
...
+// root hash to a terminal entry according to a Path
+// representing either an account address or a storage key.
+// If the proof contains an account or a storage slot for the input address or key,
+// the proof is representing this account or storage.
+type WitnessProof proofDb
+
+// CreateWitnessProof creates a witness proof for the input account address
+// and possibly storage slots of the same account under the input storage keys.
+// If the proof cannot be created because either the address or a key does not exist,
+// it returns false.
```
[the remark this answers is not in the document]
kjezek: ok makes sense -  missing ID (=address/key) will be an error like in a database

go/database/mpt/proof.go
```diff
@@ -0,0 +1,150 @@
...
+type rlpEncodedNode []byte
+
+// proofDb is a database of RLP encoded MPT nodes and their hashes that represent witness proofs.
+type proofDb map[common.Hash]rlpEncodedNode
+
+// WitnessProof represents a witness proof.
+// It contains a database of MPT nodes and their hashes.
+// The proof may be verified by iterating the proof database from the
+// root hash to a terminal entry according to a Path
+// representing either an account address or a storage key.
```
HerbertJordan, line 29: nit: users of this type are not expected to do this. This functionality is provided by the member functions defined on this type. So I think it should not be part of the documentation.

go/database/mpt/proof.go
```diff
@@ -0,0 +1,150 @@
...
+// If the proof contains an account or a storage slot for the input address or key,
+// the proof is representing this account or storage.
+type WitnessProof proofDb
```
HerbertJordan, line 32: nit: is there are reasonable need for users of proofs to access the proof like a map?
If not, maybe hide this implementation detail like
```suggestion
type WitnessProof struct {
     proofDb
}
```

go/database/mpt/proof.go
```diff
@@ -0,0 +1,150 @@
...
+
+// CreateWitnessProof creates a witness proof for the input account address
+// and possibly storage slots of the same account under the input storage keys.
+// If the proof cannot be created because either the address or a key does not exist,
+// it returns an error.
```
HerbertJordan, line 37: Outdated: all addresses and keys always exist. They may just be empty, which is valid.

go/database/mpt/proof.go
```diff
@@ -0,0 +1,150 @@
...
+		return common.Value{}, false, nil
+	}
+	return node.(*ValueNode).value, true, nil
+}
+
+// AllStatesZero checks that all storage slots are empty for the input root hash, account address and the storage key range.
+// If the witness proof contains all empty slots for the input key range, it returns true.
+// An empty slot is a slot that contains a zero value, or does not exist at all.
+// If the proof contains a slot, it returns false.
+func (p WitnessProof) AllStatesZero(root common.Hash, address common.Address, from, to common.Key) (bool, error) {
```
HerbertJordan, line 119: This function also needs a way to say that there is not enough data in the proof to answer `true` or `false`.

go/database/mpt/proof.go
```diff
@@ -0,0 +1,150 @@
...
+	node := p.getNext(root, path)
+	if node == nil {
+		return common.Value{}, false, nil
+	}
+	return node.(*ValueNode).value, true, nil
+}
+
+// AllStatesZero checks that all storage slots are empty for the input root hash, account address and the storage key range.
+// If the witness proof contains all empty slots for the input key range, it returns true.
+// An empty slot is a slot that contains a zero value, or does not exist at all.
```
HerbertJordan, line 117: All slots `exist`. There is no such thing as a non-existing slot.

go/database/mpt/proof.go
```diff
@@ -0,0 +1,150 @@
...
+// If the proof contains a slot, it returns false.
+func (p WitnessProof) AllStatesZero(root common.Hash, address common.Address, from, to common.Key) (bool, error) {
+	return false, nil
+}
+
+// AllAddressesEmpty checks that all accounts are empty for the input root hash and the address range.
+// If the witness proof contains all empty accounts for the input address range, it returns true.
+// An empty account is an account that contains a zero balance, nonce, and code hash.
+// If the proof contains an account, it returns false.
+func (p WitnessProof) AllAddressesEmpty(root common.Hash, from, to common.Address) (bool, error) {
```
HerbertJordan, line 127: Also this function needs a way to return that there is not enough data in the proof to answer the question.

HerbertJordan (changes requested):

go/database/mpt/proof.go
kjezek: ok will remove

go/database/mpt/proof.go
kjezek: right, will update

go/database/mpt/proof.go
kjezek: ok, will remove

go/database/mpt/proof.go
kjezek: ok, will return an enum

go/database/mpt/proof.go
kjezek: ok, will return an enum

go/database/mpt/proof.go
kjezek: I think it is a good idea. This way, the clients cannot directly modify the proof by adding/removing elements in the map

HerbertJordan (dismissed):

go/database/mpt/proof.go
```diff
@@ -0,0 +1,157 @@
...
+	// WitnessProofStateContainsZeros indicates that the witness proof contains all zero
+	// slots for the input key range, or empty accounts for the range of account addresses.
+	WitnessProofStateContainsZeros WitnessProofStateResult = iota
+	// WitnessProofStateContainsNonZero indicates that the witness proof contains at least one non-zero slot
+	// for the input key range, or at least one non-empty account for the range of account addresses.
+	WitnessProofStateContainsNonZero
+	// WitnessProofIncomplete indicates that the witness proof does not contain all the necessary information,
+	// in particular not all keys or addresses are present in the proof.
+	WitnessProofIncomplete
+)
```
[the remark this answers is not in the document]
kjezek: yes, I think it is a good idea

go/database/mpt/proof.go
```diff
@@ -0,0 +1,157 @@
...
+// If the witness proof contains an account for the input address, it returns its information.
+// If the proof does not contain an account, it returns false.
+// The method may return an error if the proof is invalid.
+func (p WitnessProof) GetAccountInfo(root common.Hash, address common.Address) (AccountInfo, bool, error) {
+	path := AddressToNibblePath(address, nil) // TODO hash path without the node source
+	node := p.getNext(root, path)
+	if node == nil {
+		return AccountInfo{}, false, nil
+	}
+	return node.(*AccountNode).info, true, nil
```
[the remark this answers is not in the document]
kjezek: ok, will change

HerbertJordan (approved):
Looks good :)

PR #881, opened 2024-05-24
Codes verification

### LINKED ISSUE #736: MPT verification tools do not check the presence of referenced contract codes
The verification tools for LiveDBs and Archives do check the consistency of the MPT tree but they do not check the presence of the referenced code, nor their integrity.

The following features should be added:
- check that all referenced codes are present
- check that all codes have the expected code hash
- optional: check that no extra codes are present

petr-hanzl:
This PR adds contract codes verification.

1. Verifies that byte code and inside `codes.dat` match
2. Verifies that any `CodeHash` within a `Forest` is contained within `codes.dat`
3. Warns if there are any codes inside `codes.dat` that are not present within a `Forest`

Fixes #736

### DIFF
--- a/go/database/mpt/archive_trie.go
+++ b/go/database/mpt/archive_trie.go
@@ -76,6 +76,10 @@ func OpenArchiveTrie(directory string, config MptConfig, cacheCapacity int) (*Ar
 }
 
 func VerifyArchive(directory string, config MptConfig, observer VerificationObserver) (res error) {
+	if observer == nil {
+		observer = NilVerificationObserver{}
+	}
+
 	roots, err := loadRoots(directory + "/roots.dat")
 	if err != nil {
 		return err
@@ -83,7 +87,23 @@ func VerifyArchive(directory string, config MptConfig, observer VerificationObse
 	if len(roots) == 0 {
 		return nil
 	}
-	return VerifyMptState(directory, config, roots, observer)
+	observer.StartVerification()
+	defer func() {
+		if r := recover(); r != nil {
+			panic(r)
+		}
+		observer.EndVerification(res)
+	}()
+
+	// Open stock data structures for content verification.
+	observer.Progress("Obtaining read access to files ...")
+	source, err := openVerificationNodeSource(directory, config)
+	if err != nil {
+		return err
+	}
+	defer source.Close()
+
+	return verifyFileForest(directory, config, roots, source, observer)
 }
 
 func (a *ArchiveTrie) Add(block uint64, update common.Update, hint any) error {
--- a/go/database/mpt/archive_trie.go
+++ b/go/database/mpt/archive_trie.go
@@ -76,10 +76,6 @@ func OpenArchiveTrie(directory string, config MptConfig, cacheCapacity int) (*Ar
 }
 
 func VerifyArchive(directory string, config MptConfig, observer VerificationObserver) (res error) {
-	if observer == nil {
-		observer = NilVerificationObserver{}
-	}
-
 	roots, err := loadRoots(directory + "/roots.dat")
 	if err != nil {
 		return err
@@ -87,23 +83,7 @@ func VerifyArchive(directory string, config MptConfig, observer VerificationObse
 	if len(roots) == 0 {
 		return nil
 	}
-	observer.StartVerification()
-	defer func() {
-		if r := recover(); r != nil {
-			panic(r)
-		}
-		observer.EndVerification(res)
-	}()
-
-	// Open stock data structures for content verification.
-	observer.Progress("Obtaining read access to files ...")
-	source, err := openVerificationNodeSource(directory, config)
-	if err != nil {
-		return err
-	}
-	defer source.Close()
-
-	return verifyFileForest(directory, config, roots, source, observer)
+	return VerifyMptState(directory, config, roots, observer)
 }
 
 func (a *ArchiveTrie) Add(block uint64, update common.Update, hint any) error {
--- a/go/database/mpt/archive_trie.go
+++ b/go/database/mpt/archive_trie.go
@@ -86,7 +86,7 @@ func VerifyArchive(directory string, config MptConfig, observer VerificationObse
 	if len(roots) == 0 {
 		return nil
 	}
-	return VerifyMptState(directory, config, roots, observer)
+	return VerifyCodesAndForest(directory, config, roots, observer)
 }
 
 func (a *ArchiveTrie) Add(block uint64, update common.Update, hint any) error {
--- a/go/database/mpt/archive_trie.go
+++ b/go/database/mpt/archive_trie.go
@@ -77,7 +77,7 @@ func OpenArchiveTrie(directory string, config MptConfig, cacheCapacity int) (*Ar
 
 // VerifyArchive validates file-based archive stored in the given directory.
 // If the test passes, the data stored in the respective directory
-// can be considered to be a valid Live Trie of the given configuration.
+// can be considered a valid archive database of the given configuration.
 func VerifyArchive(directory string, config MptConfig, observer VerificationObserver) error {
 	roots, err := loadRoots(directory + "/roots.dat")
 	if err != nil {
--- a/go/database/mpt/archive_trie.go
+++ b/go/database/mpt/archive_trie.go
@@ -78,7 +78,7 @@ func OpenArchiveTrie(directory string, config MptConfig, cacheCapacity int) (*Ar
 // VerifyArchive validates file-based archive stored in the given directory.
 // If the test passes, the data stored in the respective directory
 // can be considered to be a valid Live Trie of the given configuration.
-func VerifyArchive(directory string, config MptConfig, observer VerificationObserver) (res error) {
+func VerifyArchive(directory string, config MptConfig, observer VerificationObserver) error {
 	roots, err := loadRoots(directory + "/roots.dat")
 	if err != nil {
 		return err
--- a/go/database/mpt/archive_trie.go
+++ b/go/database/mpt/archive_trie.go
@@ -75,7 +75,7 @@ func OpenArchiveTrie(directory string, config MptConfig, cacheCapacity int) (*Ar
 	}, nil
 }
 
-func VerifyArchive(directory string, config MptConfig, observer VerificationObserver) error {
+func VerifyArchive(directory string, config MptConfig, observer VerificationObserver) (res error) {
 	roots, err := loadRoots(directory + "/roots.dat")
 	if err != nil {
 		return err
@@ -83,7 +83,7 @@ func VerifyArchive(directory string, config MptConfig, observer VerificationObse
 	if len(roots) == 0 {
 		return nil
 	}
-	return VerifyFileForest(directory, config, roots, observer)
+	return VerifyMptState(directory, config, roots, observer)
 }
 
 func (a *ArchiveTrie) Add(block uint64, update common.Update, hint any) error {
--- a/go/database/mpt/archive_trie.go
+++ b/go/database/mpt/archive_trie.go
@@ -75,10 +75,10 @@ func OpenArchiveTrie(directory string, config MptConfig, cacheCapacity int) (*Ar
 	}, nil
 }
 
-// VerifyArchive validates file-based archive stored in the given directory.
+// VerifyArchiveTrie validates file-based archive stored in the given directory.
 // If the test passes, the data stored in the respective directory
 // can be considered a valid archive database of the given configuration.
-func VerifyArchive(directory string, config MptConfig, observer VerificationObserver) error {
+func VerifyArchiveTrie(directory string, config MptConfig, observer VerificationObserver) error {
 	roots, err := loadRoots(directory + "/roots.dat")
 	if err != nil {
 		return err
@@ -86,7 +86,7 @@ func VerifyArchive(directory string, config MptConfig, observer VerificationObse
 	if len(roots) == 0 {
 		return nil
 	}
-	return VerifyCodesAndForest(directory, config, roots, observer)
+	return VerifyMptState(directory, config, roots, observer)
 }
 
 func (a *ArchiveTrie) Add(block uint64, update common.Update, hint any) error {
--- a/go/database/mpt/archive_trie.go
+++ b/go/database/mpt/archive_trie.go
@@ -75,11 +75,10 @@ func OpenArchiveTrie(directory string, config MptConfig, cacheCapacity int) (*Ar
 	}, nil
 }
 
-// VerifyMptStateWithArchive validates Mpt state with file-based archive trie stored
-// in the given directory. If the test passes, the data stored in the respective
-// directory can be considered to be a valid Mpt state of the given configuration.
-// This validation contains both Forest and contract codes checks.
-func VerifyMptStateWithArchive(directory string, config MptConfig, observer VerificationObserver) error {
+// VerifyArchive validates file-based archive stored in the given directory.
+// If the test passes, the data stored in the respective directory
+// can be considered to be a valid Live Trie of the given configuration.
+func VerifyArchive(directory string, config MptConfig, observer VerificationObserver) (res error) {
 	roots, err := loadRoots(directory + "/roots.dat")
 	if err != nil {
 		return err
@@ -88,20 +87,6 @@ func VerifyMptStateWithArchive(directory string, config MptConfig, observer Veri
 		return nil
 	}
 	return VerifyMptState(directory, config, roots, observer)
-}
-
-// VerifyArchiveTrie validates a file-based archive trie stored in the given
-// directory. If the test passes, the data stored in the respective directory
-// can be considered to be a valid Live Trie of the given configuration.
-func VerifyArchiveTrie(directory string, config MptConfig, observer VerificationObserver) (res error) {
-	roots, err := loadRoots(directory + "/roots.dat")
-	if err != nil {
-		return err
-	}
-	if len(roots) == 0 {
-		return nil
-	}
-	return verifyFileForest(directory, config, roots, observer)
 }
 
 func (a *ArchiveTrie) Add(block uint64, update common.Update, hint any) error {
--- a/go/database/mpt/archive_trie.go
+++ b/go/database/mpt/archive_trie.go
@@ -75,7 +75,11 @@ func OpenArchiveTrie(directory string, config MptConfig, cacheCapacity int) (*Ar
 	}, nil
 }
 
-func VerifyArchive(directory string, config MptConfig, observer VerificationObserver) (res error) {
+// VerifyMptStateWithArchive validates Mpt state with file-based archive trie stored
+// in the given directory. If the test passes, the data stored in the respective
+// directory can be considered to be a valid Mpt state of the given configuration.
+// This validation contains both Forest and contract codes checks.
+func VerifyMptStateWithArchive(directory string, config MptConfig, observer VerificationObserver) error {
 	roots, err := loadRoots(directory + "/roots.dat")
 	if err != nil {
 		return err
@@ -84,6 +88,20 @@ func VerifyArchive(directory string, config MptConfig, observer VerificationObse
 		return nil
 	}
 	return VerifyMptState(directory, config, roots, observer)
+}
+
+// VerifyArchiveTrie validates a file-based archive trie stored in the given
+// directory. If the test passes, the data stored in the respective directory
+// can be considered to be a valid Live Trie of the given configuration.
+func VerifyArchiveTrie(directory string, config MptConfig, observer VerificationObserver) (res error) {
+	roots, err := loadRoots(directory + "/roots.dat")
+	if err != nil {
+		return err
+	}
+	if len(roots) == 0 {
+		return nil
+	}
+	return verifyFileForest(directory, config, roots, observer)
 }
 
 func (a *ArchiveTrie) Add(block uint64, update common.Update, hint any) error {
--- a/go/database/mpt/archive_trie_test.go
+++ b/go/database/mpt/archive_trie_test.go
@@ -304,7 +304,7 @@ func TestArchiveTrie_VerifyArchive_Failure_Meta(t *testing.T) {
 		t.Fatalf("cannot update roots: %v", err)
 	}
 
-	if err := VerifyArchiveTrie(dir, S5ArchiveConfig, NilVerificationObserver{}); err == nil {
+	if err := VerifyArchive(dir, S5ArchiveConfig, NilVerificationObserver{}); err == nil {
 		t.Errorf("verification should fail")
 	}
 }
@@ -389,7 +389,7 @@ func TestArchiveTrie_CanProcessPrecomputedHashes(t *testing.T) {
 				t.Fatalf("failed to close resources: %v", err)
 			}
 
-			if err := VerifyArchiveTrie(archiveDir, config, NilVerificationObserver{}); err != nil {
+			if err := VerifyArchive(archiveDir, config, NilVerificationObserver{}); err != nil {
 				t.Errorf("failed to verify archive: %v", err)
 			}
 		})
@@ -400,7 +400,7 @@ func TestArchiveTrie_VerificationOfEmptyDirectoryPasses(t *testing.T) {
 	for _, config := range allMptConfigs {
 		t.Run(config.Name, func(t *testing.T) {
 			dir := t.TempDir()
-			if err := VerifyArchiveTrie(dir, config, NilVerificationObserver{}); err != nil {
+			if err := VerifyArchive(dir, config, NilVerificationObserver{}); err != nil {
 				t.Errorf("an empty directory should be fine, got: %v", err)
 			}
 		})
@@ -450,7 +450,7 @@ func TestArchiveTrie_VerificationOfFreshArchivePasses(t *testing.T) {
 				t.Fatalf("failed to close archive: %v", err)
 			}
 
-			if err := VerifyArchiveTrie(dir, config, NilVerificationObserver{}); err != nil {
+			if err := VerifyArchive(dir, config, NilVerificationObserver{}); err != nil {
 				t.Errorf("a freshly closed archive should be fine, got: %v", err)
 			}
 		})
@@ -1081,7 +1081,7 @@ func TestArchiveTrie_VerificationOfArchiveWithMissingFileFails(t *testing.T) {
 				t.Fatalf("failed to delete file")
 			}
 
-			if err := VerifyArchiveTrie(dir, config, NilVerificationObserver{}); err == nil {
+			if err := VerifyArchive(dir, config, NilVerificationObserver{}); err == nil {
 				t.Errorf("missing file should be detected")
 			}
 		})
@@ -1128,7 +1128,7 @@ func TestArchiveTrie_VerificationOfArchiveWithCorruptedFileFails(t *testing.T) {
 				t.Fatalf("failed to modify file")
 			}
 
-			if err := VerifyArchiveTrie(dir, config, NilVerificationObserver{}); err == nil {
+			if err := VerifyArchive(dir, config, NilVerificationObserver{}); err == nil {
 				t.Errorf("corrupted file should have been detected")
 			}
 		})
--- a/go/database/mpt/archive_trie_test.go
+++ b/go/database/mpt/archive_trie_test.go
@@ -389,11 +389,6 @@ func TestArchiveTrie_CanProcessPrecomputedHashes(t *testing.T) {
 				t.Fatalf("failed to close resources: %v", err)
 			}
 
-			// create dummy codes.dat file
-			if err := writeCodes(nil, filepath.Join(archiveDir, "codes.dat")); err != nil {
-				t.Fatalf("failed to write dummy code file: %v", err)
-			}
-
 			if err := VerifyArchive(archiveDir, config, NilVerificationObserver{}); err != nil {
 				t.Errorf("failed to verify archive: %v", err)
 			}
@@ -453,11 +448,6 @@ func TestArchiveTrie_VerificationOfFreshArchivePasses(t *testing.T) {
 
 			if err := archive.Close(); err != nil {
 				t.Fatalf("failed to close archive: %v", err)
-			}
-
-			// create dummy codes.dat file
-			if err := writeCodes(nil, filepath.Join(dir, "codes.dat")); err != nil {
-				t.Fatalf("failed to write dummy code file: %v", err)
 			}
 
 			if err := VerifyArchive(dir, config, NilVerificationObserver{}); err != nil {
--- a/go/database/mpt/archive_trie_test.go
+++ b/go/database/mpt/archive_trie_test.go
@@ -389,6 +389,11 @@ func TestArchiveTrie_CanProcessPrecomputedHashes(t *testing.T) {
 				t.Fatalf("failed to close resources: %v", err)
 			}
 
+			// create dummy codes.dat file
+			if err := writeCodes(nil, filepath.Join(archiveDir, "codes.dat")); err != nil {
+				t.Fatalf("failed to write dummy code file: %v", err)
+			}
+
 			if err := VerifyArchive(archiveDir, config, NilVerificationObserver{}); err != nil {
 				t.Errorf("failed to verify archive: %v", err)
 			}
@@ -448,6 +453,11 @@ func TestArchiveTrie_VerificationOfFreshArchivePasses(t *testing.T) {
 
 			if err := archive.Close(); err != nil {
 				t.Fatalf("failed to close archive: %v", err)
+			}
+
+			// create dummy codes.dat file
+			if err := writeCodes(nil, filepath.Join(dir, "codes.dat")); err != nil {
+				t.Fatalf("failed to write dummy code file: %v", err)
 			}
 
 			if err := VerifyArchive(dir, config, NilVerificationObserver{}); err != nil {
--- a/go/database/mpt/archive_trie_test.go
+++ b/go/database/mpt/archive_trie_test.go
@@ -15,8 +15,6 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
-	"github.com/Fantom-foundation/Carmen/go/backend/archive"
-	"golang.org/x/exp/maps"
 	"math/rand"
 	"os"
 	"path/filepath"
@@ -25,6 +23,9 @@ import (
 	"strings"
 	"sync"
 	"testing"
+
+	"github.com/Fantom-foundation/Carmen/go/backend/archive"
+	"golang.org/x/exp/maps"
 
 	"go.uber.org/mock/gomock"
 
@@ -303,7 +304,7 @@ func TestArchiveTrie_VerifyArchive_Failure_Meta(t *testing.T) {
 		t.Fatalf("cannot update roots: %v", err)
 	}
 
-	if err := VerifyArchive(dir, S5ArchiveConfig, NilVerificationObserver{}); err == nil {
+	if err := VerifyArchiveTrie(dir, S5ArchiveConfig, NilVerificationObserver{}); err == nil {
 		t.Errorf("verification should fail")
 	}
 }
@@ -388,7 +389,7 @@ func TestArchiveTrie_CanProcessPrecomputedHashes(t *testing.T) {
 				t.Fatalf("failed to close resources: %v", err)
 			}
 
-			if err := VerifyArchive(archiveDir, config, NilVerificationObserver{}); err != nil {
+			if err := VerifyArchiveTrie(archiveDir, config, NilVerificationObserver{}); err != nil {
 				t.Errorf("failed to verify archive: %v", err)
 			}
 		})
@@ -399,7 +400,7 @@ func TestArchiveTrie_VerificationOfEmptyDirectoryPasses(t *testing.T) {
 	for _, config := range allMptConfigs {
 		t.Run(config.Name, func(t *testing.T) {
 			dir := t.TempDir()
-			if err := VerifyArchive(dir, config, NilVerificationObserver{}); err != nil {
+			if err := VerifyArchiveTrie(dir, config, NilVerificationObserver{}); err != nil {
 				t.Errorf("an empty directory should be fine, got: %v", err)
 			}
 		})
@@ -449,7 +450,7 @@ func TestArchiveTrie_VerificationOfFreshArchivePasses(t *testing.T) {
 				t.Fatalf("failed to close archive: %v", err)
 			}
 
-			if err := VerifyArchive(dir, config, NilVerificationObserver{}); err != nil {
+			if err := VerifyArchiveTrie(dir, config, NilVerificationObserver{}); err != nil {
 				t.Errorf("a freshly closed archive should be fine, got: %v", err)
 			}
 		})
@@ -1080,7 +1081,7 @@ func TestArchiveTrie_VerificationOfArchiveWithMissingFileFails(t *testing.T) {
 				t.Fatalf("failed to delete file")
 			}
 
-			if err := VerifyArchive(dir, config, NilVerificationObserver{}); err == nil {
+			if err := VerifyArchiveTrie(dir, config, NilVerificationObserver{}); err == nil {
 				t.Errorf("missing file should be detected")
 			}
 		})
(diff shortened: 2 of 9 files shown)


go/database/mpt/verification_test.go
```diff
@@ -407,11 +409,93 @@ func TestVerification_ValueNodeHashModificationIsDetected(t *testing.T) {
...
+		}
+
+		encoder, _, _, _ := getEncoder(config)
+
+		modifyNode(t, dir+"/accounts", encoder, func(node *AccountNode) {
+			node.info.CodeHash = testHash
+		})
+
+		if err := VerifyMptState(dir, config, roots, NilVerificationObserver{}); err == nil {
+			t.Errorf("Different hash in code file should have been detected")
```
kjezek, line 464:
```suggestion
			t.Errorf("different hash in code file should have been detected")
```

kjezek (changes requested):
Thanks for this PR. I have two concerts:

1. I do not understand intention of code that recovers from panic just to panic at the next line. It looks like a redundant check to me.
2.  I do not understand how the new `VerifyMptState` method is used.  `VerifyArchive` and `VerifyFileLiveTrie` are both called from the verification CLI tool, while the newly created method is not.

I guess it should have been implemented that the two already existing methods do call the contract check and then forward to `verifyFileForest`, the new method is redundant then.

kjezek:
The tests seem to be repetitive just changing the input. In this case, it should use: https://go.dev/wiki/TableDrivenTests

go/database/mpt/live_trie.go
```diff
@@ -63,18 +63,38 @@ func OpenFileLiveTrie(directory string, config MptConfig, cacheCapacity int) (*L
...
+		observer.EndVerification(res)
+	}()
+
+	// Open stock data structures for content verification.
+	observer.Progress("Obtaining read access to files ...")
+	source, err := openVerificationNodeSource(directory, config)
+	if err != nil {
+		return err
+	}
+	defer source.Close()
```
HerbertJordan, line 92: This code is duplicated in the Archive and LiveTrie implementation.

I would suggest to rename the current `verifyFileForest` accepting a `*verificationNodeSource` to `verifyForest` and create a new `verifyFileForest` function that handles the source opening and closing and calls the `verifyForest` function for the actual checks.

go/database/mpt/verification.go
```diff
@@ -41,15 +42,11 @@ func (NilVerificationObserver) StartVerification()        {}
 func (NilVerificationObserver) Progress(msg string)       {}
 func (NilVerificationObserver) EndVerification(res error) {}
 
-// VerifyFileForest runs list of validation checks on the forest stored in the given
-// directory. These checks include:
-//   - all required files are present and can be read
-//   - all referenced nodes are present
-//   - all hashes are consistent
-func VerifyFileForest(directory string, config MptConfig, roots []Root, observer VerificationObserver) (res error) {
+func VerifyMptState(directory string, config MptConfig, roots []Root, observer VerificationObserver) (res error) {
```
HerbertJordan, line 45: Please document, as this function seems to be intended for package-external use. If not, make it private.

go/database/mpt/verification.go
```diff
@@ -58,6 +55,33 @@ func VerifyFileForest(directory string, config MptConfig, roots []Root, observer
 		observer.EndVerification(res)
 	}()
 
+	// Open stock data structures for content verification.
+	observer.Progress("Obtaining read access to files ...")
+	source, err := openVerificationNodeSource(directory, config)
+	if err != nil {
+		return err
+	}
+	defer source.Close()
```
HerbertJordan, line 64: This looks like a third copy of the code in the Archive and Live DB ... => to be unified.

go/database/mpt/verification.go
```diff
@@ -570,6 +586,73 @@ func verifyHashesStoredWithParents[N any](
...
+		}
+		// check that the code hash is present in the code file
+		byteCode, exists := codes[accountHash]
+		if !exists {
+			return fmt.Errorf("the hash %x is present in the mpt but missing from the code file", accountHash)
+		}
+		// check correctness of the code hash
+		if got, want := common.Keccak256(byteCode), &accountHash; got.Compare(want) != 0 {
+			return fmt.Errorf("unexpected code hash for address, got: %x want: %x", got, want)
+		}
```
HerbertJordan, line 624: This should be done outside the Account Node iteration. If a code is used 100x, the same code is checked 100 times.

go/database/mpt/verification.go
```diff
@@ -570,6 +586,73 @@ func verifyHashesStoredWithParents[N any](
...
+	if len(found) == len(codes) {
+		return nil
+	}
+
+	observer.Progress(fmt.Sprintf("There are %d contracts not referenced by any account:", len(codes)-len(found)))
+
+	// find any extra hashes
+	for h, bc := range codes {
+		_, exists := found[h]
+		if !exists {
```
HerbertJordan, line 645: nit:
```suggestion
		if _, exists := found[h]; !exists {
```

go/database/mpt/verification_test.go
```diff
@@ -41,9 +43,9 @@ var forestFiles = []string{
 	"values/values.dat",
 }
 
-func TestVerification_VerifyValidForest(t *testing.T) {
+func TestVerification_VerifyValidMpt(t *testing.T) {
```
HerbertJordan, line 46:
```suggestion
func TestVerification_VerifyValidMptState(t *testing.T) {
```

go/database/mpt/archive_trie.go
```diff
@@ -75,15 +75,35 @@ func OpenArchiveTrie(directory string, config MptConfig, cacheCapacity int) (*Ar
...
 	if err != nil {
 		return err
 	}
 	if len(roots) == 0 {
 		return nil
 	}
-	return VerifyFileForest(directory, config, roots, observer)
+	observer.StartVerification()
+	defer func() {
+		if r := recover(); r != nil {
```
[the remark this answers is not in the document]
petr-hanzl: This code is even in the [current version](https://github.com/Fantom-foundation/Carmen/blob/842587ea653b1fdbe8d871b40a3a9d127b8a7ffa/go/database/mpt/verification.go#L55) of verification. Its just moved to the top-most func.

go/database/mpt/verification_test.go
```diff
@@ -407,11 +409,93 @@ func TestVerification_ValueNodeHashModificationIsDetected(t *testing.T) {
...
 		})
 
-		if err := VerifyFileForest(dir, config, roots, NilVerificationObserver{}); err == nil {
+		if err := VerifyMptState(dir, config, roots, NilVerificationObserver{}); err == nil {
 			t.Errorf("Modified node should have been detected")
 		}
 	})
 }
+
+func TestVerification_CodeHashNotContainedInMptIsDetected(t *testing.T) {
```
HerbertJordan, line 418: Please fix the name ... this test is not doing what the name indicates.

Please also check all other names. Consider that Mpt != MptState.

HerbertJordan (changes requested):

go/database/mpt/archive_trie.go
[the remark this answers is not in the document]
kjezek: fix it then!  The overall effort should be to reduce technical debt, not spreading it to 3 more places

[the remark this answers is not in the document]
petr-hanzl: I see what you mean now. This was caused by calling incorrect func. Thanks for the heads up!

go/database/mpt/verification.go
petr-hanzl: I implemented an early exit. So we dont need to create a second loop.

go/database/mpt/live_trie.go
petr-hanzl: This was caused by calling `verifyFileForest` instead of `VerifyMptState`. This duplicity got removed by calling the correct func.

go/database/mpt/archive_trie.go
```diff
@@ -75,15 +75,33 @@ func OpenArchiveTrie(directory string, config MptConfig, cacheCapacity int) (*Ar
 	}, nil
 }
 
-func VerifyArchive(directory string, config MptConfig, observer VerificationObserver) error {
```
kjezek, line 78: I think this method should have been extended with `codes` verification, otherwise untouched. The other method(s) I mention should not have been introduced. It makes the design somehow not easy to follow.

go/database/mpt/live_trie.go
```diff
@@ -60,18 +60,36 @@ func OpenFileLiveTrie(directory string, config MptConfig, cacheCapacity int) (*L
 	return makeTrie(directory, forest)
 }
 
+// VerifyMptStateWithLiveTrie validates Mpt state with file-based live trie stored
+// in the given directory. If the test passes, the data stored in the respective
+// directory can be considered to be a valid Mpt state of the given configuration.
+// This validation contains both Forest and contract codes checks.
+func VerifyMptStateWithLiveTrie(directory string, config MptConfig, observer VerificationObserver) error {
```
kjezek, line 67: ... also here, it looks like the name just chains all terms we have into a method name

kjezek (changes requested):
can you please simplify (mostly revert back) method names? 

We used to have this two use-cases:
* CLI tool -> VerifyArchive -> VerifyFileForest
* CLI tool -> VerifyFileLiveTrie -> VerifyFileForest

I think the task was just add one step in between:
* CLI tool -> VerifyArchive -> {verifyCodes, verifyFileForest}
* CLI tool -> VerifyFileLiveTrie ->  {verifyCodes, verifyFileForest}

all renames introduced in this PR makes the design just hard to follow. 


To modify my comment to be aligned with the comment you mention, the call chain should be IMHO this:

* CLI tool -> VerifyArchive -> verifyMptState -> {the new code check, verifyFileForest}
* CLI tool -> VerifyFileLiveTrie -> verifyMptState -> {the new code check, verifyFileFoerst}

as far as I can see the latest commits, it seems it is done this way - i.e. its correct. 
just may tests were changed to test the MptState instead of the forest - I think if the tests still test the same as they did before, there is not reason to change them.

go/database/mpt/archive_trie.go
```diff
@@ -75,15 +75,18 @@ func OpenArchiveTrie(directory string, config MptConfig, cacheCapacity int) (*Ar
...
+// VerifyArchive validates file-based archive stored in the given directory.
+// If the test passes, the data stored in the respective directory
+// can be considered to be a valid Live Trie of the given configuration.
+func VerifyArchive(directory string, config MptConfig, observer VerificationObserver) (res error) {
```
kjezek, line 81:
```suggestion
func VerifyArchive(directory string, config MptConfig, observer VerificationObserver) error {
```

go/database/mpt/archive_trie_test.go
```diff
@@ -388,6 +389,11 @@ func TestArchiveTrie_CanProcessPrecomputedHashes(t *testing.T) {
 				t.Fatalf("failed to close resources: %v", err)
 			}
 
+			// create dummy codes.dat file
+			if err := writeCodes(nil, filepath.Join(archiveDir, "codes.dat")); err != nil {
```
kjezek, line 393: if it is expected the file must exists, it is perhaps a verification error if the file is not present.
Alternatively we have two options:
* create the file when the state is created and the file does not exist
* accept the missing file as an ok state, and read this as empty

go/database/mpt/live_trie.go
```diff
@@ -60,9 +60,10 @@ func OpenFileLiveTrie(directory string, config MptConfig, cacheCapacity int) (*L
 	return makeTrie(directory, forest)
 }
 
-// VerifyFileLiveTrie validates a file-based live trie stored in the given
-// directory. If the test passes, the data stored in the respective directory
-// can be considered to be a valid Live Trie of the given configuration.
+// VerifyFileLiveTrie validates Mpt state with file-based live trie stored
```
kjezek, line 63:
```suggestion
// VerifyFileLiveTrie validates a file-based live trie stored
```

kjezek: I think the previous comment was ok - it is still valid after your changes. We do not have to mention every intermediate layer we have added.

go/database/mpt/archive_trie_test.go
```diff
@@ -449,6 +455,11 @@ func TestArchiveTrie_VerificationOfFreshArchivePasses(t *testing.T) {
 				t.Fatalf("failed to close archive: %v", err)
 			}
 
+			// create dummy codes.dat file
+			if err := writeCodes(nil, filepath.Join(dir, "codes.dat")); err != nil {
```
kjezek, line 459: the same as above

go/database/mpt/live_trie_test.go
```diff
@@ -797,6 +797,11 @@ func TestLiveTrie_VerificationOfFreshArchivePasses(t *testing.T) {
 				t.Fatalf("failed to close trie: %v", err)
 			}
 
+			// create dummy codes.dat file
+			if err := writeCodes(nil, filepath.Join(dir, "codes.dat")); err != nil {
```
kjezek, line 801: the same as above

go/database/mpt/verification_test.go
```diff
@@ -41,9 +43,9 @@ var forestFiles = []string{
 	"values/values.dat",
 }
 
-func TestVerification_VerifyValidForest(t *testing.T) {
+func TestVerification_VerifyValidMptState(t *testing.T) {
```
kjezek, line 46: here and below it seems many tests were changed to test one layer above the forest, but I think it is an unrelated change, because it did not change what is tested.

go/database/mpt/archive_trie_test.go
petr-hanzl: There is a [test](https://github.com/Fantom-foundation/Carmen/blob/67008e3801f9718da0284207e90d72b35ca294e5/go/database/mpt/verification_test.go#L70C6-L70C44) in the verification that explicitly asks for fail when the file does not exist. So creating the file is the logical option here.

kjezek: this is a bit different situation, the test you mention checks for forest files. Without these files, the forest cannot be loaded. 

But the MPT state without the codes file can be loaded. In this case an empty map is expected:
see https://github.com/Fantom-foundation/Carmen/blob/main/go/database/mpt/state.go#L472

for this reason I think verification should not fail on a missing file, and then you do not have to create the file here.

petr-hanzl: Ic. I thought I should apply the logic from the test I mentioned to code file as well. Your comment make sense. Thanks!

go/database/mpt/verification.go
```diff
@@ -570,6 +622,72 @@ func verifyHashesStoredWithParents[N any](
...
+		byteCode, exists := codes[codeHash]
+		if !exists {
+			return fmt.Errorf("the hash %x is present in the mpt but missing from the code file", codeHash)
+		}
+		// check correctness of the code hash
+		if got, want := common.Keccak256(byteCode), &codeHash; got.Compare(want) != 0 {
+			return fmt.Errorf("unexpected code hash for address, got: %x want: %x", got, want)
+		}
+		// mark a found hash
+		found[codeHash] = true
```
kjezek, line 663: I think this should go before the `if` condition - if there is a corrupted contract used in many accounts, it would print it many times now

go/database/mpt/verification.go
```diff
@@ -570,6 +622,72 @@ func verifyHashesStoredWithParents[N any](
...
+		return nil
+	}
+
+	observer.Progress(fmt.Sprintf("There are %d contracts not referenced by any accounts:", len(codes)-len(found)))
+
+	// find any extra hashes
+	for h, bc := range codes {
+		if _, exists := found[h]; !exists {
+			observer.Progress(fmt.Sprintf("%x\n", h))
+			if got, want := common.Keccak256(bc), &h; got.Compare(want) != 0 {
```
kjezek, line 682: not sure if we need to verify codes that are not used

go/database/mpt/verification_test.go
```diff
@@ -43,7 +44,7 @@ var forestFiles = []string{
 
 func TestVerification_VerifyValidForest(t *testing.T) {
 	runVerificationTest(t, func(t *testing.T, dir string, config MptConfig, roots []Root) {
-		if err := VerifyFileForest(dir, config, roots, NilVerificationObserver{}); err != nil {
+		if err := VerifyMptState(dir, config, roots, NilVerificationObserver{}); err != nil {
```
kjezek, line 47: I do not think these tests should be all changed to test one layer above the forest when they actually test still the same as they used to test before.

go/database/mpt/verification.go
```diff
@@ -41,15 +42,63 @@ func (NilVerificationObserver) StartVerification()        {}
...
+	return nil
+}
+
+// verifyFileForest runs list of validation checks on the forest stored in the given
 // directory. These checks include:
 //   - all required files are present and can be read
 //   - all referenced nodes are present
 //   - all hashes are consistent
-func VerifyFileForest(directory string, config MptConfig, roots []Root, observer VerificationObserver) (res error) {
+func verifyFileForest(directory string, config MptConfig, roots []Root, observer VerificationObserver) (res error) {
```
kjezek, line 97: what is actually reason to change this method to private? It seems unrelated to work of this PR.

kjezek (changes requested):
Thanks for updates, I think the code is getting the shape. Please have a look at my further comments. 

Also please make sure the test coverage is 100% for new changes:

image

image

go/database/mpt/verification_test.go
kjezek: I do not think this was resolved, there is still many tests that got changed, but it seems unrelated.

go/database/mpt/verification.go
kjezek: well, there was no reason even before, because it was never called outside of this package. We can talk if it should have been private already, but I prefer if PRs really change only what is related, and potentially reduction of technical debt is done in a separate PR.

go/database/mpt/verification.go
```diff
@@ -41,15 +42,63 @@ func (NilVerificationObserver) StartVerification()        {}
...
+		return err
+	}
+	defer source.Close()
+
+	err = verifyContractCodes(directory, source, observer)
+	if err != nil {
+		return err
+	}
+
+	err = verifyForest(directory, config, roots, source, observer)
```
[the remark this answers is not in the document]
petr-hanzl: I strongly believe that creates code duplicity between `VerifyArchive` and `VerifyFileLiveTrie`

go/database/mpt/verification.go
petr-hanzl: Since it return an error, the verification is fatal hence it ends no?

go/database/mpt/verification.go
petr-hanzl: I think we should still check the hash matches the contract byte-codes. It could potentionaly tell us more...

go/database/mpt/verification_test.go
petr-hanzl: I wanted to include the whole verification but I guess you are right. Should we the tests be renamed to `TestForestVerification...` since we have now multiple verifications and they only test one part of it?

go/database/mpt/verification.go
```diff
@@ -41,15 +42,63 @@ func (NilVerificationObserver) StartVerification()        {}
...
+//  2. Non-fatal checks
+//     - there are no extra Code Hashes not referenced by any account
+func VerifyMptState(directory string, config MptConfig, roots []Root, observer VerificationObserver) (res error) {
+	if observer == nil {
+		observer = NilVerificationObserver{}
+	}
+
+	observer.StartVerification()
+	defer func() {
+		if r := recover(); r != nil {
```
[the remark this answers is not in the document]
petr-hanzl: Yeah, you are right. It does not make sense to a `recover` here. Lets keep the `EndVerification()` only.

go/database/mpt/verification.go
kjezek: I think it iterates over all accounts first, and only then accumulate errors, here:

https://github.com/Fantom-foundation/Carmen/blob/petr/verify-codes/go/database/mpt/verification.go#L900-L902

petr-hanzl: You are right. Good catch.

go/database/mpt/verification_test.go
```diff
@@ -412,6 +414,154 @@ func TestVerification_ValueNodeHashModificationIsDetected(t *testing.T) {
...
+			testHash: {1},
+		}
+		if err := writeCodes(codes, filepath.Join(dir, "codes.dat")); err != nil {
+			t.Fatalf("failed to write code file")
+		}
+
+		encoder, _, _, _ := getEncoder(config)
+
+		// modify two different nodes to append same hash to two accounts
+		modifyFirstNode(t, filepath.Join(dir, "accounts"), encoder, func(node *AccountNode) {
```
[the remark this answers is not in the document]
petr-hanzl: I have always created tests this way. Imo it removes redundancy and it also applies a clear thought how the data is incorrect right in the test.
If you still feel I should create a new data set, even tho its only for one test I will do it.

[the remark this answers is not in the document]
kjezek: I agree with what you are saying, but I do not think it is what is actually happening in the tests.  It is alright to have a correct data and then corrupt them on purpose to mimic an erroneous state. 

But here it is:  all test accounts have empty code-hash. It is not because it represents an error-free state, it is just a coincidence that these code-hashes were not used in tests before. 

you now used these tests and inserted code-hashes in the tests. This I think is wrong, the test data should be created ready to use. I think that the hashes should be created right where the data are created. I guess you do not need to create a separate test data, just insert the code hashes where the test data are created. You can have a list of test byte-code sequences and their hashes initialised in advance. 

finally you create codes.dat, where you insert data that do not match accounts. This is where you do on purpose manipulation with data to create an erroneous state - and this is a correct way. 

if you do it this way, you can revert all the newly introduced methods to modify first/last node, etc. 

btw: it is happening I think 3-times, not only in this test.

go/database/mpt/verification_test.go
```diff
@@ -570,3 +746,12 @@ func getFirstElementInSet(set stock.IndexSet[uint64]) (uint64, bool) {
 	}
 	return 0, false
 }
+
+func getLastElementInSet(set stock.IndexSet[uint64]) (uint64, bool) {
+	for i := set.GetUpperBound(); i < set.GetLowerBound(); i++ {
```
kjezek, line 751:
```suggestion
	for i := set.GetUpperBound(); i >= set.GetLowerBound(); i-- {
```

go/database/mpt/verification_test.go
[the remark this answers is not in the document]
petr-hanzl: I see, you dont like appending `emptyHash` which actually skips the validation and then corrupting the data. Make sense. Thanks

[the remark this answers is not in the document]
kjezek: Please have a look at this commit: https://github.com/Fantom-foundation/Carmen/pull/881/commits/33c354abe903fa6afa3efe5525a0ad85e3400d51 

I hope it explains better what I meant. 

It breaks one test, but this test was somehow unclear, because I do not see how does it test that a code is not checked twice. Can you please have a look and fix?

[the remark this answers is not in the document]
petr-hanzl: Thanks for the help. I was implementing something very similar.

go/database/mpt/verification.go
```diff
@@ -50,14 +93,21 @@ func VerifyFileForest(directory string, config MptConfig, roots []Root, observer
...
 	}
+
 	observer.StartVerification()
-	defer func() {
-		if r := recover(); r != nil {
-			panic(r)
-		}
-		observer.EndVerification(res)
-	}()
+	defer observer.EndVerification(res)
```
kjezek, line 98:
```suggestion
	defer observer.EndVerification(nil)
```

go/database/mpt/verification.go
```diff
@@ -570,6 +612,67 @@ func verifyHashesStoredWithParents[N any](
...
+		if _, exists := checkedHashes[codeHash]; exists {
+			return nil
+		}
+		// check that the code hash is present in the code file
+		byteCode, exists := codes[codeHash]
+		if !exists {
+			return fmt.Errorf("hash %x is missing in code file", codeHash)
+		}
+		// mark an already checked hash
+		checkedHashes[codeHash] = true
```
kjezek, line 648: I think it should be moved up even before the previous if - if there is a contract missing in the codes, and this contract is in many accounts, it will check the same account many times

go/database/mpt/verification_test.go
```diff
@@ -412,6 +414,118 @@ func TestVerification_ValueNodeHashModificationIsDetected(t *testing.T) {
...
+		observer := NewMockVerificationObserver(ctrl)
+
+		testHash1 := common.Keccak256([]byte{1})
+		testHash2 := common.Keccak256([]byte{2})
+		codes := map[common.Hash][]byte{
+			testHash1: {1},
+			testHash2: {2},
+		}
+
+		// TODO - this is fragile, it depends on precise strings
```
kjezek, line 454: I thought you resolve this todo :)

I think it would be better to let the observer pass through all strings and just make sure certain substring was passed.

kjezek (dismissed):
I think it looks great now. I added just two more comments. 

Did you check that the test coverage is ok?

go/database/mpt/verification.go
petr-hanzl: I think the `res` here should remain as this is the reason why the interface accepts an error. If we want to remove it from here I suggest we remove the parameter all together.

go/database/mpt/verification.go
petr-hanzl: I thought about it, but it would break the check for extra hash since the count would be incorrect.

go/database/mpt/verification.go
kjezek: it looked to me that `res` is always `nil`. Is it correct? If yes, I do not see reason to pass a variable that is always nil, it only creates confusion.

go/database/mpt/verification.go
kjezek: you can utilise the fact that you store `bool` in the map.  You can store `true` when the code hash was checked and exists and `false` when it was checked but does not exist.

petr-hanzl: Good idea

petr-hanzl: It actually does not mess up the verification itself. Just the count. The files codes are iterated, not the account codes.

go/database/mpt/verification.go
petr-hanzl: I dont think so. Any error return assigns the returned error to `res`

go/database/mpt/verification_test.go
```diff
@@ -412,6 +415,116 @@ func TestVerification_ValueNodeHashModificationIsDetected(t *testing.T) {
...
+		codes := map[common.Hash][]byte{
+			testHash1: {1},
+			testHash2: {2},
+		}
+
+		gomock.InOrder(
+			observer.EXPECT().StartVerification(),
+			observer.EXPECT().Progress(WithSubstring("Obtaining read access to files")),
+			observer.EXPECT().Progress(WithSubstring("Checking contract codes")),
+			observer.EXPECT().Progress(WithSubstring(fmt.Sprintf("Contract %x is not referenced", testHash2))),
```
kjezek, line 459: Is not it that you are in reality looking just for this substring: "Contract %x is not referenced"?  Why mocking all other calls? I think the newly created matcher has a little effect, because the input are still full sentence, i.e. it solves only situation when a message is extended with some extra words. Is this likely?

Why not do it this simple way? 
```

		var triggered bool
		observer.EXPECT().StartVerification()
		observer.EXPECT().EndVerification(nil)
		observer.EXPECT().Progress(gomock.Any()).Do(func(msg string) {
			if strings.Contains(msg, fmt.Sprintf("Contract %x is not referenced", testHash2)) {
				triggered = true
			}
		}).AnyTimes()
...
                 if !triggered {
			t.Errorf("extra hash in code file should have been detected")
		}

```

I think it would serve the same purpose and its much simpler and easy to follow

go/database/mpt/verification.go
```diff
@@ -570,6 +612,61 @@ func verifyHashesStoredWithParents[N any](
...
+
+	checkedHashes := make(map[common.Hash]bool)
+	check := func(acc *AccountNode) error {
+		codeHash := acc.info.CodeHash
+		// skip accounts that are not contracts
+		if codeHash == emptyCodeHash {
+			return nil
+		}
+		// no need to check the hash correctness more than once
+		if _, isChecked := checkedHashes[codeHash]; isChecked {
```
kjezek, line 639:
```suggestion
		if _, exists := checkedHashes[codeHash]; exists {
```

kjezek: I think it is standard to use either comma-ok or comma-exists idioms. It is not necessary to be too creative in naming

go/database/mpt/verification.go
kjezek: but there is no error assigned to `res` - i.e. it is always nil.  For this reason I think it is rather confusing because it looks like an error is propagated, but it will never happen

go/database/mpt/verification.go
```diff
@@ -570,6 +612,61 @@ func verifyHashesStoredWithParents[N any](
...
+		checkedHashes[codeHash] = true
+		// check that the code hash is present in the code file
+		byteCode, exists := codes[codeHash]
+		if !exists {
+			return fmt.Errorf("hash %x is missing in code file", codeHash)
+		}
+		// check correctness of the code hash
+		if got, want := common.Keccak256(byteCode), &codeHash; got.Compare(want) != 0 {
+			return fmt.Errorf("unexpected code hash for address, got: %x want: %x", got, want)
+		}
```
HerbertJordan, line 652: I would suggest to do this check once, before walking through the accounts, for all codes in the `codes` map loaded from the file. This simplifies the code structure and you can get rid of the `checkedHashes` map.

This would also eliminate the need for checking the hashes of codes not referenced by accounts below.

go/database/mpt/live_trie.go
```diff
@@ -71,7 +71,7 @@ func VerifyFileLiveTrie(directory string, config MptConfig, observer Verificatio
 	if !exists {
 		return nil
 	}
-	return VerifyFileForest(directory, config, []Root{{
+	return VerifyCodesAndForest(directory, config, []Root{{
```
HerbertJordan, line 74: I still think that this function should not check codes since LiveTrie instances are not managing codes, just the MPT structure. Codes are managed by `MptState` instances, and there should be a validation for those.

However, I guess there have been some discussions on this topic and this is the solution you got to. It will work, since right now there is no MPT without a Code store. We just need to be aware that if we ever have such a case, the verification will fail.

go/database/mpt/verification_test.go
```diff
@@ -570,3 +683,21 @@ func getFirstElementInSet(set stock.IndexSet[uint64]) (uint64, bool) {
 	}
 	return 0, false
 }
+
+// WithSubstring matches executor.State instances with the given substate.
+func WithSubstring(substr string) gomock.Matcher {
```
HerbertJordan, line 688: `WithSubstring` => `HasSubstring` ?

HerbertJordan (changes requested):

go/database/mpt/live_trie.go
kjezek: I was proposing that each method `VerifyFileLiveTrie` and `VerifyArchive` should call first a new method that verifies the codes (could be called `VerifyMptState`?) and then the original `VerifyFileForest`.  It would decompose these two verifications into two independent steps.

go/database/mpt/verification.go
petr-hanzl: I believe it would remove check for any extra hash hence removing reporting it.

petr-hanzl: If we were to do it before iterating accounts I think it would add another loop which is not necessary

go/database/mpt/live_trie.go
kjezek: I am confused about it, because `VerifyFileLiveTrie` verifies the LiveDB (perhaps the method should be better named `VerifyLiveDB`) and it does contain codes, i.e. it should be verified as well.  It is like one instance of `MptState` in the case of live db while for the archive there is conceptually many `MptState` instances. On the other hand there is only one codes file in both cases, i.e. in both cases verify one instance of `MptState`.

go/database/mpt/verification.go
kjezek: where is this test? I checked out this branch and could not find it. Anyway, when I copy and run it, it fails because of:
```

 verification.go:107: Unexpected call to *mpt.MockVerificationObserver.EndVerification([<nil>]) at /Users/kjezek/fantom/Carmen/go/database/mpt/verification.go:107 because: 
        expected call at /Users/kjezek/fantom/Carmen/go/database/mpt/verification_test.go:726 doesn't match the argument at index 0.
        Got: <nil> (<nil>)
        Want: not(is equal to <nil> (<nil>))

```

i.e. it proves the opposite

petr-hanzl: So when the defer func is called this way
`defer observer.EndVerification(res)` - the test fails - its passed at the very beginning
but when its called this way
```
defer func(){
   defer observer.EndVerification(res)
}
```
 the test passes - its passed at the end when it gets assigned with `return err`

petr-hanzl: [Here](https://stackoverflow.com/questions/47321012/difference-between-defer-call-a-function-directly-and-call-it-in-an-anonymous-fu) is an explanation.

kjezek: I got it, thanks for explanation.  It is confusing because it uses actually two not that know features:  a trick with correct defer function, and that the variable `res` gets assigned the return error.

I think it would help make this a bit less cryptic to name the return property `err` as well.

petr-hanzl: I did not know about this defer feature as well. Anyhow I think `res` here means `result` as in a verification result is either error or nothing for success. We can rename it tho

kjezek (changes requested):
I think we have to finish this discussion before merging: https://github.com/Fantom-foundation/Carmen/pull/881#discussion_r1619310167

go/database/mpt/verification.go
HerbertJordan: What is the problem with an extra loop?

It simplifies the code and eliminates extra data structures.

go/database/mpt/live_trie.go
HerbertJordan: This discussion is a consequence of the missing cleanup of the dependency structure inside the MPT directory, which is pending since months. So ultimately, there is no clean solution.

Right now we have the following situation:
- a [`Forest`](https://github.com/Fantom-foundation/Carmen/blob/ac980c2cb1983277f771c061909164f9ae3b3a74/go/database/mpt/forest.go#L78) instance owns all MPT nodes, but not the codes
- a [`LiveTrie`](https://github.com/Fantom-foundation/Carmen/blob/ac980c2cb1983277f771c061909164f9ae3b3a74/go/database/mpt/live_trie.go#L29) is a wrapper around a Forest, so it also just owns the MPT nodes, no codes
- a [`MptState`](https://github.com/Fantom-foundation/Carmen/blob/ac980c2cb1983277f771c061909164f9ae3b3a74/go/database/mpt/state.go#L135) instance owns a Forest + the codes
- a [`ArchiveTrie`](https://github.com/Fantom-foundation/Carmen/blob/ac980c2cb1983277f771c061909164f9ae3b3a74/go/database/mpt/archive_trie.go#L32) owns a [`LiveState`](https://github.com/Fantom-foundation/Carmen/blob/ac980c2cb1983277f771c061909164f9ae3b3a74/go/database/mpt/state.go#L103) which is `MptState` in disguised, not at all what you would expect from the `Trie` in the `ArchiveTrie` name

Ideally, the package would be refactored such that everything that has a `Trie` in its name should only be concerned with MPT nodes, no codes. But this is not the case right now.

Following the ideal case, I would have the following expectations:
- a function called `verifyLiveTrie` checks the MPT nodes of a LiveDB
- a function called `verifyArchiveTrie` checks the MPT nodes of an Archive
- a function called `verifyMptState` checks the MPT nodes of the forest it owns + the codes

Right now, we have a
- `verifyCodesAndForest`
- `verifyArchive`
which basically circumvents the problem of the pending refactoring by adding new terms, that are not bound to existing types. `Archive` is neither an `ArchiveTrie` nor a `MptState`, but the way they are used together; `CodesAndForest` could be either a `LiveTrie` + `Code` file or a `ArchiveTrie` + `Code` file, but it assumes that the code file is named as the `MptState` is doing it; 

So my problem is: instead of improving the situation, or at least align to the existing setup, this naming adds additional confusion.

petr-hanzl: I hope I understood your comment correctly. Could you check last commit? @HerbertJordan

go/database/mpt/live_trie.go
```diff
@@ -71,7 +71,7 @@ func VerifyFileLiveTrie(directory string, config MptConfig, observer Verificatio
 	if !exists {
 		return nil
 	}
-	return VerifyFileForest(directory, config, []Root{{
+	return VerifyMptState(directory, config, []Root{{
```
HerbertJordan, line 74: Since the `LiveTrie` does not own an `MptState` it should not check for the codes. It should call `VerifyFileForest` here.

go/database/mpt/verification_test.go
```diff
@@ -412,6 +417,144 @@ func TestVerification_ValueNodeHashModificationIsDetected(t *testing.T) {
 		}
 	})
 }
+
+func TestVerification_MissingCodeHashInCodeFileIsDetected(t *testing.T) {
+	runVerificationTest(t, func(t *testing.T, dir string, config MptConfig, roots []Root) {
+		err := VerifyMptState(dir, config, roots, NilVerificationObserver{})
```
HerbertJordan, line 423: I would appreciate if those tests could be self-contained, such that everything one needs to know for the test is present in the test.

In this case, the test checks for a missing code, but this scenario is not set up in the test. A reader needs to know that there is a code expected by the created state that by default is not there.

Better: let the `runVerificationTest` function create a correct code file and explicitly remove the code which's absence should be detected.

In this case, please also add a test case where the verification for the state created by `runVerificationTest` passes the verification test.

HerbertJordan (changes requested):

go/database/mpt/verification_test.go
petr-hanzl: This was the design I went with these tests but @kjezek did not like it so we changed it. @kjezek wdyt

kjezek: I am fine with Herbert's comment. No problem to change it.

I think what Herbert asks for is not exactly reverting to what you had before. The previous version had a problem that it customised nodes in some tests (it is what Herbert asks), but in other tests it relied on empty code hashes that were the default.  If the tests are consistent, i.e. customisation is happening everywhere, I am fine with it.

go/database/mpt/verification_test.go
```diff
@@ -460,11 +621,66 @@ func TestVerification_HashesOfEmbeddedNodesAreIgnored(t *testing.T) {
...
+		)
+
+		encoder, _, _, _ := getEncoder(config)
+
+		modifyNode(t, dir+"/accounts", encoder, func(node *AccountNode) {
+			node.info.Balance[2]++
+		})
+
+		if err := verifyFileForest(dir, config, roots, observer); err == nil {
+			t.Errorf("found unexpected error in fresh forest: %v", err)
```
HerbertJordan, line 648: this error message is wrong

go/database/mpt/verification_test.go
```diff
@@ -460,11 +621,66 @@ func TestVerification_HashesOfEmbeddedNodesAreIgnored(t *testing.T) {
...
+		)
+
+		encoder, _, _, _ := getEncoder(config)
+
+		modifyNode(t, dir+"/accounts", encoder, func(node *AccountNode) {
+			node.info.Balance[2]++
+		})
+
+		if err := VerifyMptState(dir, config, roots, observer); err == nil {
+			t.Errorf("found unexpected error in fresh forest: %v", err)
```
HerbertJordan, line 671: this error message is wrong

HerbertJordan (approved):
Thanks for working through this lengthy process! But I think the result looks good now :)

kjezek (dismissed):
thanks for this, I think the result is worth the effort

HerbertJordan (approved):

kjezek (approved):

PR #889, opened 2024-06-02
Gracefully close database on export interrupt.

### LINKED ISSUE #885: Support abort feature for stopping MPT exports gracefully
Right now, the export of genesis data from MPTs can not be aborted by the user. Once started, users have to wait until the export is finished. Killing the export process will leave the database in an undefined state.

A possible design choice would be to pass a [Context](https://pkg.go.dev/context#Context) to the export function to allow signaling the users intend to abort an export progress.

An complementary feature request is to run export tools in read-only mode (see #664).

petr-hanzl:
This PR allows interrupting when exporting genesis and not corrupting database. 
Fixes #885


go/database/mpt/io/archive.go
```diff
@@ -100,6 +100,9 @@ func ExportArchive(directory string, out io.Writer) error {
 
 	// Encode diff of each individual block.
 	for block := uint64(0); block <= maxBlock; block++ {
+		if interrupt(ctx) {
+			return errors.Join(archive.Close(), errors.New("failed exporting content: export interrupted"))
```
kjezek, line 104: should we consider an interrupted export as an error state?

go/database/mpt/io/archive.go
```diff
@@ -196,6 +199,15 @@ func ExportArchive(directory string, out io.Writer) error {
 	return archive.Close()
 }
 
+func interrupt(ctx context.Context) bool {
```
kjezek, line 202: name it `isInterrupted`?

go/database/mpt/io/archive_test.go
```diff
@@ -44,7 +45,7 @@ func TestIO_Archive_ExportAndImport(t *testing.T) {
 
 	// Export the archive into a buffer.
 	buffer := new(bytes.Buffer)
-	if err := ExportArchive(sourceDir, buffer); err != nil {
+	if err := ExportArchive(sourceDir, buffer, context.Background()); err != nil {
```
kjezek, line 48: I think it is common that the context goes as the first argument

go/database/mpt/io/live.go
```diff
@@ -302,9 +302,24 @@ func runImport(directory string, in io.Reader, config mpt.MptConfig) (root mpt.N
 type exportVisitor struct {
 	out io.Writer
 	err error
+	ctx context.Context
+}
+
+func (e *exportVisitor) interrupt() bool {
```
kjezek, line 308: why is not the already defined function `func interrupt(ctx context.Context) bool` re-used?

go/database/mpt/tool/export.go
```diff
@@ -64,23 +68,51 @@ func doExport(context *cli.Context) error {
 
 	start := time.Now()
 	logFromStart(start, "export started")
+
+	ctx, cancel := context.WithCancel(cliCtx.Context)
+	isCanceled := catchInterrupt(ctx, cancel, start)
```
kjezek, line 73: you can use the method you already defined to check the context was interrupted (called now `interrupt`) and then you can avoid `isCanceled` property

go/database/mpt/tool/export.go
```diff
@@ -64,23 +68,51 @@ func doExport(context *cli.Context) error {
...
+	isCanceled := new(atomic.Bool)
+	isCanceled.Store(false)
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, os.Interrupt, syscall.SIGTERM, syscall.SIGINT)
+	go func() {
+		defer signal.Stop(c)
+		select {
+		case <-c:
+			logFromStart(start, "Sending close signal...")
+			logFromStart(start, "Do not kill this command, you will end up with corrupted database!")
```
kjezek, line 107:
```suggestion
			logFromStart(start, "closing, please wait until proper shutdown to prevent database corruption")
```

go/database/mpt/tool/export.go
```diff
@@ -64,23 +68,51 @@ func doExport(context *cli.Context) error {
...
+func catchInterrupt(ctx context.Context, cancel context.CancelFunc, start time.Time) *atomic.Bool {
+	isCanceled := new(atomic.Bool)
+	isCanceled.Store(false)
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, os.Interrupt, syscall.SIGTERM, syscall.SIGINT)
+	go func() {
+		defer signal.Stop(c)
+		select {
+		case <-c:
+			logFromStart(start, "Sending close signal...")
```
kjezek, line 106:
```suggestion
		
```

go/database/mpt/tool/export_test.go
```diff
@@ -0,0 +1,32 @@
...
+package main
+
+import (
+	"context"
+	"syscall"
+	"testing"
+	"time"
+)
+
+func Test_CatchInterrupt(t *testing.T) {
```
kjezek, line 20: I think this is too trivial test - it mostly tests only golang build-in function.

You should test that really export can be interrupted. It may be tricky timing wise, but perhaps it could be done via a mocked io.Writer, which will be very slow. It will take e.g a second to accept data, then you could have enough time to  trigger cancel even when exporting a small sample of data

kjezek (changes requested):
Thanks for this effort. I think it looks overall good. Please have a look at a few minor recommendations. Furthermore. I think tests should be more thorough.

go/database/mpt/io/archive.go
petr-hanzl: Thinking about it now its probably a better not to return an error.

go/database/mpt/io/live.go
petr-hanzl: Good catch.

go/database/mpt/tool/export.go
petr-hanzl: Right. Good idea!

go/database/mpt/tool/export_test.go
```diff
@@ -0,0 +1,186 @@
...
+			createDB: createArchiveTrie,
+		},
+		"live": {
+			export:   cio.Export,
+			createDB: createLiveTrie,
+		},
+	}
+
+	for name, tf := range tests {
+		t.Run(name, func(t *testing.T) {
```
kjezek, line 45: I think this test is fragile.  What if the inner go function is never scheduled. The test will finish without kill was triggered and it would look like a passing test. 

I think it should work this way:
* `export`  executed in a go routine. 
* only when it is confirmed this routine was scheduled, proceed in the main thread
*  in the main thread wait a bit to make sure export starts to export. 
*  trigger kill signal
*  check the export was not finished

ideally you can utilise your mock writer to count number of writes. This way you can nicely synchronise both threads:
* trigger kill only when `count` >= 1 - it will make sure that export has started
* after kill is processed, check that `count <  maxCount`

kjezek: looking more at the test data, I think there is actually  a huge change that the go routine is never scheduled - the test data is small, i.e. export will run quickly while the go-routine is doing nothing 1s

kjezek (changes requested):
Thanks for updates. Can you please

1. have a look at tests - check my comment
2. I am not very expert in using the `CancellableContext`, bear with me. Question is:

Is not it actually an overkill to use this component here? It looks to me that it features a channel that can be read when cancelled. I.e. I would understand that there is an expensive process running in the main thread (which we have), but there is also an expensive process to cancel (which we do not have). I.e. the channel is read in the thread where many operations can happen, which is great because it is already in a parallel thread. 

But it looks to me that we utilise this machinery just to read true/false information.  Hence, the question is: why do not we have just a simple atomic boolean, which is set to true/false when the kill is triggered and we do not read this value in places where now the` IsInterrupted()` method was introduced (just to read true/false).

petr-hanzl:
@kjezek 
> Is not it actually an overkill to use this component here? It looks to me that it features a channel that can be read when cancelled. I.e. I would understand that there is an expensive process running in the main thread (which we have), but there is also an expensive process to cancel (which we do not have). I.e. the channel is read in the thread where many operations can happen, which is great because it is already in a parallel thread.

I dont think this is somewhat expensive. I  always see it being used to distribute an `interrupt` though the whole project - `ctx.Done()` being called a lot of times. I think the main reason for its usage is the thread safety and easy usage in go routines using when using`select{}`.
 
> But it looks to me that we utilise this machinery just to read true/false information. Hence, the question is: why do not we have just a simple atomic boolean, which is set to true/false when the kill is triggered and we do not read this value in places where now the` IsInterrupted()` method was introduced (just to read true/false).

Regarding your suggestion I think you might be right, since we dont use the feature of `ctx.Done()` by using it in a `select{}`. So we can change to actually simplify the code. I just remember @HerbertJordan talking about a `context` so thats why I used.

go/database/mpt/tool/export_test.go
petr-hanzl: Each write is a one second `Sleep` but your suggestions are very much reasonable. I will rework the test

kjezek: right, I forget that the writing is slow as well.  Still the current code does not assure that the go routine is planned at all, even if it is probably rare for this to happen.

go/database/mpt/tool/export_test.go
```diff
@@ -0,0 +1,217 @@
...
+	"syscall"
+	"testing"
+	"time"
+
+	"github.com/Fantom-foundation/Carmen/go/common"
+	"github.com/Fantom-foundation/Carmen/go/database/mpt"
+	cio "github.com/Fantom-foundation/Carmen/go/database/mpt/io"
+)
+
+func TestExport_CanBeInterrupted(t *testing.T) {
```
kjezek, line 27: I now noticed that this test should probably go to the IO package, it tests methods from that package, not the Tool package.

kjezek (changes requested):
I think the context creation and registration of the `kill` signal should happen independently in `io.ExportArchive` and `io.Export`, not in `tools.doExport`. 

It may look like code duplication, but it is just a few line of codes. 
Current design has shortcomings:
* the method that read state of the cancellable context is called across packages
* tests are called across packages  

While the new design will encapsulate each functionality
* instead of calling the public method, the channel `ctx.Done` may be read where the export is to be interrupted and directly interrupted when needed.  No need to check the shared context across packages.

The tests will simplify as well:
* they may use already existing methods to fill-in test data, no need to duplicate this code

go/database/mpt/tool/export_test.go
```diff
@@ -0,0 +1,217 @@
...
+}
+
+func (m *mockWriter) Write(_ []byte) (n int, err error) {
+	m.numOfWrites++
+	// slow down writting to be able to interrupt the export
+	time.Sleep(100 * time.Millisecond)
+	return 0, nil
+}
+
+func createLiveTrie(t *testing.T, dir string) {
```
kjezek, line 125: if the test is in the `io` package, there is already functionality to create test data

go/database/mpt/tool/export_test.go
```diff
@@ -0,0 +1,217 @@
...
+	_, err = db.GetHash()
+	if err != nil {
+		t.Fatalf("failed to fetch hash from DB: %v", err)
+	}
+	if err = db.Close(); err != nil {
+		t.Fatalf("failed to close DB: %v", err)
+	}
+}
+
+func createArchiveTrie(t *testing.T, dir string) {
```
kjezek, line 165: if the test is in the `io` package, there is already functionality to create test data

go/database/mpt/io/archive.go
```diff
@@ -100,6 +100,9 @@ func ExportArchive(directory string, out io.Writer) error {
 
 	// Encode diff of each individual block.
 	for block := uint64(0); block <= maxBlock; block++ {
+		if IsContextDone(ctx) {
+			return archive.Close()
```
HerbertJordan, line 104: I think an abort by the context should result in an error since the export was not completed successfully.
Something linke:
```
return errors.Join(
    ErrCanceledByUser,
    archive.Close(),
)
```

petr-hanzl: We had a discussion earlier with @kjezek and we agreed that it actually is not an error state. Should we still bring back the error?

HerbertJordan: I would argue that it is an error, since the requested operation -- the export of the archive -- is not fulfilled and the caller should discard whatever was written into the output writer.

kjezek: thinking that the program can be actually killed by system, not only the user, would suggest that it is better to return an error

go/database/mpt/tool/export_test.go
```diff
@@ -0,0 +1,217 @@
...
+			}()
+
+			time.Sleep(200 * time.Millisecond)
+			err := syscall.Kill(syscall.Getpid(), syscall.SIGINT)
+			if err != nil {
+				t.Error("failed to create a SIGINT signal")
+				return
+			}
+			select {
+			case <-success:
```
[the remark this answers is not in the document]
petr-hanzl: I want to make this test deadlock-safe, thats why I introduced a timeout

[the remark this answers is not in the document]
kjezek: how can this deadlock? If there is a bug in sense the kill signal is not correctly propagated, the worse case is that the export will complete, but it will not deadlock.

go/database/mpt/tool/export_test.go
```diff
@@ -0,0 +1,217 @@
...
+			dir := filepath.Join(t.TempDir(), name)
+			tf.createDB(t, dir)
+			ctx, cancel := context.WithCancel(context.Background())
+			catchInterrupt(ctx, cancel, time.Now())
+
+			success := make(chan any, 1)
+			writer := newMockWriter(t)
+
+			// first find number of writes
+			if err := tf.export(ctx, dir, writer); err != nil {
```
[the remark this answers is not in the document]
petr-hanzl: The logic of the interrupt has been moved to a different package so this argument has been removed hence this comment is not relevant anymore.

go/database/mpt/tool/export_test.go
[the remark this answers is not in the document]
petr-hanzl: You are right. Thanks!

go/database/mpt/tool/export_test.go
```diff
@@ -0,0 +1,217 @@
...
+			err = tf.check(t, dir)
+			if err != nil {
+				t.Fatalf("db is corrupted after export interrupt: %v", err)
+			}
+		})
+	}
+
+}
+
+func checkExportedLiveDB(t *testing.T, dir string) error {
```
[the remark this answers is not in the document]
petr-hanzl: The logic of the interrupt has been moved to a different package and this func has been removed hence this comment is not relevant anymore.

go/database/mpt/tool/export_test.go
```diff
@@ -0,0 +1,217 @@
...
+	t.Helper()
+	return checkLiveDB(dir, cio.MptInfo{Config: mpt.S5LiveConfig})
+}
+func checkExportedArchive(t *testing.T, dir string) error {
```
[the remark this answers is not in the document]
petr-hanzl: The logic of the interrupt has been moved to a different package and this func has been removed hence this comment is not relevant anymore.

go/database/mpt/tool/export_test.go
petr-hanzl: Yea I agree. I thought it would create unnecessary duplicity, but the duplicity is very small given the fact the code was a bit complicated with the interrupt check inside `tool` package. Moved to `io` package.

go/database/mpt/io/archive.go
petr-hanzl: Error added.

go/database/mpt/tool/export_test.go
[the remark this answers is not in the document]
petr-hanzl: I was thinking the same not to add an unnecessary slow down.

go/database/mpt/io/interrupt_test.go
```diff
@@ -0,0 +1,162 @@
...
+				return
+			}
+
+			// save max count and reset number of writes
+			maxCount := writer.numOfWrites
+			writer.signalWrite = true
+			writer.waitTime = 100 * time.Millisecond
+			hasEnded := new(sync.WaitGroup)
+			go func() {
+				hasEnded.Add(1)
```
kjezek, line 70: increase this wait group befor this go-routine.  I can see from the code that this counter will be always called, but it is just a typical usage to increase before the gouroutine is started, i.e. easier to read.

go/database/mpt/io/interrupt_test.go
```diff
@@ -0,0 +1,162 @@
...
+				defer hasEnded.Done()
+				if err := ExportArchive(sourceDir, writer); err != nil {
```
kjezek, line 72:
```suggestion
				if err := tf.export(sourceDir, writer); err != nil {
```

go/database/mpt/io/interrupt_test.go
```diff
@@ -0,0 +1,162 @@
...
+					got := err.Error()
+					want := errCanceled.Error()
+					if !strings.Contains(got, want) {
+						t.Errorf("unexpected error: got: %v, want: %v", got, want)
+					}
+					return
+				}
+				t.Error("export was interrupt hence should raise an error")
```
kjezek, line 80: I think you do not have to have an extra brach when the err is not nil. Just check it all the same and it simplifies the assertion.

go/database/mpt/io/interrupt_test.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+}
+
+func createTestArchive(t *testing.T, sourceDir string) {
+	t.Helper()
+	source, err := mpt.OpenArchiveTrie(sourceDir, mpt.S5ArchiveConfig, 1024)
+	if err != nil {
+		t.Fatalf("failed to create archive: %v", err)
+	}
+	_ = fillTestBlocksIntoArchive(t, source)
```
kjezek, line 117:
```suggestion
	fillTestBlocksIntoArchive(t, source)
```

go/database/mpt/io/interrupt_test.go
```diff
@@ -0,0 +1,162 @@
...
+}
+
+type mockWriter struct {
+	numOfWrites int
+	signalWrite bool
+	waitTime    time.Duration
+	wg          *sync.WaitGroup
+}
+
+func (m *mockWriter) Write(_ []byte) (n int, err error) {
```
kjezek, line 152:
```suggestion
func (m *mockWriter) Write([]byte) (n int, err error) {
```

go/database/mpt/io/interrupt_test.go
```diff
@@ -0,0 +1,162 @@
...
+
+			// first find number of writes
+			if err := tf.export(sourceDir, &mockWriter{wg: writerWg}); err != nil {
+				t.Errorf("unexpected error: %v", err)
+				return
+			}
+
+			// save max count and reset number of writes
+			maxCount := writer.numOfWrites
+			writer.signalWrite = true
```
kjezek, line 66: `numOfWrites` should be zeroed as well.  Ideally, create a new struct here, it is more error prone to re-use the previous structure

go/database/mpt/io/interrupt_test.go
```diff
@@ -0,0 +1,162 @@
...
+			writerWg.Wait()
+			err := syscall.Kill(syscall.Getpid(), syscall.SIGINT)
+			if err != nil {
+				t.Error("failed to create a SIGINT signal")
+				return
+			}
+
+			// wait for the export to finish
+			hasEnded.Wait()
+			if maxCount == writer.numOfWrites {
```
kjezek, line 93: `numOfWrites` must be thread safe - it is read here and modified in the goroutine.

go/database/mpt/io/interrupt_test.go
```diff
@@ -0,0 +1,162 @@
...
+	if err != nil {
+		t.Fatalf("failed to create archive: %v", err)
+	}
+	_ = fillTestBlocksIntoArchive(t, source)
+	if err = source.Close(); err != nil {
+		t.Fatalf("failed to close test DB: %v", err)
+	}
+}
+
+func checkLive(t *testing.T, sourceDir string) {
```
kjezek, line 123: since this method is visible across this package, it should not steal such a unique name, and it should be more explanatory what it is actually doing.  What about `checkCanOpenLiveDB`

go/database/mpt/io/interrupt_test.go
```diff
@@ -0,0 +1,162 @@
...
+	if err != nil {
+		t.Fatalf("failed to open db: %v", err)
+	}
+	err = db.Close()
+	if err != nil {
+		t.Fatalf("failed to close db: %v", err)
+	}
+}
+
+func checkArchive(t *testing.T, sourceDir string) {
```
kjezek, line 134: since this method is visible across this package, it should not steal such a unique name, and it should be more explanatory what it is actually doing.

kjezek (changes requested):
thanks for all recent changes. Please have a look at my next review comments.

go/database/mpt/io/interrupt_test.go
petr-hanzl: I think I dont understand

go/database/mpt/io/archive.go
```diff
@@ -70,6 +69,8 @@ func ExportArchive(directory string, out io.Writer) error {
 		return err
 	}
 
+	ctx := catchInterrupt()
```
HerbertJordan, line 72: The function `ExportArchive` is not the top level function and thus, I think, should not handle interrupts. 

Since support for the interrupt is part of the user interface, this should be done by the `main` package, like the [export](https://github.com/Fantom-foundation/Carmen/blob/main/go/database/mpt/tool/export.go) tool or the [sonictool](https://github.com/Fantom-foundation/Sonic/blob/main/cmd/sonictool/main.go). This also enables the tool to coordinate multiple shutdown activities if needed. 

To fix this, a context should be already passed to this function. Wdyt?

go/database/mpt/io/interrupt.go
```diff
@@ -0,0 +1,52 @@
...
+import (
+	"context"
+	"errors"
+	"log"
+	"os"
+	"os/signal"
+	"syscall"
+)
+
+var errCanceled = errors.New("export was interrupted")
```
HerbertJordan, line 22: To allow API users to check for this specific error, it should be public.

HerbertJordan, line 22: Consider using a `ConstError` to make this error immutable. An example is [here](https://github.com/Fantom-foundation/Carmen/blob/cb0032b7724512a252766f973bddcacdc9752c54/go/carmen/database.go#L22).

go/database/mpt/io/live.go
```diff
@@ -72,6 +72,8 @@ func Export(directory string, out io.Writer) error {
 	}
 	defer db.Close()
 
+	ctx := catchInterrupt()
```
HerbertJordan, line 75: Same argument as for the archive export. I think this is a too narrow scope to deal with interrupt handler.

go/database/mpt/io/interrupt_test.go
```diff
@@ -0,0 +1,168 @@
...
+
+func (m *mockWriter) Write([]byte) (n int, err error) {
+	m.numOfWrites.Add(1)
+	// inform the test that first write has happened
+	if m.signalWrite {
+		m.signalWrite = false
+		m.wg.Done()
+	}
+	// slow down writing to be able to interrupt the export
+	time.Sleep(m.waitTime)
```
HerbertJordan, line 166: The use of timing may make the test flaky and inefficient. 

How about triggering the interrupt signal from within the writer after a given number of writes? This would make the test resilient against relative timing. Also, you can simplify the test since you no longer would need multiple go-routines for testing.

HerbertJordan (changes requested):

go/database/mpt/io/archive.go
petr-hanzl: First I implemented it into `database/mpt/tool/export.go` but we agreed with @kjezek that we will move it to `io` package mainly because testing this feature inside `main` was not easy.
I think your arguments are very much valid, although I think we should keep it in `io` as it is

kjezek: I think there is a way to fulfil both requirements.  You can propagate the context from the top level function, where it also registers for the kill signal.  You can still check for the context being done in each export method in isolation.  The tests can stay where they are now in their correct packages. 

In short, you propagate the context from the top level, everything else stays as it is.

go/database/mpt/io/interrupt_test.go
kjezek: it is not (fully) resolved. You still re-use the same struct. I think it is error prone, it is better to create a new struct.

go/database/mpt/io/interrupt_test.go
kjezek: you can test:

```
err := export(...)
if err != expected err {
 // assertion
}
```

you do not have to do
```
if err != nil {
  if err != expected err {
 // assertion
} 
// assertion 2
}

```

go/database/mpt/io/interrupt_test.go
petr-hanzl: I implemented it but it still requires some `sleep` time. The export is too fast... even tho `cancel()` is called the func call ends before the interrupt is noticed. Another option would be to create a bigger database to extend the export.

go/database/mpt/io/interrupt_test.go
```diff
@@ -0,0 +1,149 @@
...
+			err := tf.export(sourceDir, writer)
+			if err == nil {
+				t.Fatal("export was interrupted, error must not be nil")
+			}
+
+			got := err.Error()
+			want := ErrCanceled.Error()
+			if !strings.Contains(got, want) {
+				t.Errorf("unexpected error: got: %v, want: %v", got, want)
+			}
```
kjezek, line 75:
```suggestion
			if got, want := err, ErrCanceled; !errors.Is(got, want) {
				t.Errorf("unexpected error: got: %v, want: %v", got, want)
			}
```

go/database/mpt/io/interrupt_test.go
```diff
@@ -0,0 +1,149 @@
...
+			// first find number of writes
+			if err := tf.export(sourceDir, writer); err != nil {
+				t.Fatalf("unexpected error: %v", err)
+			}
+
+			// save max count and reset number of writes
+			maxCount := writer.numOfWrites
+			// reset number of writes, so we can compare
+			// that export was indeed interrupted
+			writer.numOfWrites = 0
```
kjezek, line 63: please init a new struct here instead of emptying the old one.

go/database/mpt/io/archive.go
petr-hanzl: Moving `catchInterrupt` would require to move the test to that package aswell to avoid circular dependency. That requires recreating the helper functions for database creation. Do we really want that...?

kjezek: what if the whole `interrupt.go` file is made more generic feature? It is now in the `mpt` directory, but it looks like more generic feature. What if we store it in `common/interrupt/interrupt.go` and make its methods public.  It would solve the circular dependency and I think we may need this functionality in other parts of carmen, all imports, exports, verification, possibly other schemes

go/common/interrupt/interrupt.go
```diff
@@ -0,0 +1,53 @@
...
+	"context"
+	"log"
+	"os"
+	"os/signal"
+	"syscall"
+
+	"github.com/Fantom-foundation/Carmen/go/common"
+)
+
+const ErrCanceled = common.ConstError("export was interrupted")
```
kjezek, line 23: it does not have to be only export

go/common/interrupt/interrupt.go
```diff
@@ -0,0 +1,53 @@
...
+	case <-ctx.Done():
+		return true
+	default:
+		return false
+	}
+}
+
+// Catch catches SIGTERM and SIGINT signals and
+// prevents export from corrupting database by canceling its context.
+func Catch(parent context.Context) context.Context {
```
kjezek, line 38: rename:  `Catch` -> `Register` ?

go/common/interrupt/interrupt_test.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+func Test_CatchCancelsContextWhenInterrupted(t *testing.T) {
+	ctx := Catch(context.Background())
+	err := syscall.Kill(syscall.Getpid(), syscall.SIGINT)
+	if err != nil {
+		t.Fatal("failed to create a SIGINT signal")
+	}
+	select {
+	case <-time.After(1 * time.Second):
```
kjezek, line 17: do not timeout here, it complicates the test. We expect that what is tested works and when it breaks the test will hang, but it will signal that something went wrong.

go/common/interrupt/interrupt.go
```diff
@@ -0,0 +1,53 @@
...
+	"syscall"
+
+	"github.com/Fantom-foundation/Carmen/go/common"
+)
+
+const ErrCanceled = common.ConstError("export was interrupted")
+
+// IsContextDone returns true if the given context CancelFunc has been called.
+// Otherwise, returns false.
+func IsContextDone(ctx context.Context) bool {
```
kjezek, line 27: simplify `IsContextDone` -> `IsDone` or `IsCancelled` ?

go/database/mpt/io/interrupt_test.go
```diff
@@ -0,0 +1,144 @@
...
+			ctx := interrupt.Catch(context.Background())
+
+			writer := &mockWriter{}
+			writer.signalInterrupt = true
+			err := tf.export(ctx, sourceDir, writer)
+			if got, want := err, interrupt.ErrCanceled; !errors.Is(got, want) {
+				t.Errorf("unexpected error: got: %v, want: %v", got, want)
+			}
+
+			if maxCount == writer.numOfWrites {
```
kjezek, line 72:
```suggestion
			if maxCount == writer.numOfWrites || writer.numOfWrites == 0 {
```

go/database/mpt/io/interrupt_test.go
```diff
@@ -0,0 +1,144 @@
...
+	if err != nil {
+		t.Fatalf("failed to open archive: %v", err)
+	}
+	err = archive.Close()
+	if err != nil {
+		t.Fatalf("failed to close archive: %v", err)
+	}
+}
+
+type mockWriter struct {
```
kjezek, line 126: `mockWriter` -> `interruptSendingWriter`

kjezek (approved):
thanks for all these changes. It looks good to me. I added just a few last comments reg. improving naming here and there.

go/common/interrupt/interrupt.go
```diff
@@ -0,0 +1,53 @@
...
+// prevents export from corrupting database by canceling its context.
+func Catch(parent context.Context) context.Context {
+	ctx, cancel := context.WithCancel(parent)
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, os.Interrupt, syscall.SIGTERM, syscall.SIGINT)
+	go func() {
+		defer signal.Stop(c)
+		select {
+		case <-c:
+			log.Println("closing, please wait until proper shutdown to prevent database corruption")
```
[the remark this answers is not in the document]
petr-hanzl: I think we should notify the user that the kill signal has been registered by the app. How about `closing, please wait until proper shutdown` to make it more generic?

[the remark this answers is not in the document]
kjezek: My point was that now when this functionality is generally available here, it can be registered and triggered from everywhere. Hence, there may be situations where telling the user to wait does not make sense. Think of scenarios where carmen is embedded in a third party client and interrupt comes from another component. There is no way for user to decide if he waits or not. 

The proper solution would be to propagate this signal to registered observers and they decide what to do, but since we do not need it at the moment, I would not complicate this PR too much. 

I would suggest at least to make the log as generic as possible, perhaps just printing "shutting down".

Since we do not have logging anywhere in Carmen, I asked at the beginning just for removing this log here.  But if you feel it is important to have it, I guess there is no harm to log.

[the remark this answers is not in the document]
petr-hanzl: I understand, but if the client needs to use this feature it prevents some ungraceful closing hence telling user to wait until its properly closed should be there. `Shutting down` is the most generic one - lets use it.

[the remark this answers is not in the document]
petr-hanzl: Or atleast some kind of log. I can imagine a scenario where used wanted to interrupt the process (even the export) and nothing happens. This would cause confusion. Now that we at least log `shutting down` the user know that something is happening.

go/common/interrupt/interrupt.go
```diff
@@ -0,0 +1,53 @@
...
+	case <-ctx.Done():
+		return true
+	default:
+		return false
+	}
+}
+
+// Register catches SIGTERM and SIGINT signals and
+// prevents export from corrupting database by canceling its context.
+func Register(parent context.Context) context.Context {
```
HerbertJordan, line 38: Maybe rename this to `CancelOnInterrupt`? `Register` is a bit generic.

go/common/interrupt/interrupt.go
```diff
@@ -0,0 +1,53 @@
...
+	select {
+	case <-ctx.Done():
+		return true
+	default:
+		return false
+	}
+}
+
+// Register catches SIGTERM and SIGINT signals and
+// prevents export from corrupting database by canceling its context.
```
HerbertJordan, line 37: nit: the documentation is outdated; it is no longer connected with the database

HerbertJordan (dismissed):
Looks great, thanks!

PR #890, opened 2024-06-02
Prevent license script paths expansion

Mike-CZ:
This change prevents bash to expand wildcard paths. If the script was run from a directory that contains any of wildcard paths defined in `ignore_files` (`ignore_files=('cpp/third_party/*')`), it led to expansion of those paths and calling the `find` method with wrong argument. In this particular case it expanded as follows:
```
! -path cpp/third_party/crc32c cpp/third_party/ethash cpp/third_party/gperftools cpp/third_party/leveldb cpp/third_party/snappy
```
The correct way of passing `! path` argument to `find` method is
```
! -path <path_1> ! -path <path_2> ...
```

The arguments are passed into `find` function via arguments list, which prematurely prevents the expansion of wildcard paths.

### DIFF
--- a/scripts/license/add_license_header.sh
+++ b/scripts/license/add_license_header.sh
@@ -70,11 +70,18 @@ add_license_to_files() {
     local license_header="$(extend_license_header "$prefix")"
     local result=0
 
-    # expand list if ignored files to a single string
-    local ignore=$(printf " ! -path *%s" "${ignore_files[@]}")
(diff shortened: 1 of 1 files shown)


Mike-CZ:
@kjezek may I ask you to try running it on your mac device? thx

HerbertJordan (approved):

kjezek (approved):
thanks, it works on macOS

PR #892, opened 2024-06-04
introduces tribool type

kjezek:
this PR introduces a new type `Tribool`

It is a kind of boolean type, which can be apart from `true/false` also unknown.

### DIFF
--- a/go/common/tribool/tribool.go
+++ b/go/common/tribool/tribool.go
@@ -1,0 +1,61 @@
+// Copyright (c) 2024 Fantom Foundation
+//
+// Use of this software is governed by the Business Source License included
+// in the LICENSE file and at fantom.foundation/bsl11.
+//
+// Change Date: 2028-4-16
+//
+// On the date above, in accordance with the Business Source License, use of
+// this software will be governed by the GNU Lesser General Public License v3.
+
+package tribool
+
+// Tribool is a three-valued logic type.
+// It can be either True, False or Unknown.
+// It is useful for representing the result of a comparison
+// when the comparison is not possible or could not be determined.
+type Tribool struct {
+	value byte
+}
+
+// Unknown returns true if the value is unknown.
+func (t Tribool) Unknown() bool {
+	return t.value == 0 || t.value > 2
+}
+
+// True returns true if the value is true.
+func (t Tribool) True() bool {
+	return t.value == 1
+}
+
+// False returns true if the value is false.
+func (t Tribool) False() bool {
+	return t.value == 2
+}
+
+// String returns a string representation of the Tribool.
+func (t Tribool) String() string {
+	switch t.value {
+	case 1:
+		return "true"
+	case 2:
+		return "false"
+	default:
+		return "unknown"
+	}
+}
+
+// Unknown creates a new Tribool with unknown value.
+func Unknown() Tribool {
+	return Tribool{}
+}
+
+// False creates a new Tribool with false value.
+func False() Tribool {
+	return Tribool{value: 2}
+}
+
+// True creates a new Tribool with true value.
+func True() Tribool {
+	return Tribool{value: 1}
+}
--- a/go/common/tribool/tribool_test.go
+++ b/go/common/tribool/tribool_test.go
@@ -1,0 +1,162 @@
+// Copyright (c) 2024 Fantom Foundation
+//
+// Use of this software is governed by the Business Source License included
+// in the LICENSE file and at fantom.foundation/bsl11.
+//
+// Change Date: 2028-4-16
+//
+// On the date above, in accordance with the Business Source License, use of
+// this software will be governed by the GNU Lesser General Public License v3.
+
+package tribool
+
+import (
+	"fmt"
+	"testing"
+)
+
+func TestTribool_TrueFalseUnknown_Categories(t *testing.T) {
+	tests := []struct {
+		name string
+		tb   Tribool
+		want struct {
+			unknown  bool
+			trueVal  bool
+			falseVal bool
+		}
+	}{
+		{
+			name: "Unknown",
+			tb:   Unknown(),
+			want: struct {
+				unknown  bool
+				trueVal  bool
+				falseVal bool
+			}{true, false, false},
+		},
+		{
+			name: "False",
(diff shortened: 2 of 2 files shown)


go/common/tribool/tribool_test.go
```diff
@@ -0,0 +1,67 @@
...
+			}
+			if got, want := tt.tb.True(), tt.want.trueVal; got != want {
+				t.Errorf("got: True() = %v, want %v", got, want)
+			}
+			if got, want := tt.tb.False(), tt.want.falseVal; got != want {
+				t.Errorf("got: False() = %v, want %v", got, want)
+			}
+		})
+	}
+}
```
HerbertJordan, line 67: I think it would also be important to have a test that the `==` and `!=` operator work as expected. This should also allow them to be keys in maps, be automatically deep-copied in slices, ...

go/common/tribool/tribool_test.go
```diff
@@ -0,0 +1,67 @@
...
+//
+// On the date above, in accordance with the Business Source License, use of
+// this software will be governed by the GNU Lesser General Public License v3.
+
+package tribool
+
+import "testing"
+
+func TestTribool(t *testing.T) {
+	tests := []struct {
```
HerbertJordan, line 16: nit: make it a `map[string]struct{...}` and put the name of the test case as a key in the map

go/common/tribool/tribool.go
```diff
@@ -0,0 +1,49 @@
...
+// On the date above, in accordance with the Business Source License, use of
+// this software will be governed by the GNU Lesser General Public License v3.
+
+package tribool
+
+// Tribool is a three-valued logic type.
+// It can be either True, False or Unknown.
+// It is useful for representing the result of a comparison
+// when the comparison is not possible or could not be determined.
+type Tribool struct {
```
HerbertJordan, line 17: The type should implement `Stringer` to support debug printing.

HerbertJordan:
Looks good, just some minor requests to round it off. Thanks!

go/common/tribool/tribool_test.go
HerbertJordan, line 67: Another test that should be added: the default value `Tribool{}` is a defined value. Either `Unknown` or `False`. The important thing is that it is not neither of the three options.

go/common/tribool/tribool.go
```diff
@@ -0,0 +1,49 @@
...
+	return t.value == 1
+}
+
+// True returns true if the value is true.
+func (t Tribool) True() bool {
+	return t.value == 2
+}
+
+// NewUnknown creates a new Tribool with unknown value.
+func NewUnknown() Tribool {
```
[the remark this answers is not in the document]
kjezek: ok, make sense. 

Notice that I did not use "Is" prefix for getters. It seems common in goalng to call a setter  `True()` not `IsTrue() `etc. 

I used `New` for constructors as it seems to be also common, but in this case I thin you are right to avoid this prefix

go/common/tribool/tribool_test.go
kjezek: somehow this style is better understand by my IDE and I can run individual tests (perhaps it is more standard?). Does not matter in this case, but sometimes it is convenient. 

image

go/common/tribool/tribool_test.go
kjezek: right, will add this test

go/common/tribool/tribool.go
kjezek: right, will add

go/common/tribool/tribool_test.go
HerbertJordan: That's pretty cool ... k, in that case, go ahead.

go/common/tribool/tribool_test.go
HerbertJordan: Actually I miss-typed here: The important thing is that it is **one** of the three options.

go/common/tribool/tribool_test.go
```diff
@@ -0,0 +1,162 @@
...
+	for i, a := range items {
+		for j, b := range items {
+			if got, want := a == b, i == j; got != want {
+				t.Errorf("unexpected tribool [%v, %v] comparision: got: %v, want %v", a, b, got, want)
+			}
+		}
+	}
+}
+
+func TestTribool_Default_Is_False(t *testing.T) {
```
HerbertJordan, line 113: Should we define `Unknown` as the default? It "feels" like the safe, conservative fallback for most use cases.

"Is it valid?" => Unknown by default ...

HerbertJordan (dismissed):

go/common/tribool/tribool.go
```diff
@@ -0,0 +1,61 @@
...
+// True returns true if the value is true.
+func (t Tribool) True() bool {
+	return t.value == 1
+}
+
+// String returns a string representation of the Tribool.
+func (t Tribool) String() string {
+	switch t.value {
+	case 0:
+		return "False"
```
HerbertJordan, line 40: Should we align with the `bool` type and write `false`, `true`, and `unknown` with a lower-case first letter?

go/common/tribool/tribool.go
```diff
@@ -0,0 +1,61 @@
...
+	case 1:
+		return "True"
+	default:
+		return "Unknown"
+	}
+}
+
+// Unknown creates a new Tribool with unknown value.
+func Unknown() Tribool {
+	return Tribool{0xFF} // also the default
```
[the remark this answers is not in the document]
kjezek: right, removed

go/common/tribool/tribool_test.go
kjezek: ok, defined as Unknown

go/common/tribool/tribool_test.go
kjezek: ok, default is Unknown

go/common/tribool/tribool.go
kjezek: yes, I think it is a good idea

HerbertJordan (approved):

PR #893, opened 2024-06-04
Avoid including unused codes in LiveDB export

### LINKED ISSUE #891: Only include referenced codes in LiveDB genesis export
Right now, when exporting LiveDB genesis data all codes that are stored in a DB are included in the genesis file ([here](https://github.com/Fantom-foundation/Carmen/blob/ac980c2cb1983277f771c061909164f9ae3b3a74/go/database/mpt/io/live.go#L97-L104)). This, however, may include codes that are no longer referenced.

As a consequence, the genesis data that is exported is no longer unique for a given LiveDB state. Depending on the history of the DB, codes may be present that are no longer be referred to. Originally, the export was not required to exhibit this property, but over time this has turned into a desirable property.

Thus, the export code should be updated to only include codes into the exported genesis file that are actually referenced by a LiveDB state.

HerbertJordan:
This PR adds a filter to only include referenced codes in LiveDB genesis data exports. This is a source of non-determinism for the exported data that can result in different genesis file content for equal states.

The PR also extends the test coverage by adding test cases with multiple different codes and code reuse in the state. The export data of those states is checked for being deterministic over several executions.

This fixes #891.

### DIFF
--- a/go/database/mpt/io/live_test.go
+++ b/go/database/mpt/io/live_test.go
@@ -111,6 +111,10 @@ func TestIO_ExportedDataDoesNotContainExtraCodes(t *testing.T) {
 	// Modify the state by adding and removing code from an account.
 	// This temporary code should not be included in the resulting exported data.
 	modified, modifiedHash := exportExampleStateWithModification(t, func(s *mpt.MptState) {
+		codesBefore, err := s.GetCodes()
+		if err != nil {
+			t.Fatalf("failed to fetch codes: %v", err)
+		}
 		addr1 := common.Address{1}
 		code, err := s.GetCode(addr1)
 		if err != nil {
@@ -119,11 +123,18 @@ func TestIO_ExportedDataDoesNotContainExtraCodes(t *testing.T) {
 		modified := append(code, []byte("extra_code")...)
 		s.SetCode(addr1, modified)
 		s.SetCode(addr1, code)
+		codesAfter, err := s.GetCodes()
+		if err != nil {
+			t.Fatalf("failed to fetch codes: %v", err)
+		}
+		if before, after := len(codesBefore), len(codesAfter); before+1 != after {
+			t.Fatalf("modification did not had expected code-altering effect: %d -> %d", before, after)
+		}
 	})
 
 	// Check that the test indeed did not modify the state content.
 	if referenceHash != modifiedHash {
-		t.Fatalf("modified state has different hash than reference state: got: %x, want %x", modifiedHash, expectedHash)
+		t.Fatalf("modified state has different hash than reference state: got: %x, want %x", modifiedHash, referenceHash)
 	}
 
 	// The extra code that was only temporary in the state should no be included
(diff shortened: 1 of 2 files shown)


go/database/mpt/io/live_test.go
```diff
@@ -95,7 +95,50 @@ func TestIO_ExportAndImportAsArchive(t *testing.T) {
...
+			t.Fatalf("failed to fetch code: %v", err)
+		}
+		modified := append(code, []byte("extra_code")...)
+		s.SetCode(addr1, modified)
+		s.SetCode(addr1, code)
+	})
+
+	// Check that the test indeed did not modify the state content.
+	if referenceHash != modifiedHash {
+		t.Fatalf("modified state is not equivalent to reference state")
```
petr-hanzl, line 126:
```suggestion
		t.Fatalf("modified state has different hash to reference state: got: %x, want %x", modifiedHash, expectedHash)
```

go/database/mpt/io/live_test.go
```diff
@@ -95,7 +95,50 @@ func TestIO_ExportAndImportAsArchive(t *testing.T) {
...
+		t.Fatalf("exported data contains extra codes")
+	}
+}
+
 func exportExampleState(t *testing.T) ([]byte, common.Hash) {
+	t.Helper()
+	return exportExampleStateWithModification(t, nil)
+}
+
+func exportExampleStateWithModification(t *testing.T, modify func(s *mpt.MptState)) ([]byte, common.Hash) {
```
petr-hanzl, line 141: I think you can just call `exportExampleStateWithModification(t, nil)` and avoid creating this trivial func. Wdyt?

kjezek (dismissed):
It looks good to me. Perhaps adding an assertion that call to SetCode changes internal state would help with readability. See my comment.

go/database/mpt/io/live_test.go
HerbertJordan: The preexisting function `exportExampleState` has multiple prior uses in this file. Adding the additional `nil` would harm the readability of those cases, since it is not intuitively clear what the `nil` is for. Thus I favor this variant.

go/database/mpt/io/live_test.go
```diff
@@ -95,7 +95,50 @@ func TestIO_ExportAndImportAsArchive(t *testing.T) {
...
+	// This temporary code should not be included in the resulting exported data.
+	modified, modifiedHash := exportExampleStateWithModification(t, func(s *mpt.MptState) {
+		addr1 := common.Address{1}
+		code, err := s.GetCode(addr1)
+		if err != nil {
+			t.Fatalf("failed to fetch code: %v", err)
+		}
+		modified := append(code, []byte("extra_code")...)
+		s.SetCode(addr1, modified)
+		s.SetCode(addr1, code)
```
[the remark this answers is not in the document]
HerbertJordan: Done

kjezek (approved):
thanks, looks good to me

PR #895, opened 2024-06-06
sets storage hash in node testing framework

kjezek:
this PR extends the node testing framework with the storage hash computation.


HerbertJordan (approved):
Thanks for fixing this!

PR #896, opened 2024-06-06
Kjezek/fix nested string in rlp value node

kjezek:
This PR fixes a few bugs in RLP decoder:

1. path stored in Key or Account must be represented in compact hex form, not nibbles
2. value node is represented using a nested rlp.String
3. a zero value node is not marked as embedded (may be produced in tests)
4. length of nibble path is two times size of Key or Acccount bytes.

### DIFF
--- a/go/database/mpt/decoder_test.go
+++ b/go/database/mpt/decoder_test.go
@@ -14,6 +14,7 @@ import (
 	"fmt"
 	"github.com/Fantom-foundation/Carmen/go/common"
 	"github.com/Fantom-foundation/Carmen/go/database/mpt/rlp"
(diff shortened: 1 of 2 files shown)


go/database/mpt/decoder_test.go
```diff
@@ -320,6 +322,70 @@ func Test_compactPathToNibbles(t *testing.T) {
 	}
 }
 
+func Test_Decoder_Decode_Node_Instances(t *testing.T) {
```
HerbertJordan, line 325: Nice end-to-end test addition!

HerbertJordan (approved):

HerbertJordan (approved):

PR #897, opened 2024-06-09
Has empty storage

### LINKED ISSUE #887: Expose account storage root in API
StateDB (and so even Carmen facade) should provide `GetStorageRoot(address)` method to obtain the storage root of an account.

There are two use cases:
* checking if the storage of the account is empty during a contract deployment (only compared with `EmptyRootHash`)
  * Necessary for the go-ethereum upgrade, required by EIP-7610: https://github.com/ethereum/go-ethereum/pull/28912
* part of the witness proof - returned as part of `eth_getProof` output
  * (this may be already part of the witness proof integration?)

petr-hanzl:
This PR adds `HasEmptyStorage(addr)` which returns true if given address has empty storage.
Fixes #887

### DIFF
--- a/go/database/mpt/nodes.go
+++ b/go/database/mpt/nodes.go
@@ -16,10 +16,11 @@ import (
 	"encoding/binary"
 	"errors"
 	"fmt"
-	"github.com/Fantom-foundation/Carmen/go/common"
-	"github.com/Fantom-foundation/Carmen/go/database/mpt/shared"
 	"io"
 	"slices"
+
+	"github.com/Fantom-foundation/Carmen/go/common"
+	"github.com/Fantom-foundation/Carmen/go/database/mpt/shared"
 )
 
 // This file defines the interface and implementation of all node types in a
@@ -165,6 +166,9 @@ type Node interface {
 	// ClearStorage deletes the entire storage associated to an account. For
 	// parameter information and return values see SetValue().
 	ClearStorage(manager NodeManager, thisRef *NodeReference, this shared.WriteHandle[Node], address common.Address, path []Nibble) (newRoot NodeReference, changed bool, err error)
+
+	// HasEmptyStorage returns whether source node has empty storage.
+	HasEmptyStorage(source NodeSource, path []Nibble) (bool, error)
 
 	// Release releases this node and all non-frozen nodes in the sub-tree
 	// rooted by this node. Only non-frozen nodes can be released.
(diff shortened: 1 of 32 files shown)


kjezek (changes requested):
this PR is quite big. Could you split it into a PR for the LiveDB and ArchiveDB?

For the issue this PR resolves, it needs only the LiveDB, i.e. the ArchiveDB has a minor priority at the moment, if ever needed at all.

HerbertJordan:
> this PR is quite big. Could you split it into a PR for the LiveDB and ArchiveDB?

I agree, it needs to be cut down a bit. Maybe even an initial PR just with the changes targeting the `Forest`, followed by the `LiveDB` and then all the way up to the `Carmen` API.

go/database/mpt/nodes.go
```diff
@@ -1505,6 +1531,9 @@ type AccountNode struct {
 	pathLength byte
 }
 
+func (n *AccountNode) HasEmptyStorage(NodeSource, []Nibble) (bool, error) {
+	return n.storage.Id().IsEmpty(), nil
```
HerbertJordan, line 1535: At this point you may have reached the wrong `AccountNode`. Without checking the address, you would not know.

For instance, if you create a MPT with a single account with address `A` and then add a single storage slot, and then ask whether account `B` (A!=B) has an empty storage, the answer probably is no, although `B` should have an empty storage.

petr-hanzl:
> this PR is quite big. Could you split it into a PR for the LiveDB and ArchiveDB?
> 
> For the issue this PR resolves, it needs only the LiveDB, i.e. the ArchiveDB has a minor priority at the moment, if ever needed at all.

I dont think we can split it between Live and Archive. I need to extend [this](https://github.com/Fantom-foundation/Carmen/blob/54a30889e0311491d62d88d08a965e91f91be69d/go/state/state.go#L25) interface which is implemented by ArchiveState. I will create PR with just the `Mpt`/`Forest`/`LiveTrie` implementation but the rest will need to be done altogether.

kjezek:
> > this PR is quite big. Could you split it into a PR for the LiveDB and ArchiveDB?
> > For the issue this PR resolves, it needs only the LiveDB, i.e. the ArchiveDB has a minor priority at the moment, if ever needed at all.
> 
> I dont think we can split it between Live and Archive. I need to extend [this](https://github.com/Fantom-foundation/Carmen/blob/54a30889e0311491d62d88d08a965e91f91be69d/go/state/state.go#L25) interface which is implemented by ArchiveState. I will create PR with just the `Mpt`/`Forest`/`LiveTrie` implementation but the rest will need to be done altogether.

Right, thanks for explanation. I think we can keep the method in Archive unimplemented until it is finished.  Or as Herbert recommended, build bottom up all components and then only connect them from the top. But the liveDb has priority to fix #887

petr-hanzl:
This PR will be split into multiple smaller ones.

PR #898, opened 2024-06-09
Use `amount.Amount` in `VmStateDB` interface

Mike-CZ:
This PR does following changes:

- Moves `Amount` implementation into `amount` package (to overcome circular import dependency).
- Adds methods for adding and subtracting amounts as package methods, to be called as `amount.Add(x,y)`. For more details see [this](https://github.com/Fantom-foundation/Carmen/pull/870#discussion_r1611883224) discussion.
- Creates `Amount` alias in `carmen` package and exposes forwarded factory methods.
- Replaces `big.Int` with `amount.Amount` type within `VmStateDB` interface and all related dependencies.

This PR should be merged once the Sonic client is upgraded to new Facade API.


go/carmen/carmen_mock.go
```diff
@@ -694,6 +694,20 @@ func (mr *MockTransactionContextMockRecorder) GetState(arg0, arg1 any) *gomock.C
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "GetState", reflect.TypeOf((*MockTransactionContext)(nil).GetState), arg0, arg1)
 }
 
+// GetTransientState mocks base method.
+func (m *MockTransactionContext) GetTransientState(arg0 Address, arg1 Key) Value {
```
kjezek, line 698: I guess this is not related

go/carmen/carmen_mock.go
```diff
@@ -694,6 +694,20 @@ func (mr *MockTransactionContextMockRecorder) GetState(arg0, arg1 any) *gomock.C
...
+	m.ctrl.T.Helper()
+	ret := m.ctrl.Call(m, "GetTransientState", arg0, arg1)
+	ret0, _ := ret[0].(Value)
+	return ret0
+}
+
+// GetTransientState indicates an expected call of GetTransientState.
+func (mr *MockTransactionContextMockRecorder) GetTransientState(arg0, arg1 any) *gomock.Call {
```
kjezek, line 706: I guess this is not related

go/carmen/carmen_mock.go
```diff
@@ -799,6 +813,18 @@ func (mr *MockTransactionContextMockRecorder) SetState(arg0, arg1, arg2 any) *go
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "SetState", reflect.TypeOf((*MockTransactionContext)(nil).SetState), arg0, arg1, arg2)
 }
 
+// SetTransientState mocks base method.
+func (m *MockTransactionContext) SetTransientState(arg0 Address, arg1 Key, arg2 Value) {
```
kjezek, line 817: I guess this is not related

go/carmen/carmen_mock.go
```diff
@@ -799,6 +813,18 @@ func (mr *MockTransactionContextMockRecorder) SetState(arg0, arg1, arg2 any) *go
...
+	m.ctrl.T.Helper()
+	m.ctrl.Call(m, "SetTransientState", arg0, arg1, arg2)
+}
+
+// SetTransientState indicates an expected call of SetTransientState.
+func (mr *MockTransactionContextMockRecorder) SetTransientState(arg0, arg1, arg2 any) *gomock.Call {
```
kjezek, line 823: I guess this is not related

go/common/amount/amount.go
```diff
@@ -0,0 +1,139 @@
...
+// NewAmount creates a new U256 Amount from up to 4 uint64 arguments. The
+// arguments are given in the Big Endian order. No argument results in a value of zero.
+// The constructor panics if more than 4 arguments are given.
+func NewAmount(args ...uint64) Amount {
+	if len(args) > 4 {
+		panic("too many arguments")
+	}
+	result := Amount{}
+	offset := 4 - len(args)
+	for i := 0; i < len(args) && i < len(result.internal); i++ {
```
kjezek, line 34:
```suggestion
	for i := 0; i < len(args); i++ {
```

kjezek: correct me if I am wrong, but I guess this extra check is redundant, right? The input is max 4 items and the internal repres. of uin256.Int is also 4 items

go/state/state_db.go
```diff
@@ -596,84 +597,86 @@ func clone(val *big.Int) *big.Int {
...
-		s.SubBalance(addr, diff.Abs(diff))
-		return
-	}
 
 	oldValue := s.GetBalance(addr)
-	newValue := new(big.Int).Add(oldValue, diff)
+	newValue, overflow := amount.AddOverflow(oldValue, diff)
+
+	if overflow {
+		s.errors = append(s.errors, fmt.Errorf("failed to add balance for address %v: overflow", addr))
```
kjezek, line 631: this is dangerous, it will invalidate the whole carmen database. Since the overflow comes likely from a blockchain transaction, there is not way to recover, every re-run will end up in this same state. 

how is overflow dealt with in blockchains? I guess it is assumed it cannot happen.

kjezek (changes requested):
I think more investigation is needed how to deal with a possible overflow. I do not think it is right just to send carmen into an error state. 

I guess it is assumed in blockchains that overflow cannot happen - no-one has enough funds to overflow an account balance.  But please check this.

go/state/state_db.go
Mike-CZ: It is even impossible to happen. Total supply of FTM is ~3 billions. Max value of uint256 is 2^256 - 1.

go/common/amount/amount.go
Mike-CZ: Indeed, it might be simplified.

go/state/state_db.go
Mike-CZ: Yes, it should indeed not happen. The total supply is not even close to max value of uint256, so If one wallet held all tokens, it would still fit.

go/common/amount/amount.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+	result := uint256.Int{}
+	overflow := result.SetFromBig(b)
+	if overflow {
+		return Amount{}, fmt.Errorf("big.Int has more than 256 bits")
+	}
+	return Amount{internal: result}, nil
+}
+
+// Uint64 returns the amount as an uint64.
```
HerbertJordan, line 73:
```suggestion
// Uint64 returns the amount as an uint64. The result is only valid if `IsUint64()` returns true.
```

go/common/amount/amount.go
```diff
@@ -0,0 +1,139 @@
...
+
+// Amount is a 256-bit unsigned integer used for token values like balances.
+type Amount struct {
+	internal uint256.Int
+}
+
+// NewAmount creates a new U256 Amount from up to 4 uint64 arguments. The
+// arguments are given in the Big Endian order. No argument results in a value of zero.
+// The constructor panics if more than 4 arguments are given.
+func NewAmount(args ...uint64) Amount {
```
HerbertJordan, line 28: Maybe, since we have a dedicated package now, we can shorten the name of this function to `New`.
This would match `big.New` or `uint256.New`.

go/state/state_db.go
HerbertJordan: This is an inherent problem of the `StateDB` interface we adapted from `geth`.

I think in our design we should deprecate and eventually remove the `SubBalance` and `AddBalance` functions and replace them by a `SetBalance`. This way, it is the API client's job to deal with potential overflows.

HerbertJordan (dismissed):
Thanks for doing this refactoring. I really like how this turned out!

I think this really simplifies code at various places.

go/state/state_db.go
kjezek: yes, makes sense. But for this moment I guess we should not treat this situation as an error, right?

go/state/state_db.go
```diff
@@ -596,84 +597,86 @@ func clone(val *big.Int) *big.Int {
...
-	if diff.Sign() < 0 {
-		s.AddBalance(addr, diff.Abs(diff))
-		return
-	}
 
 	oldValue := s.GetBalance(addr)
-	newValue := new(big.Int).Sub(oldValue, diff)
+	newValue, underflow := amount.SubUnderflow(oldValue, diff)
+
+	if underflow {
```
kjezek, line 652: the same here, I guess we should just ignore this situation

kjezek (approved):
thanks. I guess we should just ignore overflow/underflow. Otherwise it looks good to me.

go/state/state_db.go
Mike-CZ: It is already treated as an error. As it used to be stored as `big.Int`, it was not checked immediately in Add/Sub methods. Instead it was checked in `EndBlock` method when the `big.Int` was converted to `Balance` with [ToBalance](https://github.com/Fantom-foundation/Carmen/blob/46353a6b7b9ab551a557f897e1175b755c5d70cc/go/common/types.go#L190) method. If the [conversion](https://github.com/Fantom-foundation/Carmen/blob/46353a6b7b9ab551a557f897e1175b755c5d70cc/go/state/state_db.go#L1193) failed, it produced an error.

kjezek: thanks for explanation. We can let it be as it is now I guess

kjezek (approved):

HerbertJordan (approved):

PR #899, opened 2024-06-09
fix embedded nodes rlp decoder

kjezek:
This PR fixes RLP decoder of embedded nodes. 

* It extends the testing framework to track embedded nodes
* It fixes decoding of embedded nodes, in particular they are encoded as an` rlp.List` which contains a value node, not an `rlp.String`, which was previously wrongly assumed. 
* furthermore decoding of common.Value was fixed in cases when prefix bytes contain zeros, the decoded value used to be wrongly aligned 
* tests were fixed to follow new changes
* new end-to-end tests were added, which tests the embedded nodes and values that are < 32bytes
* previous method that decodes embedded nodes was removed as embedded nodes are stored in standard rlp code and can be decoded by calling the common decoding method

### DIFF
--- a/go/database/mpt/hasher.go
+++ b/go/database/mpt/hasher.go
@@ -458,9 +458,11 @@ func (h ethHasher) getHash(ref *NodeReference, source NodeSource) (common.Hash, 
 		return common.Hash{}, err
 	}
 
-	// The hash for embedded nodes is 0.
+	// The hash for embedded nodes is the node representation.
 	if len(data) < 32 {
-		return common.Hash{}, nil
+		var hash common.Hash
+		copy(hash[:], data)
+		return hash, nil
 	}
 
 	return common.Keccak256(data), nil
--- a/go/database/mpt/rlp/rlp.go
+++ b/go/database/mpt/rlp/rlp.go
@@ -55,7 +55,15 @@ func EncodeInto(dst []byte, item Item) []byte {
 func Decode(rlp []byte) (Item, error) {
 	item, consumed, err := decode(rlp)
 	if consumed < uint64(len(rlp)) {
-		return nil, fmt.Errorf("unexpected trailing data: %x", rlp[consumed:])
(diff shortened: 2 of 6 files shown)


go/database/mpt/rlp/rlp.go
```diff
@@ -55,7 +55,11 @@ func EncodeInto(dst []byte, item Item) []byte {
 func Decode(rlp []byte) (Item, error) {
 	item, consumed, err := decode(rlp)
 	if consumed < uint64(len(rlp)) {
-		return nil, fmt.Errorf("unexpected trailing data: %x", rlp[consumed:])
+		for _, b := range rlp[consumed:] {
```
HerbertJordan, line 58: A line of comment justifying trailing zeros here would be useful. I assume it is due to the padding of embedded nodes?

go/database/mpt/hasher.go
```diff
@@ -458,9 +458,11 @@ func (h ethHasher) getHash(ref *NodeReference, source NodeSource) (common.Hash,
 		return common.Hash{}, err
 	}
 
-	// The hash for embedded nodes is 0.
+	// The hash for embedded nodes is the node representation.
 	if len(data) < 32 {
-		return common.Hash{}, nil
+		var hash common.Hash
```
HerbertJordan, line 463: Please add a unit test in `hasher_test.go` for this to document this requirement.

HerbertJordan (approved):

go/database/mpt/rlp/rlp.go
kjezek: np, added

go/database/mpt/hasher.go
kjezek: added, see` TestEthereumLikeHasher_EmbeddedNode_Hashes_As_Itself`

HerbertJordan (approved):

PR #902, opened 2024-06-10
allows for tracking embedded node in the node vistior

kjezek:
This PR enables tracking of embedded node in the node visitor. The flag is filled in the `visitPathTo` method.

This feature will be used in witness proof not to include embedded nodes as entries in the proof database.

### DIFF
--- a/go/database/mpt/nodes.go
+++ b/go/database/mpt/nodes.go
@@ -16,11 +16,11 @@ import (
 	"encoding/binary"
 	"errors"
 	"fmt"
+	"github.com/Fantom-foundation/Carmen/go/common"
+	"github.com/Fantom-foundation/Carmen/go/common/tribool"
+	"github.com/Fantom-foundation/Carmen/go/database/mpt/shared"
 	"io"
 	"slices"
-
-	"github.com/Fantom-foundation/Carmen/go/common"
-	"github.com/Fantom-foundation/Carmen/go/database/mpt/shared"
 )
 
 // This file defines the interface and implementation of all node types in a
@@ -285,6 +285,7 @@ func visitPathTo(source NodeSource, root *NodeReference, path []Nibble, address 
 	var last shared.ViewHandle[Node]
 	var found, done bool
 	var lastNodeId *NodeReference
+	var nextEmbedded, currentEmbedded bool
 	for !done {
 		handle, err := source.getViewAccess(nodeId)
 		if last.Valid() {
@@ -302,6 +303,7 @@ func visitPathTo(source NodeSource, root *NodeReference, path []Nibble, address 
 			if n.path.IsPrefixOf(path) {
 				nodeId = &n.next
 				path = path[n.path.Length():]
+				nextEmbedded = n.nextIsEmbedded
 				done = len(path) == 0
 			} else {
 				done = true
@@ -311,6 +313,7 @@ func visitPathTo(source NodeSource, root *NodeReference, path []Nibble, address 
 				done = true
 			} else {
 				nodeId = &n.children[path[0]]
+				nextEmbedded = n.isEmbedded(byte(path[0]))
 				path = path[1:]
 			}
 		case *AccountNode:
@@ -328,9 +331,10 @@ func visitPathTo(source NodeSource, root *NodeReference, path []Nibble, address 
 			return false, nil
 		}
 
-		if res := visitor.Visit(last.Get(), NodeInfo{Id: lastNodeId.Id()}); res != VisitResponseContinue {
+		if res := visitor.Visit(last.Get(), NodeInfo{Id: lastNodeId.Id(), Embedded: tribool.New(currentEmbedded)}); res != VisitResponseContinue {
 			done = true
 		}
+		currentEmbedded = nextEmbedded
 	}
 
 	last.Release()
--- a/go/database/mpt/nodes_test.go
+++ b/go/database/mpt/nodes_test.go
@@ -14,6 +14,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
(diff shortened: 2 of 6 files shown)


go/database/mpt/nodes.go
```diff
@@ -327,7 +330,7 @@ func visitPathTo(source NodeSource, root *NodeReference, path []Nibble, address
 			return false, nil
 		}
 
-		if res := visitor.Visit(last.Get(), NodeInfo{Id: lastNodeId.Id()}); res != VisitResponseContinue {
+		if res := visitor.Visit(last.Get(), NodeInfo{Id: lastNodeId.Id(), Embedded: nextEmbedded && found}); res != VisitResponseContinue {
```
HerbertJordan, line 333: Why is `nextEmbedded` combined with `found`? 

It is very hard for me to think about the semantic of this flag. If I write a Visitor, the NodeInfo`s "Embedded" field may lie to me?

go/database/mpt/nodes_test.go
```diff
@@ -6822,6 +6822,63 @@ func TestVisitPathToStorage_CanReachTerminalNodes(t *testing.T) {
...
+
+	ref, _ := ctxt.Build(desc)
+
+	var valueFound bool
+	visitor := NewMockNodeVisitor(ctrl)
+	visitor.EXPECT().Visit(gomock.Any(), gomock.Any()).DoAndReturn(func(node Node, info NodeInfo) VisitResponse {
+		var wantEmbedded bool
+		switch n := node.(type) {
+		case *AccountNode:
+			found, err := VisitPathToStorage(ctxt, &n.storage, key, visitor)
```
HerbertJordan, line 6853: If you visit another key with the same 40-byte prefix here (hard to generate because you would have to partially break keccak), the final "ValueNode" would not be indicated as embedded since `found` is false?

go/database/mpt/nodes.go
kjezek: Basically I re-used this flag to forward the `embedded` flag to next iteration. At some point a node (must be branch or ext) tells you that the next node will be embedded.  You have to save this information for the next loop to pass it to the visitor.  I used the `found` flag for it, assuming that the only option for the next iteration is that it must reach a correct value node, which will set the `found` flag to true.

As you write in the other comment, if you iterated to another value node (or even a different type of node), `found` would be false and the information passed to the visitor would be incorrect - but this is a situation which I guess cannot happen, i.e. I did not considered it. 

If you like, I can use one more bool flag serving only for the embedded flag, it may be more readable, I see that.

HerbertJordan: If you can improve the readability, please do so. Right now, it is very hard to wrap ones head around it.

kjezek: done, I introduced a new flag that is passed through. 

also I could create a test that brakes the previous version, i.e. it was not entirely correct assumption. .

go/database/mpt/nodes.go
```diff
@@ -327,9 +330,10 @@ func visitPathTo(source NodeSource, root *NodeReference, path []Nibble, address
 			return false, nil
 		}
 
-		if res := visitor.Visit(last.Get(), NodeInfo{Id: lastNodeId.Id()}); res != VisitResponseContinue {
+		if res := visitor.Visit(last.Get(), NodeInfo{Id: lastNodeId.Id(), Embedded: currentEmbedded}); res != VisitResponseContinue {
```
HerbertJordan, line 333: Thanks, this is way easier to understand :+1:

go/database/mpt/nodes_test.go
```diff
@@ -6822,6 +6822,86 @@ func TestVisitPathToStorage_CanReachTerminalNodes(t *testing.T) {
...
+		},
+		"mismatch embedded": {
+			desc: &Extension{
+				path: AddressToNibblePath(address, ctxt)[0:30],
+				next: &Account{address: address, pathLength: 34, info: AccountInfo{Nonce: common.Nonce{0x01}},
+					storage: &Extension{
+						path:         KeyToNibblePath(key, ctxt)[0:40],
+						nextEmbedded: true,
+						next:         &Value{key: key2, length: 24, value: value},
+					}}},
```
HerbertJordan, line 6859: These two setups are nearly identical - the only difference is the key used in the Value node, right?

How about replacing the `desc` field in the test specification with just the key and assemble the description in the test loop below based on the given key?

go/database/mpt/visitor.go
```diff
@@ -37,8 +37,9 @@ type NodeVisitor interface {
 }
 
 type NodeInfo struct {
-	Id    NodeId // the ID of the visited node
-	Depth *int   // the nesting level of the visited node, only set for tree visits
+	Id       NodeId // the ID of the visited node
+	Depth    *int   // the nesting level of the visited node, only set for tree visits
+	Embedded bool   // true if this node is embedded in another node, tracked in visitPathTo
```
HerbertJordan, line 42: It is not good that this field is only set in `visitPathTo`. 

Visitors should not be aware of the visiting process they are used in. That is part of their design philosophy. 
A fix might be to make it an optional value by turning it into a `*bool` like the `Depth` field above, or `Tribool`, which has the same meaning.

HerbertJordan (changes requested):

go/database/mpt/nodes_test.go
kjezek: good idea, done

go/database/mpt/visitor.go
kjezek: I would consider it ok when the value is not set and then the default is false. But ok, I changed it to Tribool

HerbertJordan (approved):

HerbertJordan (dismissed):

HerbertJordan (approved):

PR #903, opened 2024-06-10
proof addresses and keys

kjezek:
This PR implements proving of account addresses and keys agains the given witness proof.

### DIFF
--- a/go/database/mpt/proof.go
+++ b/go/database/mpt/proof.go
@@ -16,6 +16,7 @@ import (
 	"github.com/Fantom-foundation/Carmen/go/common"
 	"github.com/Fantom-foundation/Carmen/go/common/tribool"
 	"golang.org/x/exp/maps"
+	"slices"
 	"sort"
 	"strings"
 )
@@ -54,27 +55,27 @@ func (p WitnessProof) Add(other WitnessProof) {
 }
 
 // Extract extracts a sub-proof for a given account and selected storage locations from this proof.
-// It returns a copy that contains only the data necessary for proofing the given address and storage keys.
+// It returns a copy that contains only the data necessary for proving the given address and storage keys.
 // The resulting proof covers proofs for the intersection of the requested properties (account information and slots)
-// and the properties covered by this proof. The second return parameter indicates whether everything that was requested could be covered. If so
-// it is set to true, otherwise it is set to false.
+// and the properties covered by this proof. The second return parameter indicates whether everything that
+// was requested could be covered. If so it is set to true, otherwise it is set to false.
 func (p WitnessProof) Extract(root common.Hash, address common.Address, keys ...common.Key) (WitnessProof, bool) {
 	result := proofDb{}
 	visitor := &proofCollectingVisitor{visited: result}
-	found, err := visitWitnessPathTo(p.proofDb, root, addressToHashedNibbles(address), visitor)
+	found, complete, err := visitWitnessPathTo(p.proofDb, root, addressToHashedNibbles(address), visitor)
 	if err != nil || !found {
-		return WitnessProof{result}, found
+		return WitnessProof{result}, complete
 	}
 
 	storageRoot := visitor.visitedAccount.storageHash
 	for _, key := range keys {
-		foundKey, err := visitWitnessPathTo(p.proofDb, storageRoot, keyToHashedPathNibbles(key), visitor)
-		if err != nil || !foundKey {
-			found = false
+		_, completeKey, err := visitWitnessPathTo(p.proofDb, storageRoot, keyToHashedPathNibbles(key), visitor)
+		if err != nil || !completeKey {
+			complete = false
 		}
 	}
 
-	return WitnessProof{result}, found
+	return WitnessProof{result}, complete
 }
 
 // IsValid checks that this proof is self-consistent. If the result is true, the proof can be used
@@ -93,43 +94,68 @@ func (p WitnessProof) IsValid() bool {
 }
 
 // GetAccountInfo extracts an account info from the witness proof for the input root hash and the address.
-// If the witness proof contains an account for the input address, it returns its information.
-// If the proof does not contain an account, it returns false.
+// This method returns true, if the inputs could be proven. In this case, the first return parameter gives
+// the actual value. If the methods return false, the input could not be proved, and the returned value
+// is undefined.
 // The method may return an error if the proof is invalid.
 func (p WitnessProof) GetAccountInfo(root common.Hash, address common.Address) (AccountInfo, bool, error) {
-	panic("not implemented")
+	return witnessAccountFieldGetter(p.proofDb, root, address, func(n AccountNode) AccountInfo {
+		return n.Info()
+	})
 }
 
 // GetBalance extracts a balance from the witness proof for the input root hash and the address.
-// If the witness proof contains the requested account for the input address for the given root hash, it returns its balance.
-// If the proof does not cover the requested account, it returns false.
+// This method returns true, if the inputs could be proven. In this case, the first return parameter gives
+// the actual value. If the methods return false, the input could not be proved, and the returned value
+// is undefined.
 // The method may return an error if the proof is invalid.
 func (p WitnessProof) GetBalance(root common.Hash, address common.Address) (common.Balance, bool, error) {
-	panic("not implemented")
+	return witnessAccountFieldGetter(p.proofDb, root, address, func(n AccountNode) common.Balance {
+		return n.Info().Balance
+	})
 }
 
 // GetNonce extracts a nonce from the witness proof for the input root hash and the address.
-// If the witness proof contains the account for the input address, it returns its nonce.
-// If the proof does not contain the account, it returns false.
+// This method returns true, if the inputs could be proven. In this case, the first return parameter gives
+// the actual value. If the methods return false, the input could not be proved, and the returned value
+// is undefined.
 // The method may return an error if the proof is invalid.
 func (p WitnessProof) GetNonce(root common.Hash, address common.Address) (common.Nonce, bool, error) {
-	panic("not implemented")
+	return witnessAccountFieldGetter(p.proofDb, root, address, func(n AccountNode) common.Nonce {
+		return n.Info().Nonce
+	})
 }
 
 // GetCodeHash extracts a code hash from the witness proof for the input root hash and the address.
-// If the witness proof contains the account for the input address, it returns its code hash.
-// If the proof does not contain the account, it returns false.
+// This method returns true, if the inputs could be proven. In this case, the first return parameter gives
+// the actual value. If the methods return false, the input could not be proved, and the returned value
+// is undefined.
 // The method may return an error if the proof is invalid.
 func (p WitnessProof) GetCodeHash(root common.Hash, address common.Address) (common.Hash, bool, error) {
-	panic("not implemented")
+	return witnessAccountFieldGetter(p.proofDb, root, address, func(n AccountNode) common.Hash {
+		return n.Info().CodeHash
+	})
 }
 
 // GetState extracts a storage slot from the witness proof for the input root hash, account address and the storage key.
-// If the witness proof contains the input storage slot for the input key, it returns its value.
-// If the proof does not contain the slot, it returns false.
+// If the proof was complete, this method returns true, otherwise it returns false.
+// The proof was complete if it could fully determine either existence or non-existence of the slot.
+// In other words, it was possible to reach either a value node or an empty node.
 // The method may return an error if the proof is invalid.
 func (p WitnessProof) GetState(root common.Hash, address common.Address, key common.Key) (common.Value, bool, error) {
-	panic("not implemented")
+	visitor := &proofCollectingVisitor{}
+	found, complete, err := visitWitnessPathTo(p.proofDb, root, addressToHashedNibbles(address), visitor)
+	if err != nil || !found {
+		return common.Value{}, complete, err
+	}
+
+	storageRoot := visitor.visitedAccount.storageHash
+	found, complete, err = visitWitnessPathTo(p.proofDb, storageRoot, keyToHashedPathNibbles(key), visitor)
+	if err != nil || !found {
+		return common.Value{}, complete, err
+	}
+
+	return visitor.visitedValue.value, true, nil
 }
 
 // AllStatesZero checks that all storage slots are empty for the input root hash,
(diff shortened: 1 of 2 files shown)


go/database/mpt/proof_test.go
```diff
@@ -401,6 +402,139 @@ func TestWitnessProof_Extract_EmbeddedNode_In_Proof(t *testing.T) {
...
+			storage: &Extension{
+				path: KeyToNibblePath(key, ctxt)[0:40],
+				next: &Value{key: key, length: 24, value: value},
+			}},
+	}
+
+	root, node := ctxt.Build(desc)
+	totalProof := createReferenceProof(t, ctxt, &root, node)
+
+	handle := node.GetReadHandle()
```
HerbertJordan, line 485: nit: "View" permission required;

go/database/mpt/proof_test.go
```diff
@@ -401,6 +402,139 @@ func TestWitnessProof_Extract_EmbeddedNode_In_Proof(t *testing.T) {
...
+	defer handle.Release()
+
+	rootHash, dirty := handle.Get().GetHash()
+	if dirty {
+		t.Fatalf("expected node to be clean")
+	}
```
HerbertJordan, line 491: blocks like this could be shortened to
```suggestion
	rootHash, _ := ctxt.getHashFor(&root)
```

go/database/mpt/proof_test.go
```diff
@@ -401,6 +402,139 @@ func TestWitnessProof_Extract_EmbeddedNode_In_Proof(t *testing.T) {
...
+			if err != nil {
+				t.Fatalf("unexpected error: %v", err)
+			}
+			if exists {
+				t.Fatalf("proof should not exist")
+			}
+		})
+	}
+}
+
```
HerbertJordan, line 537: I think this test should pass, but it fails right now:
```
func TestWitnessProof_Proof_Zero_Nonces(t *testing.T) {
	ctrl := gomock.NewController(t)
	ctxt := newNodeContextWithConfig(t, ctrl, S5LiveConfig)

	root, node := ctxt.Build(&Account{
		pathLength: 64,
		info:       AccountInfo{Balance: common.Balance{1}},
	})

	rootHash, _ := ctxt.getHashFor(&root)
	totalProof := createReferenceProof(t, ctxt, &root, node)
	for _, addr := range []common.Address{{0}, {1}} { // one non-empty account and an empty account
		t.Run(fmt.Sprintf("account %v", addr), func(t *testing.T) {
			nonce, found, err := totalProof.GetNonce(rootHash, addr)
			if err != nil {
				t.Fatalf("unexpected error extracting proof: %v", err)
			}
			if !found {
				t.Errorf("proof should cover account")
			}
			if got, want := nonce, (common.Nonce{0}); got != want {
				t.Errorf("unexpected nonce: got %v, want %v", got, want)
			}
		})
	}
}
```

Can you have a look?

HerbertJordan (changes requested):
I added a test case that I think may point out an issue in the code. Please have a look.

go/database/mpt/proof_test.go
kjezek: Thanks for this test case. It helped to uncover one more issue comparing address/key converted to path. In some cases the address or key was padded with zero, while this zero is not included in the iterated path. I fixed it by comparing nibbles to nibbles, not Path instances.

Furthermore, the test did not pass because at the moment there is limitation of address size to 40 nibbles, while this test needed 64 nibbles.  It is to be fixed before release. 

Finally, I modified the test and impl. and at the moment it passes for the address `0` . It still fails for the address `1`, but here I do not understand why it should pass? This account is not in the proof and for this reason `false` is returned.

HerbertJordan: It should pass for the account `1` because all the information needed to proof that its nonce is `0` is in the proof.

The argument goes as follows:
 - The proof contains a root node which encodes the account information for address `0`. 
 - The fact that this is the root node guarantees that there is no account node in the MPT covering account `1`.  
 - The default nonce value of every account that is not in an MPT is 0;

Thus, in this case, the proof for account `0` also covers the information of all other possible accounts.

This example is just the smallest possible example of a proof for missing information. You can also argue it the other way around:
- create an MPT with a single account `0` with a positive balance
- ask the MPT for a proof about the nonce of account `1`
- the resulting proof should be equivalent to the proof for the nonce of account `0`, which is what is extracted in the test.

kjezek: thanks for explanation. I think the point of misunderstanding was that I thought that the resulting `bool` mens if the requested account/slot is in the proof. But it should represent if the proof is complete, or not, right.

In practices, should it work that everytime you ask for an account or storage, and it can find any `AccountNode`, `ValueNode`, or `EmptyNode`, it returns true.  Otherwise - `Branch`/`Extension` Node it returns false?

I guess it should work the same for the method that `Extract`s the proof - for consistency. Right?

kjezek: I implemented it in this commit: b948266c739b9814166414735a4d66d503824300 

please have a look if it is what you meant

HerbertJordan: The traversal through the proof may also end in an `Extension` node and data can be derived from that.

For instance, assume you have two accounts A and B such that their hashes have the first 2 bytes in common -- this is a chance of 1 : 65536, so not that unlikely. A trie with those two nodes would look like this:
```
Extension{
  path: {a, b}
  next: Branch{
     c: Account{ address: A },
     d: Account{ address: B },
  }
}
```
For this, if I asked for a proof for an account C, where C's hash starts with an `aa`, the navigation would get me to the extension node and then it should terminate. The proof only contains the extension node.

When then I ask the proof for C for C's nonce, the algorithm would navigate through the proof and stop in the Extension node. There, it concludes that since the extension-nodes path is `ab` there is no account with a prefix `aa` in the trie, and thus that C`s state is empty. Consequently, C`s nonce is `0`.

To create a test case for this, you might want to use the two keys from [here](https://github.com/Fantom-foundation/Carmen/blob/ad1657431037430a60f7456088947e5bfd16972a/go/database/mpt/verification_test.go#L577-L584). Back, when I created it, I did some brute-forcing to find a key pair with a long common prefix, and this was the result.

go/database/mpt/proof.go
```diff
@@ -93,43 +94,68 @@ func (p WitnessProof) IsValid() bool {
 }
 
 // GetAccountInfo extracts an account info from the witness proof for the input root hash and the address.
-// If the witness proof contains an account for the input address, it returns its information.
-// If the proof does not contain an account, it returns false.
+// If the proof was complete, this method returns true, otherwise it returns false.
+// The proof was complete if it could fully determine either existence or non-existence of the account
+// in the proof. In other words, it was possible to reach either an account node or an empty node.
```
HerbertJordan, line 99: nit: I think this describes implementation details not really relevant for the user of this API.

The essential thing is:
- the `bool` is true => the returned info is proven and can be used
- the `bool` is false or an error is produced => the returned info must not be used

go/database/mpt/proof.go
```diff
@@ -268,32 +316,72 @@ type witnessProofVisitor interface {
 	Visit(hash common.Hash, rlpNode rlpEncodedNode, node Node, isEmbedded bool)
 }
 
+type latVisitedNode int
```
HerbertJordan, line 319: typo: `lat` => `last`

go/database/mpt/proof.go
```diff
@@ -268,32 +316,72 @@ type witnessProofVisitor interface {
...
+
+const (
+	// latVisitedNodeDefault is the default value for the last visited node.
+	lastVisitedNodeDefault latVisitedNode = iota
+	// latsVisitedNodeAccount is the last visited account node.
+	latsVisitedNodeAccount
```
HerbertJordan, line 325: typo: `lats` => `last`

go/database/mpt/proof_test.go
HerbertJordan: Also, thinking about it, a `Branch` node can also never be the terminal node.

The only time when you should return `false` is if you do a lookup in the proof map and there is a node missing. Simple as that.

kjezek: good point, I got it. It will simplify the implementation a lot

go/database/mpt/proof.go
kjezek: ok, make sense, updated

go/database/mpt/proof.go
kjezek: removed

go/database/mpt/proof.go
kjezek: removed

go/database/mpt/proof_test.go
```diff
@@ -402,6 +384,276 @@ func TestWitnessProof_Extract_EmbeddedNode_In_Proof(t *testing.T) {
...
+				t.Fatalf("proof should be complete")
+			}
+		})
+		t.Run(fmt.Sprintf("%s invalid", name), func(t *testing.T) {
+			_, err := test.test(t, corruptedProof, rootHash, address, key)
+			if err == nil {
+				t.Fatalf("expected error")
+			}
+		})
+	}
```
HerbertJordan, line 492: Code duplication in these tests could be removed like this (also helps readability):
```suggestion
		tests := map[string]struct {
		get  func(WitnessProof, common.Hash) (any, bool, error)
		want any
	}{
		"GetAccountInfo": {
			get: func(proof WitnessProof, root common.Hash) (any, bool, error) {
				return proof.GetAccountInfo(root, address)
			},
			want: info,
		},
		"GetNonce": {
			get: func(proof WitnessProof, root common.Hash) (any, bool, error) {
				return proof.GetNonce(root, address)
			},
			want: nonce,
		},
		"GetBalance": {
			get: func(proof WitnessProof, root common.Hash) (any, bool, error) {
				return proof.GetBalance(root, address)
			},
			want: balance,
		},
		"GetCodeHash": {
			get: func(proof WitnessProof, root common.Hash) (any, bool, error) {
				return proof.GetCodeHash(root, address)
			},
			want: hash,
		},
		"GetValue": {
			get: func(proof WitnessProof, root common.Hash) (any, bool, error) {
				return proof.GetState(root, address, key)
			},
			want: value,
		},
	}

	desc := &Extension{
		path: AddressToNibblePath(address, ctxt)[0:30],
		next: &Account{address: address, pathLength: 34, info: info,
			storage: &Extension{
				path: KeyToNibblePath(key, ctxt)[0:40],
				next: &Value{key: key, length: 24, value: value},
			}},
	}

	root, node := ctxt.Build(desc)
	totalProof := createReferenceProof(t, ctxt, &root, node)
	rootHash, _ := ctxt.getHashFor(&root)

	// corrupt non-root node in the proof
	corruptedProof := WitnessProof{maps.Clone(totalProof.proofDb)}
	for k := range corruptedProof.proofDb {
		if k != rootHash {
			corruptedProof.proofDb[k] = []byte{0xAA, 0xBB, 0xCC, 0xDD}
		}
	}

	for name, test := range tests {
		t.Run(name, func(t *testing.T) {
			got, complete, err := test.get(totalProof, rootHash)
			if err != nil {
				t.Fatalf("unexpected error: %v", err)
			}
			if !complete {
				t.Fatalf("proof should be complete")
			}
			if got, want := got, test.want; !reflect.DeepEqual(got, want) {
				t.Errorf("unexpected value: got %v, want %v", got, want)
			}
		})
		t.Run(fmt.Sprintf("%s invalid", name), func(t *testing.T) {
			_, _, err := test.get(corruptedProof, rootHash)
			if err == nil {
				t.Fatalf("expected error")
			}
		})
	}
```

go/database/mpt/proof_test.go
```diff
@@ -402,6 +384,276 @@ func TestWitnessProof_Extract_EmbeddedNode_In_Proof(t *testing.T) {
...
+	differentAddress := common.Address{2}
+	key := common.Key{2}
+	info := AccountInfo{Nonce: common.Nonce{1}, Balance: common.Balance{1}, CodeHash: common.Hash{1}}
+
+	ctxt := newNodeContextWithConfig(t, ctrl, S5LiveConfig)
+
+	testGetters := map[string]struct {
+		test func(*testing.T, WitnessProof, common.Hash, common.Address, common.Key) (bool, error)
+	}{
+		"GetAccountInfo": {test: func(t *testing.T, proof WitnessProof, rootHash common.Hash, address common.Address, key common.Key) (bool, error) {
```
HerbertJordan, line 507: I guess, as for the test above, the common parts of those `test` functions could be factored out.

HerbertJordan (changes requested):
Looks way leaner now - just one detail in the Extract function to be checked.

go/database/mpt/proof.go
```diff
@@ -236,26 +280,26 @@ func visitWitnessPathTo(source proofDb, root common.Hash, path []Nibble, visitor
...
-			return false, nil // EmptyNode -> do not visit, and terminate
+			return false, true, nil // EmptyNode -> do not visit, and terminate, proof is complete
 		}
 
 		visitor.Visit(nodeHash, rlpNode, node, found && nextEmbedded)
 		nodeHash = nextHash
 	}
 
-	return found, nil
+	return found, true, nil // node was found, and the proof is complete
```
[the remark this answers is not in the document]
kjezek: ok, I removed this comment, I guess the semantics is not that complicated

go/database/mpt/proof.go
```diff
@@ -54,21 +55,21 @@ func (p WitnessProof) Add(other WitnessProof) {
...
 // The resulting proof covers proofs for the intersection of the requested properties (account information and slots)
 // and the properties covered by this proof. The second return parameter indicates whether everything that was requested could be covered. If so
 // it is set to true, otherwise it is set to false.
 func (p WitnessProof) Extract(root common.Hash, address common.Address, keys ...common.Key) (WitnessProof, bool) {
 	result := proofDb{}
 	visitor := &proofCollectingVisitor{visited: result}
-	found, err := visitWitnessPathTo(p.proofDb, root, addressToHashedNibbles(address), visitor)
+	found, _, err := visitWitnessPathTo(p.proofDb, root, addressToHashedNibbles(address), visitor)
 	if err != nil || !found {
 		return WitnessProof{result}, found
```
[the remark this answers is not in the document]
kjezek: ok, updated to return `complete`, not `found`

go/database/mpt/proof_test.go
kjezek: done

HerbertJordan (approved):

PR #904, opened 2024-06-10
implements extraction and merge of witness proof

kjezek:
This PR implements proof manipulation utils, in particular it offers:
* Extraction of a proof from another proof based on address and key(s)
* Merge of proofs

### DIFF
--- a/go/database/mpt/nibble.go
+++ b/go/database/mpt/nibble.go
@@ -64,6 +64,24 @@ func KeyToNibblePath(key common.Key, source NodeSource) []Nibble {
 	return res
 }
 
+// addressToHashedNibbles converts the given path into a slice of Nibbles.
+// It always hashes the path before converting it.
+func addressToHashedNibbles(address common.Address) []Nibble {
+	path := common.Keccak256(address[:])
+	res := make([]Nibble, len(path)*2)
+	parseNibbles(res, path[:])
+	return res
+}
+
+// keyToHashedPathNibbles converts the given path into a slice of Nibbles.
+// It always hashes the path before converting it.
+func keyToHashedPathNibbles(key common.Key) []Nibble {
+	path := common.Keccak256(key[:])
+	res := make([]Nibble, len(path)*2)
+	parseNibbles(res, path[:])
+	return res
+}
+
 func parseNibbles(dst []Nibble, src []byte) {
 	for i := 0; i < len(src); i++ {
 		dst[2*i] = Nibble(src[i] >> 4)
--- a/go/database/mpt/path.go
+++ b/go/database/mpt/path.go
@@ -100,6 +100,11 @@ func (p *Path) IsPrefixOf(list []Nibble) bool {
 	return p.GetCommonPrefixLength(list) == int(p.length)
 }
 
+// IsEqualTo determines whether the given nibble sequence is equal to this path.
+func (p *Path) IsEqualTo(list []Nibble) bool {
+	return p.Length() == len(list) && p.GetCommonPrefixLength(list) == int(p.length)
+}
+
 // GetCommonPrefixLength determines the common prefix of the given Nibble
 // slice and this path.
 func (p *Path) GetCommonPrefixLength(list []Nibble) int {
--- a/go/database/mpt/proof.go
+++ b/go/database/mpt/proof.go
@@ -11,14 +11,24 @@
 package mpt
 
 import (
+	"bytes"
+	"fmt"
 	"github.com/Fantom-foundation/Carmen/go/common"
 	"github.com/Fantom-foundation/Carmen/go/common/tribool"
+	"golang.org/x/exp/maps"
+	"sort"
+	"strings"
 )
 
-//go:generate mockgen -source proof.go -destination proof_mocks.go -package witness
+//go:generate mockgen -source proof.go -destination proof_mocks.go -package mpt
 
 // rlpEncodedNode is an RLP encoded MPT node.
 type rlpEncodedNode []byte
+
+// rlpEncodedNodeEquals returns true if the two RLP encoded nodes are equal.
+func rlpEncodedNodeEquals(a, b rlpEncodedNode) bool {
+	return bytes.Equal(a, b)
+}
 
 // proofDb is a database of RLP encoded MPT nodes and their hashes that represent witness proofs.
 type proofDb map[common.Hash]rlpEncodedNode
@@ -38,7 +48,9 @@ func CreateWitnessProof(nodeSource NodeSource, root *NodeReference, address comm
 
 // Add merges the input witness proof into the current witness proof.
 func (p WitnessProof) Add(other WitnessProof) {
-	panic("not implemented")
+	for k, v := range other.proofDb {
+		p.proofDb[k] = v
+	}
 }
 
 // Extract extracts a sub-proof for a given account and selected storage locations from this proof.
@@ -47,7 +59,22 @@ func (p WitnessProof) Add(other WitnessProof) {
 // and the properties covered by this proof. The second return parameter indicates whether everything that was requested could be covered. If so
 // it is set to true, otherwise it is set to false.
 func (p WitnessProof) Extract(root common.Hash, address common.Address, keys ...common.Key) (WitnessProof, bool) {
-	panic("not implemented")
+	result := proofDb{}
+	visitor := &proofCollectingVisitor{visited: result}
+	found, err := visitWitnessPathTo(p.proofDb, root, addressToHashedNibbles(address), visitor)
+	if err != nil || !found {
+		return WitnessProof{result}, found
+	}
+
+	storageRoot := visitor.visitedAccount.storageHash
+	for _, key := range keys {
+		foundKey, err := visitWitnessPathTo(p.proofDb, storageRoot, keyToHashedPathNibbles(key), visitor)
+		if err != nil || !foundKey {
+			found = false
+		}
+	}
+
+	return WitnessProof{result}, found
 }
 
 // IsValid checks that this proof is self-consistent. If the result is true, the proof can be used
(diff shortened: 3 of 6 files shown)


go/database/mpt/path.go
```diff
@@ -100,6 +100,11 @@ func (p *Path) IsPrefixOf(list []Nibble) bool {
 	return p.GetCommonPrefixLength(list) == int(p.length)
 }
 
+// IsEqualTo determines whether the given nibble sequence is equal to this path.
+func (p *Path) IsEqualTo(list []Nibble) bool {
```
HerbertJordan, line 104: Can you add a unit test for this?

go/database/mpt/nibble.go
```diff
@@ -64,6 +64,24 @@ func KeyToNibblePath(key common.Key, source NodeSource) []Nibble {
 	return res
 }
 
+// addressToHashedPath converts the given path into a slice of Nibbles.
```
HerbertJordan, line 67: It might be reasonable to add a few small unit tests for those helper functions.

go/database/mpt/proof.go
```diff
@@ -115,3 +139,135 @@ func (p WitnessProof) AllStatesZero(root common.Hash, address common.Address, fr
 func (p WitnessProof) AllAddressesEmpty(root common.Hash, from, to common.Address) (tribool.Tribool, error) {
 	panic("not implemented")
 }
+
+// Equals returns true if the two witness proofs are equal.
+func (p WitnessProof) Equals(other WitnessProof) bool {
+	return maps.EqualFunc(p.proofDb, other.proofDb, rlpEncodedNodeEquals)
+}
+
+// MergeProofs merges the input witness proofs into the first witness proof.
```
HerbertJordan, line 148: nit: The comment seems to be outdated; the function now supports a list of proofs to be merged, which is nice.

go/database/mpt/proof.go
```diff
@@ -115,3 +139,135 @@ func (p WitnessProof) AllStatesZero(root common.Hash, address common.Address, fr
...
+// If no more nodes are available on the path, the execution ends.
+// If the path does not exist, the function returns false.
+// The function returns an error if the path cannot be iterated due to error propagated from the input proof.
+// When the function reaches either an account node or a value node it is compared to the remaining input path
+// that was not iterated yet.
+// If the path matches, the function terminates and returns true.
+// It means this function can be used to find either an account node or a value node,
+// but it cannot find both at the same time.
+func visitWitnessPathTo(source proofDb, root common.Hash, path []Nibble, visitor witnessProofVisitor) (bool, error) {
+	nodeId := root
```
HerbertJordan, line 169: nit: in this package, there is a type `NodeID` which is a different concept that what it is used for here. Maybe rename to `nodeHash`?

go/database/mpt/proof.go
```diff
@@ -115,3 +139,135 @@ func (p WitnessProof) AllStatesZero(root common.Hash, address common.Address, fr
...
+			if !exists {
+				return false, nil
+			}
+		}
+		node, err := DecodeFromRlp(rlpNode)
+		if err != nil {
+			return false, err
+		}
+
+		var nextId common.Hash
```
HerbertJordan, line 189: nit: `nextId` => `nextHash`?

go/database/mpt/proof.go
```diff
@@ -115,3 +139,135 @@ func (p WitnessProof) AllStatesZero(root common.Hash, address common.Address, fr
...
+			}
+			done = true
+		default:
+			done = true
+		}
+
+		// visit when we are in the middle of the path or when we found the result
+		if !done || found {
+			visitor.Visit(nodeId, rlpNode, node, found && nextEmbedded)
+		}
```
HerbertJordan, line 226: If I understand correctly, in all cases where the path was leading to a wrong node (e.g. an `ExtensionNode` with the wrong prefix or a leaf node with the wrong path), this final node is not included in the proof. Is this right?

This final "wrong direction" node, however, should be included. It is needed to proof that what was asked for is not present, and thus has the default value (zero, empty code, or empty storage).

go/database/mpt/proof.go
```diff
@@ -115,3 +139,135 @@ func (p WitnessProof) AllStatesZero(root common.Hash, address common.Address, fr
...
+		nodeId = nextId
+	}
+
+	return found, nil
+}
+
+// witnessProofVisitor is a visitor that visits witness prof nodes.
```
HerbertJordan, line 233:
```suggestion
// witnessProofVisitor is a visitor that visits witness proof nodes.
```

go/database/mpt/proof.go
```diff
@@ -115,3 +139,135 @@ func (p WitnessProof) AllStatesZero(root common.Hash, address common.Address, fr
...
+type witnessProofVisitor interface {
+
+	// Visit visits the witness proof node.
+	// It provides the hash of the node, the RLP encoded node and the node itself.
+	// The method returns a response that determines the further processing of the visit.
+	Visit(hash common.Hash, rlpNode rlpEncodedNode, node Node, isEmbedded bool) VisitResponse
+}
+
+type proofCollectingVisitor struct {
+	visited        proofDb     // all visited witness proofs
```
HerbertJordan, line 245:
```suggestion
	visited        proofDb     // all visited nodes
```

go/database/mpt/nibble.go
```diff
@@ -64,6 +64,24 @@ func KeyToNibblePath(key common.Key, source NodeSource) []Nibble {
...
+// It always hashes the path before converting it.
+func addressToHashedPath(address common.Address) []Nibble {
```
HerbertJordan, line 69: The other functions in this file refer to `[]Nibble` as a `NibblePath` since `Path` is a different type in this package.

go/database/mpt/proof_test.go
```diff
@@ -0,0 +1,341 @@
...
+	})
+
+	rootHandle := node.GetReadHandle()
+	rootHash, dirty := rootHandle.Get().GetHash()
+	if dirty {
+		t.Fatalf("expected node to be clean")
+	}
+	rootHandle.Release()
+
+	// create following shadow proofs
```
HerbertJordan, line 80: what is a "shadow proof"?

go/database/mpt/proof_test.go
```diff
@@ -0,0 +1,341 @@
...
+	}
+
+	t.Run("Extract", func(t *testing.T) {
+		// Test proofs can be extracted and that they match the shadow proofs
+		extractedProofAddress1Key1, exists := totalProof.Extract(rootHash, address1, key1)
+		if !exists {
+			t.Errorf("proof for %v %v %v not found", rootHash, address1, key1)
+		}
+		if got, want := extractedProofAddress1Key1, address1Key1ProofShadow; !got.Equals(want) {
+			t.Errorf("unexpected proof: got %v, want %v", got, want)
```
HerbertJordan, line 111: nit: do proofs have a human readable print format? Otherwise, this output may be messy ... in particular since the content is a map, and their output is in random order afaik ...

go/database/mpt/proof_test.go
```diff
@@ -0,0 +1,341 @@
...
+	"testing"
+)
+
+func TestWitnessProof_Extract_and_Merge_Proofs(t *testing.T) {
+	ctrl := gomock.NewController(t)
+
+	address1 := common.Address{0xAB, 0xCD, 0xEF}
+	address2 := common.Address{0xAF, 0xCD, 0xEF}
+	key1 := common.Key{0x12, 0x34, 0x56, 0x78}
+	key2 := common.Key{0x23, 0x45, 0x67, 0x78}
```
HerbertJordan, line 26: nit: since addresses and keys are hashed, simpler values like
```suggestion
	address1 := common.Address{1}
	address2 := common.Address{2}
	key1 := common.Key{1}
	key2 := common.Key{2}
```
would be sufficient. The extra size leaves the impression that those values are essential for the test.

go/database/mpt/proof_test.go
```diff
@@ -0,0 +1,341 @@
...
+			t.Errorf("unexpected proof: got %v, want %v", got, want)
+		}
+	})
+}
+
+func TestWitnessProof_Extract_Various_NodeTypes_NotFoundProofs(t *testing.T) {
+	ctrl := gomock.NewController(t)
+
+	address := common.Address{0xAB, 0xCD, 0xEF}
+	key := common.Key{0x12, 0x34, 0x56, 0x78}
```
HerbertJordan, line 135: Same here, the test basically works with any address or key.
```suggestion
	address := common.Address{}
	key := common.Key{}
```

go/database/mpt/proof_test.go
```diff
@@ -0,0 +1,341 @@
...
+							},
+						}},
+					}},
+				},
+			}},
+		}},
+	})
+
+	rootHandle := node.GetReadHandle()
+	rootHash, dirty := rootHandle.Get().GetHash()
```
HerbertJordan, line 74: Note: here and below, technically "[View](https://github.com/Fantom-foundation/Carmen/blob/54a30889e0311491d62d88d08a965e91f91be69d/go/database/mpt/shared/shared.go#L27)" Access is required to access the hash. It does not make a difference here, but when operating on the real MPT it will. Thus, I would prefer to keep this consistent.

go/database/mpt/proof_test.go
```diff
@@ -0,0 +1,341 @@
...
+	proof, exists := totalProof.Extract(rootHash, address, key)
+	if !exists {
+		t.Errorf("proof should exist")
+	}
+	if got, want := proof, totalProof; !got.Equals(want) {
+		t.Errorf("unexpected proof: got %v, want %v", got, want)
+	}
+}
+
+// testCreateShadowProof creates a shadow witness proof for the given root node.
```
HerbertJordan, line 294: maybe rename "shadow" to "reference"?

go/database/mpt/proof_test.go
```diff
@@ -0,0 +1,341 @@
...
+// The shadow proof is created simply that all nodes in the MPT subtree are stored in the proof.
+// Only Empty nodes are excluded.
+func testCreateShadowProofForLabels(t *testing.T, ctxt *nodeContext, labels ...string) WitnessProof {
```
HerbertJordan, line 297: nit: rename to `createShadowProofForLabels` or `createReferenceProofForLabels`?

The `test` prefix seems unnecessary and I initially read it as it would test the creation of that proof ..

HerbertJordan (changes requested):
Overall I think it looks good, just some minor remarks besides one semantic issue: I think the last node of a path to a non-contained node that indicates the lack of the present of the requested node must be included in the proof. I added a comment regarding this at the corresponding code section.

go/database/mpt/proof.go
kjezek: Yes, I thought that it was requirement not to include the final not matching node - I remember I was adding this condition after a review round. 

No problem to include the node, it actually simplifies the number of bool conditions. 

Reg. proving non-presence, i think it does not matter for proving a particular key, because when iterating the proof from root, the path will will not be found anyway does not matter if it is because the final node is not in the proof, or the node is not the fitting one.  

I guess a missing node may be needed for proving range of keys right?  We create a a proof for 
key1: 0xAA
key2: 0xAD

when we want to check this as a range, the now missing node could contain a path for 0xAB, 

right?

go/database/mpt/nibble.go
kjezek: I do not understand this comment

go/database/mpt/nibble.go
kjezek: added. Just I created basically a new function in tests that hashes and converts to nibble. I do not know how to test it better

go/database/mpt/proof.go
kjezek: fixed

go/database/mpt/proof.go
kjezek: fixed

go/database/mpt/proof.go
kjezek: fixed

go/database/mpt/proof_test.go
kjezek: right, renamed not to use `shadow`

go/database/mpt/proof_test.go
kjezek: fixed

go/database/mpt/proof_test.go
kjezek: done

go/database/mpt/proof_test.go
kjezek: done

go/database/mpt/path.go
kjezek: added

go/database/mpt/proof.go
kjezek: ok, I think I updated the implementation to include the terminal nodes as well, except EmptyNode. 

It is also tested in the test: https://github.com/Fantom-foundation/Carmen/pull/904/files#diff-4324ecc3cdc67efb6e6fd6d00022d12c3345c7a1fdfe278090825c2a7062c60bR189

go/database/mpt/proof_test.go
kjezek: I added String(): https://github.com/Fantom-foundation/Carmen/pull/904/files#diff-14bcdb7e916407fed7057b15c0b1bb5cd85247c8542c941d7ad5065ae454f5b8R160

go/database/mpt/proof_test.go
```diff
@@ -0,0 +1,468 @@
...
+		"branch without account": {
+			trie: &Tag{"A", &Branch{children: Children{
+				0x1: &Tag{"B", &Empty{}},
+				0x2: &Tag{"C", &Account{address: address}},
+			}}},
+			path: []string{"A"},
+		},
+		"branch with wrong account": {
+			trie: &Tag{"A", &Branch{children: Children{
+				nibbles[0]: &Tag{"B", &Account{address: address, pathLength: 1}},
```
HerbertJordan, line 227: Maybe instead of changing the path length change the address?

HerbertJordan (dismissed):
Looks good, thanks!

go/database/mpt/proof_test.go
```diff
@@ -0,0 +1,468 @@
...
+			path: []string{},
+		},
+		"wrong account": {
+			trie: &Tag{"A", &Account{}},
+			path: []string{"A"},
+		},
+		"branch without account": {
+			trie: &Tag{"A", &Branch{children: Children{
+				0x1: &Tag{"B", &Empty{}},
+				0x2: &Tag{"C", &Account{address: address}},
```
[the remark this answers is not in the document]
kjezek: It does not matter, the path lead to an empty node, this account is here only for the branch not being empty

go/database/mpt/proof_test.go
```diff
@@ -0,0 +1,468 @@
...
+				nibbles[0]: &Tag{"B", &Account{address: address, pathLength: 1}},
+			}}},
+			path: []string{"A", "B"},
+		},
+		"extension with common prefix lead to empty": {
+			trie: &Tag{"A", &Extension{
+				path: nibbles[0:63],
+				next: &Tag{"B", &Branch{children: Children{
+					nibbles[63]: &Tag{"C", &Empty{}},
+					0x0:         &Tag{"D", &Account{address: address}},
```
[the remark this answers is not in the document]
kjezek: There must be at least one non-empty child of the account, otherwise it the branch node is encoded as an empty node, not the branch

go/database/mpt/proof_test.go
kjezek: ok, changed

go/database/mpt/proof_test.go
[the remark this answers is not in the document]
HerbertJordan: I guess what is happening is that the Branch-Node becomes an embedded node, which in a real-world setup is impossible. Adding this one node prevents that; so I guess it is fine.

HerbertJordan (approved):

PR #906, opened 2024-06-11
Add HasEmptyStorage func to LiveTrie.

petr-hanzl:
This PR adds functionality to ask whether given account has empty storage using `HasEmptyStorage` implemented into `Node`, `Forest` and `LiveTrie`.


go/database/mpt/nodes.go
```diff
@@ -166,6 +167,9 @@ type Node interface {
 	// parameter information and return values see SetValue().
 	ClearStorage(manager NodeManager, thisRef *NodeReference, this shared.WriteHandle[Node], address common.Address, path []Nibble) (newRoot NodeReference, changed bool, err error)
 
+	// HasEmptyStorage returns whether source node has empty storage.
+	HasEmptyStorage(source NodeSource, address common.Address, path []Nibble) (bool, error)
```
kjezek, line 171: Instead of adding a method to this interface, the same functionality may be implemented by iterating nodes using this method: https://github.com/Fantom-foundation/Carmen/blob/main/go/database/mpt/nodes.go#L256

I am not 100% sure what is the best option, but it looks to me that we may have over time requests to provide various fields from the nodes and all such getters would do the same - iterate the nodes via descending to a child node, just to reach the requested node and get the field. 

It looks to me as better to use the method mentioned above for this. Wdyt?

go/database/mpt/forest_test.go
```diff
@@ -2155,6 +2155,54 @@ func TestForest_VisitPathToStorage(t *testing.T) {
 	}
 }
 
+func TestForest_HasEmptyStorage(t *testing.T) {
+	for _, variant := range variants {
+		for _, config := range allMptConfigs {
+			for forestConfigName, forestConfig := range forestConfigs {
+				t.Run(fmt.Sprintf("%s-%s-%s", variant.name, config.Name, forestConfigName), func(t *testing.T) {
```
kjezek, line 2162: I think it should also test situations where a storage is deleted either by setting an empty value or calling `ClearStorage`

kjezek: and when deleting the whole account

go/database/mpt/live_trie_test.go
```diff
@@ -893,6 +893,57 @@ func TestLiveTrie_VerificationOfLiveTrieWithCorruptedFileFails(t *testing.T) {
 	}
 }
 
+func TestLiveTrie_HasEmptyStorage(t *testing.T) {
+	for _, config := range allMptConfigs {
+		t.Run(config.Name, func(t *testing.T) {
```
kjezek, line 898: it should also test when the storage emptied by other ways: ClearStorage, set Value to zero

go/database/mpt/nodes.go
petr-hanzl: Good idea. I think this could even simplify the code. Please do check the usage, Im quite not sure whether I used it correctly although the tests pass.

go/database/mpt/forest_test.go
kjezek: When you set an empty AccountInfo: https://github.com/Fantom-foundation/Carmen/blob/main/go/database/mpt/nodes.go#L1546

go/database/mpt/forest.go
```diff
@@ -383,6 +383,24 @@ func (s *Forest) SetValue(rootRef *NodeReference, addr common.Address, key commo
 	return newRoot, err
 }
 
+func (s *Forest) HasEmptyStorage(rootRef *NodeReference, addr common.Address) (isEmpty bool, err error) {
+	v := &lambdaVisitor{
```
kjezek, line 387: there is a constructor for this visitor: https://github.com/Fantom-foundation/Carmen/blob/main/go/database/mpt/visitor.go#L71

go/database/mpt/forest.go
```diff
@@ -383,6 +383,24 @@ func (s *Forest) SetValue(rootRef *NodeReference, addr common.Address, key commo
...
+		visit: func(node Node, info NodeInfo) VisitResponse {
+			isEmpty = node.(*AccountNode).storage.Id().IsEmpty()
```
kjezek, line 389: this visitor provides all nodes on the path. I.e. it should check that the cast is possible that the node is of the `AccountNode` type.  The tests must be improved if they pass at the moment.

go/database/mpt/forest.go
```diff
@@ -383,6 +383,24 @@ func (s *Forest) SetValue(rootRef *NodeReference, addr common.Address, key commo
...
+	}
+	exists, err := VisitPathToAccount(s, rootRef, addr, v)
+	if err != nil {
+		return true, err
+	}
+	if !exists {
+		return true, nil
+	}
+
+	return isEmpty, nil
```
kjezek, line 401:
```suggestion
	return isEmpty || !exists, err
```

go/database/mpt/live_trie_test.go
```diff
@@ -893,6 +893,100 @@ func TestLiveTrie_VerificationOfLiveTrieWithCorruptedFileFails(t *testing.T) {
 	}
 }
 
+func TestLiveTrie_HasEmptyStorage(t *testing.T) {
```
kjezek, line 896: this test seems a copy+paste of the test for forest. Since LiveTrie only pass through to the forest, I think it would be enough just to test that a call is routed to the forest, perhaps via a mock.

go/database/mpt/state.go
```diff
@@ -297,7 +300,9 @@ func (s *MptState) GetStorage(address common.Address, key common.Key) (value com
 func (s *MptState) SetStorage(address common.Address, key common.Key, value common.Value) error {
 	return s.trie.SetValue(address, key, value)
 }
-
+func (s *MptState) HasEmptyStorage(address common.Address) (bool, error) {
```
kjezek, line 303: please ad doc + add a simple test checking that the call pass through,

kjezek (changes requested):
Thanks, I think it looks good.  Just please extend a test bit. It is useful to test on a bit bigger MPT, not just one account and one storage. I helps avoid situations like a wrong path is traversed or a wrong account is reached

go/database/mpt/forest.go
```diff
@@ -383,6 +383,18 @@ func (s *Forest) SetValue(rootRef *NodeReference, addr common.Address, key commo
 	return newRoot, err
 }
 
+func (s *Forest) HasEmptyStorage(rootRef *NodeReference, addr common.Address) (isEmpty bool, err error) {
```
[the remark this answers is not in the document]
petr-hanzl: The documentation for forest is in the `Database` interface.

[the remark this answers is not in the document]
kjezek: ok, no need to repeat here then

go/database/mpt/state.go
petr-hanzl: I combined `live_trie` and `state_test` into one. Since it only checks for call this covers both. Wdyt?

go/database/mpt/forest_test.go
```diff
@@ -2155,6 +2155,120 @@ func TestForest_VisitPathToStorage(t *testing.T) {
...
+							continue
+						}
+						root, err = forest.SetValue(&root, address, common.Key{byte(i)}, common.Value{byte(i)})
+						if err != nil {
+							t.Fatalf("cannot update storage: %v", err)
+						}
+					}
+
+					// Account with empty storage
+					isEmpty, err := forest.HasEmptyStorage(&root, addresses[0])
```
kjezek, line 2219: I think it would be much stronger to test all accounts you created for them being empty or not. This cherrypicking of a few accounts is a bit awkward. 

You can repeat the loop and check that half is empty, half is not. During this loop you can empty more accounts using various strategies 
1/4 deleted by ClearStorage, 1/4 deleted by SetSlot, 1/4 deleted by SetAccountInfo, 1/4 remains non-empty
and after that re-check that expected accounts become empty.

go/database/mpt/state.go
kjezek: LiveTrie should have its test as well - a very simple one, checking the the method call passed through

go/database/mpt/forest_test.go
```diff
@@ -2155,6 +2155,147 @@ func TestForest_VisitPathToStorage(t *testing.T) {
...
+						if i < 50 {
+							continue
+						}
+						root, err = forest.SetValue(&root, address, common.Key{byte(i)}, common.Value{byte(i)})
+						if err != nil {
+							t.Fatalf("cannot update storage: %v", err)
+						}
+					}
+
+					for i := 0; i < N; i++ {
```
kjezek, line 2219: this loop has a few issues:
* it seems that all conditions starting with `if i >= 60 && i < 70 {` and below are unreachable
* constants 50, 60, 70, ... should be computed from `N`
* I understand that each if condition branch deletes the account by a different means, but why is the call to  `HasEmptyStorage` + consequent assertions repeated every-time? The sequence is always the same (except a different log message)

I think it would be more readable just to divide the list addresses into four sub-slices and for-each on them independently.  
```

for i, address := range addresses[0:N/4] {
 // one way of deleting a value 
}

```
and then the last for-each would check in one iteration that all addresses are empty or not

go/database/mpt/forest_test.go
```diff
@@ -2155,6 +2155,147 @@ func TestForest_VisitPathToStorage(t *testing.T) {
...
+								t.Fatalf("failed to ask whether account has empty storage: %v", err)
+							}
+							if !isEmpty {
+								t.Error("deleted account has empty storage forest returned false")
+							}
+							continue
+						}
+
+						// Set empty storage for the rest
+						root, err = forest.SetValue(&root, addresses[i], common.Key{byte(i)}, common.Value{})
```
kjezek, line 2280: I do not understand why all addresses that did not pass previous conditions are deleted here anyway. If the all possible options were already checked? What is being checked here?

go/database/mpt/live_trie_test.go
```diff
@@ -893,6 +893,35 @@ func TestLiveTrie_VerificationOfLiveTrieWithCorruptedFileFails(t *testing.T) {
...
+
+	for _, config := range allMptConfigs {
+		t.Run(config.Name, func(t *testing.T) {
+			dir := t.TempDir()
+
+			addr := common.Address{0x1}
+			ctrl := gomock.NewController(t)
+			db := NewMockDatabase(ctrl)
+			db.EXPECT().HasEmptyStorage(gomock.Any(), addr).Return(true, nil)
```
kjezek, line 905: I think it is not needed to set actual return value.

go/database/mpt/live_trie_test.go
```diff
@@ -893,6 +893,35 @@ func TestLiveTrie_VerificationOfLiveTrieWithCorruptedFileFails(t *testing.T) {
...
+
+			mpt, err := OpenFileLiveTrie(dir, config, 1024)
+			if err != nil {
+				t.Fatalf("failed to open live trie: %v", err)
+			}
+			mpt.forest = db
+
+			isEmpty, err := mpt.HasEmptyStorage(addr)
+			if err != nil {
```
kjezek, line 914: I think it is not needed to check actual return, it is mocked anyway

go/database/mpt/state_test.go
```diff
@@ -341,6 +341,50 @@ func TestState_StateModifications_Failing(t *testing.T) {
...
+				state, err := open(dir)
+				if err != nil {
+					t.Fatalf("cannot open state: %s", err)
+				}
+
+				addr := common.Address{0x1}
+				injectedErr := errors.New("injected error")
+				ctrl := gomock.NewController(t)
+				db := NewMockDatabase(ctrl)
+				db.EXPECT().HasEmptyStorage(gomock.Any(), addr).Return(true, nil)
```
kjezek, line 359: I think no need to test all possible outputs as long as it still test the same controlflow.

go/database/mpt/state_test.go
```diff
@@ -341,6 +341,50 @@ func TestState_StateModifications_Failing(t *testing.T) {
...
+				db.EXPECT().HasEmptyStorage(gomock.Any(), addr).Return(false, nil)
+				db.EXPECT().HasEmptyStorage(gomock.Any(), addr).Return(false, injectedErr)
+
+				state.trie.forest = db
+
+				isEmpty, err := state.HasEmptyStorage(addr)
+				if err != nil {
```
kjezek, line 366: no need to test returned mocked values

kjezek (changes requested):
I put a few remarks to the test. The most important is that it seems half of the loop in the forest test is never executed.

go/database/mpt/forest_test.go
petr-hanzl: First condition checks that the storage has actually been created.
Second condition calls `ClearStorage` storage for addreses between index 60 and 70 and checks that `HasEmptyStorage` returns true
Third condition deletes the account etween index 70 and 80 and checks that `HasEmptyStorage` returns true.
The last code snipet takes the rest of the indexes hence <=80  and calls `SetValue` with an empty value to remove the storage and checks that `HasEmptyStorage` returns true.

go/database/mpt/forest_test.go
petr-hanzl: There was a lost `continue` in the code, so the code was unreachable. Thanks for noticing.
Anyhow, I dont think it changes the code that much to compute the constants from `N`. But if you feel like it suits the testing pattern used in the project, I will redo it accordingly.

petr-hanzl: > * I understand that each if condition branch deletes the account by a different means, but why is the call to  `HasEmptyStorage` + consequent assertions repeated every-time? The sequence is always the same (except a different log message)

I think I dont understand this comment. I tried to implement your thought in [this](https://github.com/Fantom-foundation/Carmen/pull/906#discussion_r1636628859) comment.

kjezek: I wanted to say that this block: https://github.com/Fantom-foundation/Carmen/pull/906/files#diff-ebf52a24a2853cb8ec5fd9f7a9be48535f4ff59fc670f0121e51f9c26fec3b89R2190-R2196

is repeated many times.  You iterate over sequences of addresses, but check still the same - i.e. the repeated lines of code.  It would be better to iterate the addresses in one loop and do the test.

kjezek: about the `N` constant, the problem is that it was set to 100 and the conditions rely on this value, but if `N` is changed to 10, all will fall apart.  I.e. it is not much about simplicity, but functionality.

kjezek (approved):
looks good, thanks

PR #907, opened 2024-06-11
implements is-valid for witness proof

kjezek:
This PR implements IsValid function for the witness proof.  The proof is invalid if
* hash does not match hashed RLP representation
* if RLP cannot be decoded to a node

### DIFF
--- a/go/database/mpt/proof_test.go
+++ b/go/database/mpt/proof_test.go
@@ -15,6 +15,7 @@ import (
 	"github.com/Fantom-foundation/Carmen/go/common"
 	"github.com/Fantom-foundation/Carmen/go/database/mpt/shared"
 	"go.uber.org/mock/gomock"
+	"golang.org/x/exp/maps"
 	"testing"
 )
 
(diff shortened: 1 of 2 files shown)


go/database/mpt/proof_test.go
```diff
@@ -419,6 +420,57 @@ func TestWitnessProof_String(t *testing.T) {
...
+	desc := &Extension{
+		path: AddressToNibblePath(address, ctxt)[0:30],
+		next: &Account{address: address, pathLength: 10, info: AccountInfo{Nonce: common.Nonce{0x01}, Balance: common.Balance{0x02}, CodeHash: common.Hash{0x03}}},
+	}
+
+	root, node := ctxt.Build(desc)
+	totalProof := createReferenceProof(t, ctxt, &root, node)
+	handle := node.GetReadHandle()
+	defer handle.Release()
+	rootHash, dirty := handle.Get().GetHash()
```
HerbertJordan, line 439: nit: reading the hash technically requires View access, but here it does not matter.

go/database/mpt/proof_test.go
```diff
@@ -419,6 +420,57 @@ func TestWitnessProof_String(t *testing.T) {
...
+		proof := WitnessProof{maps.Clone(totalProof.proofDb)}
+		if !proof.IsValid() {
+			t.Fatalf("proof should be valid")
+		}
+	})
+
+	t.Run("hash mismatch", func(t *testing.T) {
+		proof := WitnessProof{maps.Clone(totalProof.proofDb)}
+		for k := range proof.proofDb {
+			if k != rootHash {
```
HerbertJordan, line 454: Why the special case for the root hash?

HerbertJordan (dismissed):
Looks good, thanks!

go/database/mpt/proof_test.go
kjezek: ah, sorry for repeating the same mistake. Anyway, this code was removed, since the rootHash is not needed anymore

go/database/mpt/proof_test.go
kjezek: right, it brings nothing, removed

HerbertJordan (approved):
Nice!

PR #910, opened 2024-06-12
Allow interrupting verification

petr-hanzl:
This PR allows interrupting verification and gracefully closing the database.

### DIFF
--- a/go/database/mpt/verification_test.go
+++ b/go/database/mpt/verification_test.go
@@ -11,6 +11,7 @@
 package mpt
 
 import (
+	"context"
 	"encoding/hex"
 	"errors"
 	"fmt"
@@ -18,11 +19,14 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"syscall"
 	"testing"
+	"time"
 
 	"github.com/Fantom-foundation/Carmen/go/backend/stock"
 	"github.com/Fantom-foundation/Carmen/go/backend/stock/file"
 	"github.com/Fantom-foundation/Carmen/go/common"
+	"github.com/Fantom-foundation/Carmen/go/common/interrupt"
 	"go.uber.org/mock/gomock"
 )
 
@@ -50,7 +54,7 @@ var forestFiles = []string{
 
 func TestVerification_VerifyValidForest(t *testing.T) {
 	runVerificationTest(t, func(t *testing.T, dir string, config MptConfig, roots []Root) {
-		if err := verifyFileForest(dir, config, roots, NilVerificationObserver{}); err != nil {
+		if err := verifyFileForest(context.Background(), dir, config, roots, NilVerificationObserver{}); err != nil {
 			t.Errorf("found unexpected error in fresh forest: %v", err)
 		}
 	})
@@ -68,7 +72,7 @@ func TestVerification_VerificationObserverIsKeptUpdatedOnEvents(t *testing.T) {
 			observer.EXPECT().EndVerification(nil),
 		)
 
-		if err := verifyFileForest(dir, config, roots, observer); err != nil {
+		if err := verifyFileForest(context.Background(), dir, config, roots, observer); err != nil {
 			t.Errorf("found unexpected error in fresh forest: %v", err)
 		}
 	})
@@ -81,7 +85,7 @@ func TestVerification_MissingFileIsDetected(t *testing.T) {
 				if err := os.RemoveAll(dir + "/" + file); err != nil {
 					t.Fatalf("failed to delete file %v: %v", file, err)
 				}
-				if err := verifyFileForest(dir, config, roots, NilVerificationObserver{}); err == nil {
+				if err := verifyFileForest(context.Background(), dir, config, roots, NilVerificationObserver{}); err == nil {
 					t.Errorf("The missing file %v should have been detected", file)
 				}
 			})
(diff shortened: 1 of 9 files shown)


kjezek (changes requested):
Thanks for this initiative. Did you try it on a big dataset? It looks to me that interruption checks are done way too coarse grained at the moment.  In particular, it is done only after each node type verification is finished.  I think for instance that verification on branches takes a long time. It seems the user does not have to be patient enough to wait. 

I think the right place to check for interrupt is in the `verifyHashes` and possible `forAllInnerNodes` methods.  Checking there would also simplify the code because when a verification block ends with an error, the next block does not continue. 

But when moving the check, also make sure that reading if the context was interrupted does not have a high overhead - to avoid situation which would prolong verification if the interruption is checked after checking each hash.

petr-hanzl:
> I think the right place to check for interrupt is in the `verifyHashes` and possible `forAllInnerNodes` methods. Checking there would also simplify the code because when a verification block ends with an error, the next block does not continue.

I thought about this place in the code but I felt like thats too often. I will test both codes with somewhat large database.

petr-hanzl:
> I think the right place to check for interrupt is in the `verifyHashes` and possible `forAllInnerNodes` methods. Checking there would also simplify the code because when a verification block ends with an error, the next block does not continue.

After testing `forAllInnerNodes`. I was a bit sceptical it would affect the performance quite significantly, but fortunately its not the case.

kjezek:
> > I think the right place to check for interrupt is in the `verifyHashes` and possible `forAllInnerNodes` methods. Checking there would also simplify the code because when a verification block ends with an error, the next block does not continue.
> 
> After testing `forAllInnerNodes`. I was a bit sceptical it would affect the performance quite significantly, but fortunately its not the case.

can you give some numbers? Size of the db, speed of verification before and after?

petr-hanzl:
`aida-vm-sdb` from 0 to 10m with current change took `296 minutes`. Without this change it took `281 minutes`. So the slowdown is ~6%

kjezek:
> `aida-vm-sdb` from 0 to 10m with current change took `296 minutes`. Without this change it took `281 minutes`. So the slowdown is ~6%

thanks for providing the numbers. It is a quite a bit as I expected. Can you check for the channel being closed only once in a while? It can be done via a simple counter.

go/database/mpt/verification_test.go
```diff
@@ -668,20 +672,77 @@ func TestVerification_VerificationObserverReportsError(t *testing.T) {
...
 
+func TestVerification_CanInterrupt(t *testing.T) {
+	type verifyFunc func(context.Context, string, MptConfig, []Root, VerificationObserver) error
+	tests := map[string]verifyFunc{"VerifyMptState": VerifyMptState, "VerifyFileForest": verifyFileForest}
+
+	for name, verify := range tests {
+		runVerificationTest(t, func(t *testing.T, dir string, config MptConfig, roots []Root) {
+			// First count the occurrence
+			counter := &interruptObserver{t: t}
+			if err := verifyFileForest(context.Background(), dir, config, roots, counter); err != nil {
```
kjezek, line 697: I think this should call `verify(...)`, right?

go/database/mpt/verification_test.go
```diff
@@ -668,20 +672,77 @@ func TestVerification_VerificationObserverReportsError(t *testing.T) {
...
+	}
+}
+
+type interruptObserver struct {
+	t               *testing.T
+	signalInterrupt bool
+	count           int
+}
+
+func (c *interruptObserver) StartVerification() {
```
kjezek, line 731: I do not very like how the interrupt is triggered here. First I think that syscall is asynchronous, and thus this test may be fragile. Second and most importantly, interrupt is triggered based on the number of execution of `StartVerification`. It is not obvious what is the relation between the number of execution of this method and the number of call to the `IsCancelled` method. 

I think much better is to really track the number of execution to `IsCancelled` first, and then simulate a cancelled context in next loops. For idea how to do it, please have a look here:

https://github.com/Fantom-foundation/Carmen/pull/947/files#diff-8bb57076b2dc35d3c0b28ecd9488c8c002f0408e571713e8462c7b7e8a6d06aeR640

kjezek (changes requested):
Thanks for this change. I think there are two problems in the test:
1. minor - I guess `verify(...) `should be called instead of` verifyFileForset `
2. the way the number of execution to `IsCancelled` is determined  is very counterintuitive.  It is determined by the number of executions to `StartVerification` of the verification observer.  If these methods are at the moment called in pair, I guess it is just a coincidence and not a natural connection which would align using one to determine the number of execution to the other one.

If you want to count the number of execution of IsCancelled, then count this method. A way how this can be done, can be seen for instance here: https://github.com/Fantom-foundation/Carmen/pull/947/files#diff-8bb57076b2dc35d3c0b28ecd9488c8c002f0408e571713e8462c7b7e8a6d06aeR640

go/database/mpt/verification_test.go
kjezek: somehow the link does not lead to the line number I wanted to link. Please search for this line

`type countingWhenDoneContext struct {`

go/database/mpt/verification_test.go
```diff
@@ -668,20 +670,77 @@ func TestVerification_VerificationObserverReportsError(t *testing.T) {
...
+			}
+
+			for i := 0; i < ctx.count; i++ {
+				ctx := newCountingWhenDoneContext(context.Background(), i)
+				if err := verify(ctx, dir, config, roots, nil); !errors.Is(err, interrupt.ErrCanceled) {
+					t.Errorf("%v: found unexpected error: %v", name, err)
+				}
+			}
+
+			// Make sure data is not corrupted
```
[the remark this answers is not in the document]
kjezek: I think it is too strong  to say that this code makes sure that data is not corrupted. If I understand correctly it only checks the database can be opened - i.e. the lock and dirty flag was released

kjezek (dismissed):
looks good, thanks

kjezek (dismissed):

kjezek (approved):

PR #911, opened 2024-06-12
propagates all terminal node except empty one via path iterators

kjezek:
This PR updates methods to visit path in the MPT in the way that terminal nodes are visited even when their path in the MPT was not matched. In this case such nodes are terminal and are last returned. 

Exception are Empty node and a Branch node when the path was exhausted - these nodes are not returned.

### DIFF
--- a/go/database/mpt/nodes.go
+++ b/go/database/mpt/nodes.go
@@ -306,9 +306,11 @@ func visitPathTo(source NodeSource, root *NodeReference, path []Nibble, address 
 				done = true
 			}
 		case *BranchNode:
+			if len(path) == 0 {
+				done = true
+			}
 			nodeId = &n.children[path[0]]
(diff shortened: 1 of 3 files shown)


go/database/mpt/nodes.go
```diff
@@ -301,6 +301,7 @@ func visitPathTo(source NodeSource, root *NodeReference, path []Nibble, address
 			if n.path.IsPrefixOf(path) {
 				nodeId = &n.next
 				path = path[n.path.Length():]
+				done = len(path) == 0
```
HerbertJordan, line 304: This should actually not be possible in a consistent tree. Also, for BranchNodes, the path should never be fully consumed. However, I guess it does not harm to have this code here.

HerbertJordan (approved):

PR #908, opened 2024-06-11
Integrate `amount.Amount` into `State` interface

Mike-CZ:
This PR replaces `common.Balance` with new `amount.Amount` type in `State` interface. This PR should be merged after https://github.com/Fantom-foundation/Carmen/pull/898.

PR #920, opened 2024-06-17
Use `amount.Amount` type in `LiveDB` interface

Mike-CZ:
This PR uses `amount.Amount` type in `LiveDB` interface. Should be merged after [this](https://github.com/Fantom-foundation/Carmen/pull/908).


go/database/mpt/state_test.go
```diff
@@ -403,7 +404,7 @@ func TestState_StateModificationsWithoutErrorHaveExpectedEffects(t *testing.T) {
 				t.Errorf("account should not exist: %v err: %s", exists, err)
 			}
 
-			var emptyBalance common.Balance
+			emptyBalance := amount.New()
```
kjezek, line 407: perhaps we should have also an alias `amont.Zero()`

kjezek (approved):
looks good to me, thanks

HerbertJordan (approved):
Looks good, thanks!

go/database/mpt/state_test.go
Mike-CZ: Once it's all merged, I'll create a new PR with `Zero` factory method and replaced all `New()` instances.

kjezek (approved):

PR #921, opened 2024-06-17
Use `amount.Amount` type in `Archive` interface

Mike-CZ:
This PR uses `amount.Amount` type in `Archive` interface. Should be merged after https://github.com/Fantom-foundation/Carmen/pull/920.


HerbertJordan (dismissed):
Looks good, thanks!

go/backend/archive/ldb/archive.go
```diff
@@ -239,18 +242,18 @@ func (a *Archive) Exists(block uint64, account common.Address) (exists bool, err
...
 	it := a.db.NewIterator(&keyRange, nil)
 	defer it.Release()
 
 	if it.Next() {
-		copy(balance[:], it.Value())
+		balance = amount.NewFromBytes(it.Value()[:]...)
 		return balance, nil
 	}
-	return common.Balance{}, it.Error()
+	return amount.New(), it.Error()
```
kjezek, line 256: should not we have `amount.Zero()` ?

kjezek (approved):
thanks, it looks good

go/backend/archive/ldb/archive.go
Mike-CZ: Once it's all merged, I'll create a new PR with `Zero` factory method and all `New()` instances replaced.

kjezek (approved):

kjezek (approved):

PR #922, opened 2024-06-17
integrates witness proof into public facade

kjezek:
This PR extends the public API of the  `WitnessProof` support. 

* This defines a new package `witness` with a simple `Proof` interface. 
* This interface is implemented by `mpt.WitnessProof`, i.e. the MPT support of the witness proof. 
* The `carmen.HistoricBlockContext` is extended of a new  features to extract the proof, i.e. a new method `GetProof`
* `State` exposes a method  `CreateWitnessProof` called from `carmen.HistoricBlockContext` and it provides `witness.Proof`, all schemes except S5 do `panic` when calling this method, only the `ArchiveTrie` supports it.  

An example is provided showing how to use.

### DIFF
--- a/go/backend/archive/archive.go
+++ b/go/backend/archive/archive.go
@@ -14,7 +14,11 @@ package archive
 
 import (
 	"github.com/Fantom-foundation/Carmen/go/common"
+	"github.com/Fantom-foundation/Carmen/go/common/witness"
 )
+
+// ErrWitnessProofNotSupported is returned when the archive does not support witness proofs.
+const ErrWitnessProofNotSupported = common.ConstError("witness proof not supported")
 
 // An Archive retains a history of state mutations in a blockchain on a
 // block-level granularity. The history is recorded by adding per-block updates.
@@ -51,6 +55,9 @@ type Archive interface {
 
 	// GetHash provides a hash of the state at the given block height.
 	GetHash(block uint64) (hash common.Hash, err error)
+
+	// CreateWitnessProof creates a witness proof for the given account and keys.
+	CreateWitnessProof(block uint64, address common.Address, keys ...common.Key) (witness.Proof, error)
 
 	// MemoryFootprintProvider provides the size of the store in memory in bytes.
 	common.MemoryFootprintProvider
--- a/go/backend/archive/archive_test.go
+++ b/go/backend/archive/archive_test.go
@@ -12,6 +12,7 @@ package archive_test
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"io"
 	"testing"
@@ -256,6 +257,68 @@ func TestAccountDeleteCreate(t *testing.T) {
 				if hash, err := a.GetHash(9); err != nil || fmt.Sprintf("%x", hash) != "49a86f0a7e53b6d7411edad398e2f807ffbe79d179eebbd42117425343b38fa9" {
 					t.Errorf("unexpected hash of block 9: %x; %v", hash, err)
 				}
+			}
+		})
+	}
+}
+
+func TestArchive_CreateWitnessProof(t *testing.T) {
+	for _, factory := range getArchiveFactories(t) {
+		t.Run(factory.label, func(t *testing.T) {
+			a := factory.getArchive(t.TempDir())
+			defer func() {
+				if err := a.Close(); err != nil {
+					t.Fatalf("failed to close archive; %s", err)
+				}
+			}()
+
+			if err := a.Add(1, common.Update{
+				CreatedAccounts: []common.Address{{1}},
+				Balances: []common.BalanceUpdate{
+					{Account: common.Address{1}, Balance: common.Balance{0x12}},
+				},
+				Slots: []common.SlotUpdate{
+					{Account: common.Address{1}, Key: common.Key{2}, Value: common.Value{3}},
+				},
+			}, nil); err != nil {
+				t.Fatalf("failed to add block: %v", err)
+			}
+
+			proof, err := a.CreateWitnessProof(1, common.Address{1}, common.Key{2})
+			if err != nil {
+				if errors.Is(err, archive.ErrWitnessProofNotSupported) {
+					t.Skip(err)
+				}
+				t.Fatalf("failed to create witness proof; %s", err)
+			}
+
+			if !proof.IsValid() {
+				t.Errorf("invalid proof")
+			}
+
+			hash, err := a.GetHash(1)
+			if err != nil {
+				t.Fatalf("failed to get hash; %s", err)
+			}
+			balance, complete, err := proof.GetBalance(hash, common.Address{1})
+			if err != nil {
+				t.Fatalf("failed to get balance; %s", err)
+			}
+			if !complete {
+				t.Errorf("balance proof is incomplete")
+			}
+			if got, want := balance, (common.Balance{0x12}); got != want {
+				t.Errorf("unexpected balance; got: %x, want: %x", got, want)
+			}
+			value, complete, err := proof.GetState(hash, common.Address{1}, common.Key{2})
+			if err != nil {
+				t.Fatalf("failed to get state; %s", err)
+			}
+			if !complete {
+				t.Errorf("state proof is incomplete")
+			}
+			if got, want := value, (common.Value{3}); got != want {
+				t.Errorf("unexpected value; got: %x, want: %x", got, want)
 			}
 		})
 	}
--- a/go/carmen/block.go
+++ b/go/carmen/block.go
@@ -13,6 +13,7 @@ package carmen
 import (
 	"errors"
 	"fmt"
+	"github.com/Fantom-foundation/Carmen/go/common"
 	"sync"
 
 	"github.com/Fantom-foundation/Carmen/go/state"
@@ -111,7 +112,8 @@ func (c *headBlockContext) end() error {
 
 type archiveBlockContext struct {
 	commonContext
-	state state.NonCommittableStateDB
+	state        state.NonCommittableStateDB
+	archiveState state.State
 }
 
 func (c *archiveBlockContext) BeginTransaction() (TransactionContext, error) {
@@ -135,6 +137,27 @@ func (c *archiveBlockContext) BeginTransaction() (TransactionContext, error) {
 
 func (c *archiveBlockContext) RunTransaction(run func(TransactionContext) error) error {
 	return runTransaction(c, run)
+}
+
+func (c *archiveBlockContext) GetProof(address Address, keys ...Key) (WitnessProof, error) {
+	c.lock.Lock()
+	if c.db == nil {
+		c.lock.Unlock()
+		return nil, fmt.Errorf("cannot get proof in invalid block context")
+	}
+	c.lock.Unlock()
+
+	commonKeys := make([]common.Key, len(keys))
+	for i, k := range keys {
+		commonKeys[i] = common.Key(k)
+	}
+
+	proof, err := c.archiveState.CreateWitnessProof(common.Address(address), commonKeys...)
+	if err != nil {
+		return nil, err
+	}
+
+	return witnessProof{proof}, nil
 }
 
 func (c *archiveBlockContext) Close() error {
--- a/go/carmen/carmen.go
+++ b/go/carmen/carmen.go
@@ -12,6 +12,7 @@ package carmen
 
 import (
 	"github.com/Fantom-foundation/Carmen/go/common"
+	"github.com/Fantom-foundation/Carmen/go/common/tribool"
 	"github.com/Fantom-foundation/Carmen/go/state"
 )
 
@@ -187,6 +188,11 @@ type HeadBlockContext interface {
 // to the blockchain happens.
 type HistoricBlockContext interface {
 	blockContext
+
+	// GetProof creates a witness proof for the given account and keys.
+	// Error may be produced when it occurs in the underlying database;
+	// otherwise, the proof is returned.
+	GetProof(address Address, keys ...Key) (WitnessProof, error)
 
 	// Close releases resources held by this context. All modifications made
 	// within this context are discarded. This context is invalid after this
@@ -453,6 +459,8 @@ type Value common.Value
 
 // Hash is a 32byte hash.
 type Hash common.Hash
+
+type Tribool tribool.Tribool
 
 // Log summarizes a log message recorded during the execution of a contract.
 type Log struct {
--- a/go/carmen/database_test.go
+++ b/go/carmen/database_test.go
@@ -845,6 +845,374 @@ func TestDatabase_CloseDB_Unfinished_Queries(t *testing.T) {
 	}
 }
 
+func TestDatabase_GetProof(t *testing.T) {
+	db, err := openTestDatabase(t)
+	if err != nil {
+		t.Fatalf("failed to open database: %v", err)
+	}
+
+	codeHashes := make(map[Address]Hash)
+
+	// init data
+	const numBlocks = 100
+	const numAccounts = 11
+	const numKeys = 12
+	for i := 0; i < numBlocks; i++ {
+		if err := db.AddBlock(uint64(i), func(context HeadBlockContext) error {
+			for j := 0; j < numAccounts; j++ {
+				if err := context.RunTransaction(func(tc TransactionContext) error {
+					addr := Address{byte(j)}
+					tc.CreateAccount(addr)
+					tc.SetCode(addr, []byte{byte(j)})
+					codeHashes[addr] = Hash(common.Keccak256([]byte{byte(j)}))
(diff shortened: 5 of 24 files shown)


go/carmen/carmen.go
```diff
@@ -206,6 +217,7 @@ type HistoricBlockContext interface {
 // context before another one can be opened.
 // Methods of this interface are not thread safe.
 type TransactionContext interface {
+	QueryContext
```
petr-hanzl, line 220: I believe `TransactionContext` implicitly implements `QueryContext` as all method within `QrCtx` are contained within `TxCtx` hence this is not necessary

go/carmen/example_test.go
```diff
@@ -282,3 +282,134 @@ func ExampleDatabase_GetHistoricContext() {
...
+
+	// serialise to a format to be passed to others
+	stream := carmen.MptWitnessProofToNodes(completeProof)
+	// recover at the other side
+	recoveredProof := carmen.MptWitnessProofFromNodes(stream)
+
+	// ------- Values can be proven outside this client  -------
+
+	// Prove the values for each address and key offline without access to database
+	rootHashes := make([]carmen.Hash, N) // TODO get real root hashes
```
petr-hanzl, line 348: nit: todo?

go/carmen/example_test.go
```diff
@@ -282,3 +282,134 @@ func ExampleDatabase_GetHistoricContext() {
...
+
+	if err := db.Close(); err != nil {
+		log.Fatalf("cannot close db: %v", err)
+	}
+
+	if err := os.RemoveAll(dir); err != nil {
+		log.Fatalf("cannot remove dir: %v", err)
+	}
+
+	// Output: TODO
```
petr-hanzl, line 414: nit: todo?

go/database/mpt/proof.go
```diff
@@ -297,3 +298,121 @@ func createPathFromKeyPrefix(key common.Key, nibbles uint8) Path {
...
+
+// AllAddressesEmpty checks that all accounts are empty for the input root hash and the address range.
+// If the witness proof contains all empty accounts for the input address range, it returns true.
+// An empty account is an account that contains a zero balance, nonce, and code hash.
+// If there is at least one non-empty account, it returns false. If the proof is not complete,
+// it returns unknown. An incomplete proof is a proof where the input address terminates in a node
+// that is not a correct account node.
+func (c *mptWitnessProofState) AllAddressesEmpty(from, to common.Address) (tribool.Tribool, error) {
+	return c.proof.AllAddressesEmpty(c.root, from, to)
+}
```
petr-hanzl, line 418: I dont think the documentation here is needed. It seems like a copy of the documentation of the implemented interface `witness.Proof`.

petr-hanzl (commented):
Looks good. Consider removing the documentation of `mptWitnessProofState`.

go/carmen/carmen.go
```diff
@@ -188,6 +189,16 @@ type HeadBlockContext interface {
 type HistoricBlockContext interface {
 	blockContext
 
+	// CreateWitnessProof creates a witness proof for the given account and keys.
+	// Error may be produced when it occurs in the underlying database,
+	// otherwise, the proof is returned.
+	CreateWitnessProof(address Address, keys ...Key) (WitnessProof, error)
```
HerbertJordan, line 195: How about renaming this to just `GetProof`?

go/common/witness/proof.go
```diff
@@ -0,0 +1,66 @@
...
+	IsValid() bool
+
+	// Add adds a proof node to the witness proof.
+	Add(Proof)
+}
+
+// ProofState allows for getting properties of accounts and account storage slots.
+// It contains methods to determine values of properties based on addresses and keys,
+// and returns the values and information if the proof was sufficient to determine the value.
+type ProofState interface {
```
HerbertJordan, line 22: Why the differentiation between a `Proof` and a `ProofState`. Could this not be the same?

go/database/mpt/proof.go
```diff
@@ -297,3 +298,121 @@ func createPathFromKeyPrefix(key common.Key, nibbles uint8) Path {
...
 }
+
+// NewWitnessProof creates a new witness proof.
+func NewWitnessProof() witness.Proof {
+	return witnessProofAdapter{WitnessProof{proofDb{}}}
+}
+
+// WitnessProofToNodes converts the witness proof to a list of strings.
+// Each string is an RLP node of the witness proof.
+func WitnessProofToNodes(proof witness.Proof) []string {
```
HerbertJordan, line 309: I think this should be a member function of the proof instead of a stand-alone function.
It might also be part of the proof interface.

go/database/mpt/proof.go
```diff
@@ -297,3 +298,121 @@ func createPathFromKeyPrefix(key common.Key, nibbles uint8) Path {
...
+	p := proof.(witnessProofAdapter).WitnessProof
```
HerbertJordan, line 310: The way it is now, it really only works for this one type of proof implementation and crash for all others.

HerbertJordan (commented):
Before reviewing it in depth, I would like to discuss the idea behind the differentiation between a `Proof` and a `ProofState`. What is the intention behind this? Wouldn't it be simpler to merge those two interfaces into a single `Proof` interface?

go/common/witness/proof.go
kjezek: I was at first sticked to the idea that you create a proof first --> you get `Proof`, and "later" you query the proof (for a root hash), which gives you the `ProofState` -> you query this. 

I think it sounds plausible to merge these two together, perhaps I will try to implement and see how would this look like

go/database/mpt/proof.go
kjezek: it would loos genericity, the idea was that `witness.Proof` serves for all schemes while the nodes are only in MPT.

I was also thinking about something more general, like a method `Serialise/Deserialise`, which would work with `[]byte` etc, it could then go to the interface.

go/database/mpt/proof.go
kjezek: right, it was meant to work this way.  If we can create a proof from other scheme, it will produce a `witness.Proof `as well, but if a user uses it to convert as an MPT witness proof, it crashes - it is considered a user mistake. He is suppose to use only a proof instance created from respective scheme.

go/carmen/carmen.go
```diff
@@ -206,6 +212,7 @@ type HistoricBlockContext interface {
 // context before another one can be opened.
 // Methods of this interface are not thread safe.
 type TransactionContext interface {
+	QueryContext
```
HerbertJordan, line 215: This looks unrelated ... if so, please remove this line.

go/carmen/block.go
```diff
@@ -137,6 +139,27 @@ func (c *archiveBlockContext) RunTransaction(run func(TransactionContext) error)
 	return runTransaction(c, run)
 }
 
+func (c *archiveBlockContext) GetProof(address Address, keys ...Key) (WitnessProof, error) {
+	c.lock.Lock()
+	if c.db == nil {
+		c.lock.Unlock()
+		return WitnessProof{}, fmt.Errorf("cannot get proof in invalid block context")
+	}
+	c.lock.Unlock()
```
HerbertJordan, line 148: I think this lock should be held until the end since otherwise the database underlying the `archiveState` may get closed while running this function.

go/carmen/proof.go
```diff
@@ -0,0 +1,56 @@
...
+)
+
+// CreateWitnessProofFromNodes creates a witness proof from a list of strings.
+// Each string is an RLP node of the witness proof.
+func CreateWitnessProofFromNodes(elements ...string) WitnessProof {
+	proof := mpt.CreateWitnessProofFromNodes(elements)
+	return WitnessProof{proof}
+}
+
+type WitnessProof struct {
```
HerbertJordan, line 16: To enable future refactoring hidden from external API users, we should hide this behind an interface.

go/database/mpt/proof.go
```diff
@@ -201,6 +213,14 @@ func (p WitnessProof) String() string {
 	return b.String()
 }
 
+func (p WitnessProof) GetElements() []string {
+	res := make([]string, 0, len(p.proofDb))
+	for _, v := range p.proofDb {
```
HerbertJordan, line 218: nit: for a future update, we might consider storing the elements inside the proof as strings instead of []byte.

go/database/mpt/proof.go
```diff
@@ -201,6 +213,14 @@ func (p WitnessProof) String() string {
 	return b.String()
 }
 
+func (p WitnessProof) GetElements() []string {
```
HerbertJordan, line 216: nit: document

go/database/mpt/proof.go
kjezek: what is the benefit of using strings instead?

go/database/mpt/proof.go
kjezek: done differently at the end

go/carmen/carmen.go
kjezek: done

go/carmen/carmen.go
kjezek: ok

go/carmen/block.go
kjezek: I guess we should prevent closing of database when a witness proof creation is running, similarly as we do for the head block state, because locking the whole method may be too coarse?

kjezek: Actually it seems it is already in-place:
https://github.com/Fantom-foundation/Carmen/blob/main/go/carmen/database.go#L207
and
https://github.com/Fantom-foundation/Carmen/blob/main/go/carmen/database.go#L264

go/carmen/proof.go
kjezek: ok, done

go/database/mpt/proof.go
kjezek: done

go/carmen/example_test.go
kjezek: finished, fixed

go/carmen/example_test.go
kjezek: finished, fixed

go/state/cppstate/cpp_state.go
```diff
@@ -279,6 +280,10 @@ func (cs *CppState) Check() error {
 	return nil
 }
 
+func (cs *CppState) CreateWitnessProof(address common.Address, keys ...common.Key) (witness.Proof, error) {
+	panic("not implemented")
+}
```
petr-hanzl, line 285: I think this 'not implemented' issues might get piled up - we had the same thing with `HasEmptyStorage` func and I believe we might be extending these interfaces in near future. 

How about we create a general error `ErrNotImplemented` and return it instead of panicking and we will have this prepared for future unimplemented functions within unused schemas?

go/state/gostate/go_state.go
```diff
@@ -424,3 +425,7 @@ func (s *GoState) GetSnapshotVerifier(metadata []byte) (backend.SnapshotVerifier
 	}
 	return backend.NewComposedSnapshotVerifier(verifiers, partCounts), nil
 }
+
+func (s *GoState) CreateWitnessProof(address common.Address, keys ...common.Key) (witness.Proof, error) {
+	panic("not implemented")
```
petr-hanzl, line 430: Same case as above...

petr-hanzl (dismissed):

go/state/cppstate/cpp_state.go
kjezek: We could do this, but I wonder what would be the difference. `panic` is for sure more severe, but these methods should not be ever called. When only the public API and supported configuration is called, there is no way to execute these methods. If the client bypasses this and calls something which is not supported, I do not see a lot of difference in which way he is informed about calling an unsupported feature.

petr-hanzl: ic. Yeah, you are right. There is actually no way this gets called by the public API. Lets keep the panic.

go/database/mpt/proof.go
HerbertJordan: You can skip the conversion step here and there, which is always a memory allocation and copy.

go/carmen/block.go
HerbertJordan: Alright, this may allow multiple proof-queries in parallel.

go/carmen/database_test.go
```diff
@@ -845,6 +845,374 @@ func TestDatabase_CloseDB_Unfinished_Queries(t *testing.T) {
...
+				}
+				if got, want := value, (Value{byte(i), byte(j), byte(k)}); got != want {
+					t.Errorf("unexpected value, wanted %v, got %v", want, got)
+				}
+			}
+		}
+	}
+}
+
+func TestDatabase_CloseDB_Unfinished_ProofE_CannotCloseDb(t *testing.T) {
```
HerbertJordan, line 1103: nit:
```suggestion
func TestDatabase_CloseDB_Unfinished_Proof_CannotCloseDb(t *testing.T) {
```

go/carmen/database_test.go
```diff
@@ -845,6 +845,374 @@ func TestDatabase_CloseDB_Unfinished_Queries(t *testing.T) {
...
+	go func() {
+		// attempts to close db must keep failing
+		if err := db.Close(); !errors.Is(err, errBlockContextRunning) {
+			t.Errorf("closing db should fail: %v", err)
+		}
+		done.Store(true)
+	}()
+
+	for !done.Load() {
+		_, err = block.GetProof(Address{1})
```
HerbertJordan, line 1142: This loop may never be executed if the go routine above is fast.

I don't think this test benefits from the concurrent execution of the close attempt and the reading of the proof. What would get lost if those steps would be done in sequence:
- attempt to close the db => it should fail
- read a proof => should work
- attempt to close the db => it should fail
- close the block context => should work
- attempt to close the db => should work

go/carmen/database_test.go
```diff
@@ -845,6 +845,374 @@ func TestDatabase_CloseDB_Unfinished_Queries(t *testing.T) {
...
+	block, err := db.GetHistoricContext(1)
+	if err != nil {
+		t.Fatalf("cannot get block: %v", err)
+	}
+
+	if err := block.Close(); err != nil {
+		t.Fatalf("cannot close block: %v", err)
+	}
+
+	if err := db.Close(); err != nil {
```
HerbertJordan, line 1189: I think the `db.Close()` should be moved after the attempt to fetch a proof since the closing of the block context needs to prevent further proof reading, not the closing of the DB.

HerbertJordan (approved):

go/database/mpt/archive_trie.go
```diff
@@ -241,6 +242,13 @@ func (a *ArchiveTrie) GetHash(block uint64) (hash common.Hash, err error) {
 	return res, nil
 }
 
+func (a *ArchiveTrie) CreateWitnessProof(block uint64, address common.Address, keys ...common.Key) (witness.Proof, error) {
+	a.rootsMutex.Lock()
```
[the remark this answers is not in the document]
kjezek: added

go/carmen/database_test.go
kjezek: ok, right, changed

go/carmen/database_test.go
kjezek: right, moved

go/backend/archive/archive_test.go
```diff
@@ -261,6 +262,72 @@ func TestAccountDeleteCreate(t *testing.T) {
...
+			defer func() {
+				if err := a.Close(); err != nil {
+					t.Fatalf("failed to close archive; %s", err)
+				}
+			}()
+
+			if err := a.Add(1, common.Update{
+				CreatedAccounts: []common.Address{{1}},
+				Balances: []common.BalanceUpdate{
+					{Account: common.Address{1}, Balance: common.Balance{0x12}},
```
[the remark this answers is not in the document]
HerbertJordan: Witness proofs have priority I'd say.

HerbertJordan (approved):

go/backend/archive/archive.go
```diff
@@ -14,8 +14,12 @@ package archive
 
 import (
 	"github.com/Fantom-foundation/Carmen/go/common"
+	"github.com/Fantom-foundation/Carmen/go/common/witness"
 )
 
+// ErrWitnessProfNotSupported is returned when the archive does not support witness proofs.
+const ErrWitnessProfNotSupported = common.ConstError("witness proof not supported")
```
HerbertJordan, line 21: I guess it should be `Proof` nit `Prof` ... although a `Prof` can also be a Witness :)
```suggestion
const ErrWitnessProofNotSupported = common.ConstError("witness proof not supported")
```

HerbertJordan (dismissed):

HerbertJordan (approved):

PR #924, opened 2024-06-18
HasEmptyStorage to the public API

petr-hanzl:
This PR exposes `HasEmptyStorage` to the public API under `go/carmen`


kjezek (changes requested):
Thanks for this PR, I would like to discuss following: 

Would not it make more sense to have the method `HasEmptyStorage(address Address)` in `HeadBlockContext`. It does not seem to be related to `TransactionContext`

So far the API was oriented to a) block processing, b) historic query. I think this new functionality stays apart, i.e. question where is the right place to put it. @HerbertJordan  what do you think?

Furthermore:
* I think returning `Tribool` from this method is not right. Shall we just return error?

petr-hanzl:
@kjezek 
> Would not it make more sense to have the method `HasEmptyStorage(address Address)` in `HeadBlockContext`. It does not seem to be related to `TransactionContext`
> 
> So far the API was oriented to a) block processing, b) historic query. I think this new functionality stays apart, i.e. question where is the right place to put it. @HerbertJordan what do you think?

Looking at what methods are inside `HeadBlockContext` it does not feel like the correct place.
Lets wait for Herberts opinion on that.

> Furthermore:
> 
> * I think returning `Tribool` from this method is not right. Shall we just return error?

I wanted to keep the pattern of the `QueryContext` and the `TransactionContext` of not returning an error so thats why I did `Tribool`. We never return an error for any `Getter`.

kjezek (changes requested):
I am not really sure this check should be used:
https://github.com/Fantom-foundation/Carmen/pull/924/files#diff-ce3e7c8603135541dce63d9fbc9e0772cde7bc2db731109808be2c3a85559a5fR839-R845

Looking at Ethereum, they verify only the state root hash, which is computed at the end of a block. 

It means that if there was a transaction that creates an account, modify its storage, and creates the same account again (dont know if it is possible), this would be entirely possible in Ethereum. 

Our implementation would diverge then, because it would prevent creation of the account. 

What would be the right semantics then? First account would be created, while the creation of the same account within the same transaction would blocked - i.e. would the account be still created? If yes, then it looks like it does not make any difference if the cache content is not checked, and we can remove it.

petr-hanzl:
@kjezek 
> I am not really sure this check should be used: https://github.com/Fantom-foundation/Carmen/pull/924/files#diff-ce3e7c8603135541dce63d9fbc9e0772cde7bc2db731109808be2c3a85559a5fR839-R845
>
> Looking at Ethereum, they verify only the state root hash, which is computed at the end of a block.

We had a discussion with @HerbertJordan and we agreed on this is correct. Lets bring him to the conversation. 

Could you provide a link to the `geth` implementation of this?

kjezek:
> @kjezek
> 
> > I am not really sure this check should be used: https://github.com/Fantom-foundation/Carmen/pull/924/files#diff-ce3e7c8603135541dce63d9fbc9e0772cde7bc2db731109808be2c3a85559a5fR839-R845
> > Looking at Ethereum, they verify only the state root hash, which is computed at the end of a block.
> 
> We had a discussion with @HerbertJordan and we agreed on this is correct. Lets bring him to the conversation.
> 
> Could you provide a link to the `geth` implementation of this?

the only place a variant of this new method is used in go-etheremum is: 
https://github.com/ethereum/go-ethereum/blob/master/core/vm/evm.go#L468

as you can see, they check only the storage root, which is known (non-empty) only for already sealed blocks. 

Also I did reading related to this new feature and it sounded to me as if it targeted only possibility of not to replace already existing accounts. 

I understand that semantics of this feature is a bit chaotic, but since we are introducing it only for compatibility, I would not be very creative and implement it the same way. 

Perhaps even a cleaner way of doing this change would be to check in the `CreateAccount` of `StateDB`, and do not expose any new method, but this would need to perfectly understand the semantics and implement it properly.

petr-hanzl:
> the only place a variant of this new method is used in go-etheremum is: https://github.com/ethereum/go-ethereum/blob/master/core/vm/evm.go#L468
> 
> as you can see, they check only the storage root, which is known (non-empty) only for already sealed blocks.
> 
> Also I did reading related to this new feature and it sounded to me as if it targeted only possibility of not to replace already existing accounts.
> 
> I understand that semantics of this feature is a bit chaotic, but since we are introducing it only for compatibility, I would not be very creative and implement it the same way.
> 
> Perhaps even a cleaner way of doing this change would be to check in the `CreateAccount` of `StateDB`, and do not expose any new method, but this would need to perfectly understand the semantics and implement it properly.
I believe the `EVM` needs this method exposed. Ie. Aida is currently waiting for this implementation.

In discussion with Herbert we found out we actually dont need to implement this method. The thing is we might want to run the Ethereum data as tests. Or the `eth-test` package.
From what I understand, this wont ever be called by new transactions.

kjezek:
> > the only place a variant of this new method is used in go-etheremum is: https://github.com/ethereum/go-ethereum/blob/master/core/vm/evm.go#L468
> > as you can see, they check only the storage root, which is known (non-empty) only for already sealed blocks.
> > Also I did reading related to this new feature and it sounded to me as if it targeted only possibility of not to replace already existing accounts.
> > I understand that semantics of this feature is a bit chaotic, but since we are introducing it only for compatibility, I would not be very creative and implement it the same way.
> > Perhaps even a cleaner way of doing this change would be to check in the `CreateAccount` of `StateDB`, and do not expose any new method, but this would need to perfectly understand the semantics and implement it properly.
> > I believe the `EVM` needs this method exposed. Ie. Aida is currently waiting for this implementation.
> 
> In discussion with Herbert we found out we actually dont need to implement this method. The thing is we might want to run the Ethereum data as tests. Or the `eth-test` package. From what I understand, this wont ever be called by new transactions.

I am confused now. Which method we do not need? The one you are implementing in this PR? I.e. this PR is not to be merged?

HerbertJordan:
My understanding is like this:

- before [EIP-158](https://eips.ethereum.org/EIPS/eip-158) introduced by the [Tangerine Whistle](https://ethereum.org/en/history/#tangerine-whistle) fork it was possible to create accounts that have a balance of 0, a nonce of 0, empty code, but some storage state
- with EIP-158 the rules changed, and every new account's initial nonce is now set to 1 instead of 0 (see [here](https://github.com/ethereum/go-ethereum/blob/00675c5876d84bffd44f3b18db9db7892fa3f87e/core/vm/evm.go#L490-L492))
- by the common definition of "empty accounts", which only checks the balance, nonce, and code, these accounts are considered empty
- new accounts can only be created for addresses where the associated account is empty; the problem is that accounts created before EIP-158 that have a balance of 0, a nonce of 0, and empty code would thus be considered empty, although they have state
- [EIP-7610](https://eips.ethereum.org/EIPS/eip-7610) identified this problem and reported 28 contracts on the Ethereum chain that have this property; to protect those, the rules for creating new accounts got changed: now the balance and nonce need to be 0, the code empty, and the storage empty. To identify empty storage, the storage hash is checked.

Thus, the fix introduced by  EIP-7610 is only relevant for contracts that have been introduced before EIP-158 went live.

In Fantom, this is never the case, since the chain was started using the Istanbul revision, including EIP-158. Thus, EIP-7610 could be ignored.

However, in order to pass general Ethereum tests suits, and potentially be able to process the Ethereum history, it was decided to provide support for this feature in Carmen. Since in Carmen it is simpler to determine whether storage is empty than to retrieve the storage hash, we went for the `HasEmptyStorage` function instead of the `GetStorageHash` feature.

As we are now offering a `HasEmptyStorage` function, we have the freedom to define its semantic to some degree, and we should consider it an independent feature and not just as a stand-in for `GetStorageHash` and the one use case it is used for in `geth`.

go/carmen/carmen.go
```diff
@@ -277,6 +277,9 @@ type TransactionContext interface {
 	// that gets reset after each transaction.
 	SetTransientState(Address, Key, Value)
 
+	// HasEmptyStorage returns true if account has empty storage.
+	HasEmptyStorage(address Address) bool
```
HerbertJordan, line 281: This function should be moved up to the other `State` functions, before the "transient" functions which target another concept.

HerbertJordan, line 281: To reduce confusion between `State` and `Storage` we should rename this function to `HasEmptyState`.

go/carmen/carmen.go
```diff
@@ -277,6 +277,9 @@ type TransactionContext interface {
 	// that gets reset after each transaction.
 	SetTransientState(Address, Key, Value)
 
+	// HasEmptyStorage returns true if account has empty storage.
```
HerbertJordan, line 280: The term "empty" has lead to a bit of confusion in the past, in particular since the storage of an account is not really empty. In fact, the storage of an account always has 2^256 elements in it, they may just be all zero.

We could rename this function to "HasZeroState" or "HasAllZeroState", but I guess "HasEmptyState" reads nicer. If we stick to that, however, we should explain in the documentation that an "empty" state is a state where all values are zero.

go/state/cppstate/cpp_state.go
```diff
@@ -139,7 +139,9 @@ func (cs *CppState) SetStorage(address common.Address, key common.Key, value com
 	update.AppendSlotUpdate(address, key, value)
 	return cs.Apply(0, update)
 }
-
+func (cs *CppState) HasEmptyStorage(common.Address) (bool, error) {
+	panic("HasEmptyStorage: not implemented for CppState")
```
HerbertJordan, line 143: This will let all Aida tests targeting a C++ implementation fail because they will still ask for the storage state.

For replying Fantom history, returning `true` here should be safe.

A cleaner alternative would be to define a "ErrEmptyStateNotImplemented" in the file defining the `State` interface and allow this error to be returned. At the calling this error should be checked and ignored by returning the "safe" fallback of `true` -- which should be valid for all Fantom networks.

However, a TODO + issue should be added requesting the implementation of this function.

go/state/gostate/archive_state.go
```diff
@@ -76,7 +76,9 @@ func (s *ArchiveState) GetStorage(address common.Address, key common.Key) (commo
 	}
 	return storage, s.archiveError
 }
-
+func (s *ArchiveState) HasEmptyStorage(common.Address) (bool, error) {
+	panic("HasEmptyStorage: not implemented for ArchiveState")
```
HerbertJordan, line 80: As for the C++ case, it might be that this is breaking nightly builds of Carmen on LaScala. We should apply the same (temporary) fix here (and at other places) as for the C++ implementation.

go/state/gostate/go_state.go
```diff
@@ -144,6 +144,14 @@ func (s *GoState) GetStorage(address common.Address, key common.Key) (common.Val
 	return val, s.stateError
 }
 
+func (s *GoState) HasEmptyStorage(address common.Address) (bool, error) {
+	if err := s.stateError; err != nil {
+		return false, err
+	}
+
+	return s.live.HasEmptyStorage(address)
```
HerbertJordan, line 152: I guess this would be the place to check for the dedicated not-implemented error to avoid `panic` triggered crashes in Aida.

go/state/gostate/go_state_test.go
```diff
@@ -822,6 +822,16 @@ func TestState_All_Archive_Operations_May_Cause_Failure(t *testing.T) {
...
 }
 
+func TestGoState_HasEmptyStorage(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	liveDB := state.NewMockLiveDB(ctrl)
+	db := newGoState(liveDB, nil, []func(){})
+
+	liveDB.EXPECT().HasEmptyStorage(address1)
+
+	db.HasEmptyStorage(address1)
```
HerbertJordan, line 832: nit: let the mock return a value and make sure your test call produces the same

go/state/state.go
```diff
@@ -85,6 +88,8 @@ type LiveDB interface {
 	GetBalance(address common.Address) (balance common.Balance, err error)
 	GetNonce(address common.Address) (nonce common.Nonce, err error)
 	GetStorage(address common.Address, key common.Key) (value common.Value, err error)
+	// HasEmptyStorage returns true if account has empty storage.
```
HerbertJordan, line 91: nit: I would remove this document line just because nothing else here is documented ...

go/state/state_db.go
```diff
@@ -52,6 +52,7 @@ type VmStateDB interface {
 	SetState(common.Address, common.Key, common.Value)
 	GetTransientState(common.Address, common.Key) common.Value
 	SetTransientState(common.Address, common.Key, common.Value)
+	HasEmptyStorage(common.Address) bool
```
HerbertJordan, line 55: "HasEmptyStorage" => "HasEmptyState"; also, move it 2 lines up, before the transient state functions.

go/state/state_db.go
```diff
@@ -827,6 +828,33 @@ func (s *stateDB) SetState(addr common.Address, key common.Key, value common.Val
 	}
 }
 
+func (s *stateDB) HasEmptyStorage(address common.Address) bool {
+	var (
+		isEmpty, shouldReturn bool
+	)
+	if !s.Exist(address) {
```
HerbertJordan, line 835: The joke is: an account may not exist but have some non-empty storage. This is what EIP-7610 is addressing. So I think  the `Exist` check should not be done her.

go/state/state_db.go
```diff
@@ -827,6 +828,33 @@ func (s *stateDB) SetState(addr common.Address, key common.Key, value common.Val
...
+		return true
+	}
+
+	s.data.ForEach(func(id slotId, val *slotValue) {
+		if id.addr == address && val.current != (common.Value{}) {
+			isEmpty = false
```
HerbertJordan, line 841: The `isEmpty` flag is unnecessary. You initialize it to `false` and you update it here to `false`.

Simpler: rename "shouldReturn" to "hasNonZeroValue" and in line 847 just return `false`.

go/state/state_db_test.go
```diff
@@ -4372,6 +4372,43 @@ func TestStateDB_SetTransientState_RollbackWithMultipleSteps(t *testing.T) {
...
+
+func TestStateDB_HasEmptyStorage_StateIsCheckedIfNotFoundInCache(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	mock := NewMockState(ctrl)
+	db := CreateStateDBUsing(mock)
+
+	mock.EXPECT().Exists(address1).Return(true, nil)
+	mock.EXPECT().HasEmptyStorage(address1)
+
+	db.HasEmptyStorage(address1)
```
HerbertJordan, line 4409: You should check the result of this call.

HerbertJordan (changes requested):
Looks good, but there are a few more things to improve. Please see the comments.

HerbertJordan:
While reviewing the PR I also realized: in Carmen S5 the type of account EIP-7610 tries to protect can not exist.

Such an account needs to have a balance of 0, a nonce of 0, and an empty code. In S5, this would mean that its `AccountInfo` struct is zero, and thus the account node would be removed from the MPT -- implicitly deleting its storage.

We have the rules of EIP-158 baked into our MPT design. Thus, it might not actually be possible to recreate test cases checking for this with Carmen.

go/state/state_db.go
petr-hanzl: I agree, after your explanation above, this check makes zero sense

petr-hanzl:
> We have the rules of EIP-158 baked into our MPT design. Thus, it might not actually be possible to recreate test cases checking for this with Carmen.

This might cause issues with ethereum test cases, do we need to talk about this more or do we just keep that in mind for now.

I feel like this code is not even needed, because there should not be a single call that would return `false`...?

go/state/gostate/go_state.go
petr-hanzl: This place wont work for `cpp` state, will it?

kjezek:
Looks to me like we do not have to merge this PR, it just opens more questions. I thought at the beginning we need it for the ethereum update, but it does not seem to be the case.

HerbertJordan:
Indeed, there might not be an actual need for this and we should avoid introducing stuff that is dead code.

Let's close this PR and the associated issue with a summary of our findings and if some day we actually do identify a need for this, it can be revived.

PR #925, opened 2024-06-18
Use `amount.Amount` in `UpdateTarget` interface

Mike-CZ:
This PR uses `amount.Amount` type in `UpdateTarget` interface. Should be merged after https://github.com/Fantom-foundation/Carmen/pull/921.


go/common/update.go
```diff
@@ -283,15 +285,16 @@ func UpdateFromBytes(data []byte) (Update, error) {
 
 	// Read list of balance updates
 	if balancesSize > 0 {
-		if len(data) < int(balancesSize)*(len(Address{})+len(Balance{})) {
+		balanceLength := len(amount.Amount{}.Bytes32())
```
HerbertJordan, line 288: nit: this is a long form of writing
```
balanceLength := 32
```
in which case it could also be written as
```suggestion
		const balanceLength = 32
```

go/common/update.go
```diff
@@ -353,7 +356,7 @@ func (u *Update) ToBytes() []byte {
 	size := 1 + 6*4 // version + sizes
 	size += len(u.DeletedAccounts) * addrLength
 	size += len(u.CreatedAccounts) * addrLength
-	size += len(u.Balances) * (addrLength + len(Balance{}))
+	size += len(u.Balances) * (addrLength + len(amount.Amount{}.Bytes32()))
```
HerbertJordan, line 359: I guess you can use a constant 32 here too.

go/database/mpt/io/archive.go
```diff
@@ -304,11 +305,11 @@ func importArchive(liveDbDir, archiveDbDir string, in io.Reader) (err error) {
 			context.deleteAccount()
 
 		case 'B':
-			balance := common.Balance{}
+			balance := amount.New().Bytes32()
```
HerbertJordan, line 308: maybe just
```suggestion
			var balance [32]byte
```

go/database/mpt/io/live.go
```diff
@@ -238,10 +238,11 @@ func runImport(directory string, in io.Reader, config mpt.MptConfig) (root mpt.N
 			if _, err := io.ReadFull(in, addr[:]); err != nil {
 				return root, hash, err
 			}
-			if _, err := io.ReadFull(in, balance[:]); err != nil {
+			b := amount.New().Bytes32()
```
HerbertJordan, line 241:
```suggestion
			var b [32]byte
```

HerbertJordan (approved):
Thanks :)

go/common/update.go
Mike-CZ: I wanted to avoid "magic" constants, but in this case it might sense.

kjezek (approved):
looks good to me, thanks

kjezek (approved):

PR #928, opened 2024-06-19
witness proof, add end to end test

kjezek:
This PR creates an end-to-end test for witness proof.

It fills a forest with test addresses and keys and then try to extract proofs and prove the selected keys


kjezek:
will by replaced by #942

PR #930, opened 2024-06-20
introduces decoded account node to store potentially long paths

### LINKED ISSUE #900: account node decoded from RLP need to store 32byte hashed address
Currently implemented RLP decoder stores hashes part of the account address in to the field `AccountNode.address`. This field is only 20bytes wide, while the hashed address may be up to 32bytes. 

The idea is to create a specialised node type that could handle 32byte paths:

```
type DecodedAccountNode struct {
   AccountNode
   partialPath Path
}
```

the RLP decoder would store to this node type instead of the `AccountNode`

kjezek:
This PR introduces a new variant of node `DecodedAccountNode`. This is used for an `AccountNode` when decoding from RLP.  The hashed address - the path - may be up to 32bytes and for this reason it does not fit into 20bytes addresses. 

This new type wraps the AccountNode and adds a path to store the RLP decoded Path.  This is furthermore used in `WitnessProof` utilities. 

It fixes #900

### DIFF
--- a/go/database/mpt/decoder.go
+++ b/go/database/mpt/decoder.go
@@ -141,9 +141,6 @@ func decodeAccountFromRlp(path Path, items rlp.List) (Node, error) {
 		return nil, fmt.Errorf("invalid balance: %v", err)
 	}
 
-	var address common.Address
-	copy(address[:], path.GetPackedNibbles()) // it does not cover full key as it is not available in RLP.
-
 	storageHashStr, ok := items.Items[2].(rlp.String)
 	if !ok {
 		return nil, fmt.Errorf("invalid storage hash type: got: %T, wanted: String", items.Items[2])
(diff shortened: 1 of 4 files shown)


go/database/mpt/decoder.go
```diff
@@ -261,3 +257,11 @@ func compactPathToNibbles(path []byte) []Nibble {
 
 	return res[2-odd:]
 }
+
+// DecodedAccountNode is an extension of the  account node with the path.
+// It is used for storing hashed path. which is part of the address
+// potentially up to 32bytes long, i.e. exceeding plain 20bytes long address.
+type DecodedAccountNode struct {
```
HerbertJordan, line 264: Should the visibility be private?

HerbertJordan (approved):
Looks good, I think this is a clean solution.

Maybe consider making the `DecodedAccountNode` private.

go/database/mpt/decoder.go
kjezek: yes, I think it makes sense

HerbertJordan (dismissed):

HerbertJordan (approved):

Issue #919, opened 2024-06-17
Add support for EIP-6780 (new Self-Destruct)


HerbertJordan:
With [EIP-6780](https://eips.ethereum.org/EIPS/eip-6780) the semantic of the Self-Destruct operation was modified. While up until this point and account could self-destruct itself, with this change this ability was restricted to accounts created as part of the same transaction.

This is a feature of the transaction layer, between the EVM and the database. Since this layer is covered by Carmen, it needs to be implemented.

The proper place is likely the [VmStateDB](https://github.com/Fantom-foundation/Carmen/blob/ad1657431037430a60f7456088947e5bfd16972a/go/state/state_db.go#L31) interface, by providing a modified version of the [Sucide](https://github.com/Fantom-foundation/Carmen/blob/ad1657431037430a60f7456088947e5bfd16972a/go/state/state_db.go#L37) function. Furthermore, this new `Self-Destruct` function needs to be exposed through the Carmen interface, analog to the current `Suicide`/`SelfDestruct` function, as well as a `CreateContract` function to be used by the EVM to signal the creation of accounts. With the latter, accounts are identified which are allowed to be destructed within the same transaction.

PR #933, opened 2024-06-23
Add support for EIP-6780

Mike-CZ:
This PR solves this issue https://github.com/Fantom-foundation/Carmen/issues/919

### DIFF
--- a/go/state/state_db.go
+++ b/go/state/state_db.go
@@ -206,9 +206,9 @@ type stateDB struct {
 	// Tracks the clearing state of individual accounts.
 	clearedAccounts map[common.Address]accountClearingState
 
-	// A list of contracts created during the current transaction. This is used
+	// A set of contracts created during the current transaction. This is used
 	// to mark eligible accounts to be self-destructed according to EIP-6780.
-	createdContracts map[common.Address]bool
+	createdContracts map[common.Address]struct{}
 
 	// A list of operations undoing modifications applied on the inner state if a snapshot revert needs to be performed.
 	undo []func()
@@ -456,7 +456,7 @@ func createStateDBWith(state State, storedDataCacheCapacity int, canApplyChanges
 		accountsToDelete:  make([]common.Address, 0, 100),
 		undo:              make([]func(), 0, 100),
 		clearedAccounts:   make(map[common.Address]accountClearingState),
-		createdContracts:  make(map[common.Address]bool),
+		createdContracts:  make(map[common.Address]struct{}),
 		emptyCandidates:   make([]common.Address, 0, 100),
 		canApplyChanges:   canApplyChanges,
 	}
@@ -542,8 +542,11 @@ func (s *stateDB) CreateAccount(addr common.Address) {
 }
 
 func (s *stateDB) CreateContract(addr common.Address) {
+	if _, exists := s.createdContracts[addr]; exists {
+		return
+	}
 	s.createAccountIfNotExists(addr)
-	s.createdContracts[addr] = true
+	s.createdContracts[addr] = struct{}{}
 	s.undo = append(s.undo, func() {
 		delete(s.createdContracts, addr)
 	})
@@ -597,20 +600,15 @@ func (s *stateDB) Suicide(addr common.Address) bool {
 
 // SuicideNewContract marks the given contract account as suicided.
 // If called in the same transaction scope after CreateContract, works
-// same as Suicide, otherwise has no effect.
+// the same as Suicide, otherwise has no effect.
 func (s *stateDB) SuicideNewContract(addr common.Address) bool {
-	if !s.Exist(addr) {
-		return false
-	}
-
 	// If called in the same tx, behave as previously
-	if s.createdContracts[addr] {
+	if _, exists := s.createdContracts[addr]; exists {
 		return s.Suicide(addr)
 	}
 
 	// Otherwise do nothing
-
-	return true
+	return false
 }
 
 func (s *stateDB) HasSuicided(addr common.Address) bool {
@@ -1373,7 +1371,7 @@ func (s *stateDB) resetTransactionContext() {
 	s.undo = s.undo[0:0]
 	s.emptyCandidates = s.emptyCandidates[0:0]
 	s.logs = s.logs[0:0]
-	s.createdContracts = make(map[common.Address]bool)
+	s.createdContracts = make(map[common.Address]struct{})
 }
 
 func (s *stateDB) ResetBlockContext() {
--- a/go/state/state_db.go
+++ b/go/state/state_db.go
@@ -34,7 +34,9 @@ type VmStateDB interface {
 	Exist(common.Address) bool
 	Empty(common.Address) bool
 
+	CreateContract(common.Address)
 	Suicide(common.Address) bool
+	SuicideNewContract(common.Address) bool
 	HasSuicided(common.Address) bool
 
 	// Balance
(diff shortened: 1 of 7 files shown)


go/state/state_db.go
```diff
@@ -580,6 +595,24 @@ func (s *stateDB) Suicide(addr common.Address) bool {
...
+	}
+
+	// If called in the same tx, behave as previously
+	if s.createdContracts[addr] {
+		return s.Suicide(addr)
+	}
+
+	// Otherwise do nothing
+
+	return true
```
Mike-CZ, line 613: Should we return `false` when it act as a no-op? In fact it is not getting removed, so `false` might make more sense...

go/state/state_db.go
```diff
@@ -534,6 +541,14 @@ func (s *stateDB) CreateAccount(addr common.Address) {
 	})
 }
 
+func (s *stateDB) CreateContract(addr common.Address) {
+	s.createAccountIfNotExists(addr)
+	s.createdContracts[addr] = true
```
HerbertJordan, line 546: Although there are external mechanisms that should prevent this, we should be defensive in the implementation of this function. Thus, it should be checked whether the given address is already in the `createdContracts` set, and if so, the function should ignore the extra call. Otherwise, snapshot restorations may reproduce wrong values.

go/state/state_db.go
```diff
@@ -204,6 +206,10 @@ type stateDB struct {
 	// Tracks the clearing state of individual accounts.
 	clearedAccounts map[common.Address]accountClearingState
 
+	// A list of contracts created during the current transaction. This is used
```
HerbertJordan, line 209:
```suggestion
	// A set of contracts created during the current transaction. This is used
```

go/state/state_db.go
```diff
@@ -204,6 +206,10 @@ type stateDB struct {
...
+	// to mark eligible accounts to be self-destructed according to EIP-6780.
+	createdContracts map[common.Address]bool
```
HerbertJordan, line 211: Since the `bool` value type has no semantic interpretation consider using an empty struct instead to stress the fields interpretation as a set:
```suggestion
	createdContracts map[common.Address]struct{}
```

go/state/state_db.go
HerbertJordan: I agree, the result should be `false`.

go/state/state_db.go
```diff
@@ -580,6 +595,24 @@ func (s *stateDB) Suicide(addr common.Address) bool {
 	return true
 }
 
+// SuicideNewContract marks the given contract account as suicided.
+// If called in the same transaction scope after CreateContract, works
+// same as Suicide, otherwise has no effect.
+func (s *stateDB) SuicideNewContract(addr common.Address) bool {
+	if !s.Exist(addr) {
```
HerbertJordan, line 602: Since the "Exist" is also tested in the `Suicide` you might skip it here and just do the following:
```
if s.createdContracts[addr] {
    return s.Suicide(addr)
}
return false
```

go/state/state_db_test.go
```diff
@@ -4389,6 +4389,79 @@ func TestNonCommittableStateDB_resetState_ClearsTransientStorage(t *testing.T) {
...
+	ctrl := gomock.NewController(t)
+	mock := NewMockState(ctrl)
+	db := CreateStateDBUsing(mock)
+
+	// the account should exist prior to CreateContract call
+	mock.EXPECT().Exists(address1).Return(true, nil)
+	db.CreateContract(address1)
+
+	// suicide the newly created contract in the same tx scope
+	db.SuicideNewContract(address1)
```
HerbertJordan, line 4402: nit: check the return value of this call; also in the other tests;

HerbertJordan (approved):
Looks good to me, thanks!

go/state/state_db.go
```diff
@@ -580,6 +595,24 @@ func (s *stateDB) Suicide(addr common.Address) bool {
 	return true
 }
 
+// SuicideNewContract marks the given contract account as suicided.
+// If called in the same transaction scope after CreateContract, works
+// same as Suicide, otherwise has no effect.
```
kjezek, line 600:
```suggestion
// the same as Suicide, otherwise has no effect.
```

go/state/state_db_test.go
```diff
@@ -4389,6 +4389,79 @@ func TestNonCommittableStateDB_resetState_ClearsTransientStorage(t *testing.T) {
...
+
+	// the contract account should stop existing at the end of the transaction
+	db.EndTransaction()
+	db.BeginTransaction()
+
+	if db.Exist(address1) {
+		t.Errorf("Account still exists after suicide")
```
kjezek, line 4409:
```suggestion
		t.Errorf("account still exists after suicide")
```

go/state/state_db_test.go
```diff
@@ -4389,6 +4389,79 @@ func TestNonCommittableStateDB_resetState_ClearsTransientStorage(t *testing.T) {
...
+
+	// suicide the contract from previous transaction
+	db.SuicideNewContract(address1)
+
+	// the contract should still exist when suicided in different transaction
+	db.EndTransaction()
+	db.BeginTransaction()
+
+	if !db.Exist(address1) {
+		t.Errorf("Account should exist after suicide in different transaction")
```
kjezek, line 4434:
```suggestion
		t.Errorf("account should exist after suicide in different transaction")
```

go/state/state_db_test.go
```diff
@@ -4389,6 +4389,79 @@ func TestNonCommittableStateDB_resetState_ClearsTransientStorage(t *testing.T) {
...
+
+	// suicide the contract
+	db.SuicideNewContract(address1)
+
+	// the contract account should still exist, because CreateContract method was rolled back
+	db.EndTransaction()
+	db.BeginTransaction()
+
+	if !db.Exist(address1) {
+		t.Errorf("Account should still exist after CreateContract call rollback")
```
kjezek, line 4461:
```suggestion
		t.Errorf("account should still exist after CreateContract call rollback")
```

go/state/state_db_test.go
```diff
@@ -4389,6 +4389,79 @@ func TestNonCommittableStateDB_resetState_ClearsTransientStorage(t *testing.T) {
...
+	// the contract should still exist when suicided in different transaction
+	db.EndTransaction()
+	db.BeginTransaction()
+
+	if !db.Exist(address1) {
+		t.Errorf("Account should exist after suicide in different transaction")
+	}
+}
+
+func TestStateDB_ContractCreationAndDeletionCanBeRolledBack(t *testing.T) {
```
kjezek, line 4438: I somehow cannot understand this test.
1. it creates an account
2. it rollbacks via snaphost - i.e. previous did not happen
3. It suicides the contract (which does not exist)
4.  the result is that the contract does exist?

kjezek (approved):
it looks good to me, though I may have a limited knowledge of what the semantics of this change is (even though I read the EIP)

One of the tests is not very clear to me, can you perhaps extend the doc?

go/state/state_db_test.go
Mike-CZ: In fact it does not have to necessarily create an account. The purpose of `CreateContract` is to mark an account as a contract (eventually creates it if it doesn't exist), so the `Suicide` method knows that it was "created" and can be removed in the same transaction scope.

 The test is verifying, that the `undo` operation on `CreateContract` is done. So in the end the account should still exist, because the `CreateContract` call was rolled back and the `Suicide` method worked as no-op, because there was no preceding `CreateContract` call.

kjezek: I see that the account was not suicided, but how can  it exist? I do not see where it was created at first place?

Mike-CZ: In the mock, I have set, that the state function `Exists` returns true. So when the state DB tries to check, wether the account exists, it gets true from the state and then it stores it into cache. https://github.com/Fantom-foundation/Carmen/blob/0d25cc18cb5920e00ea48ac44390faf3874ecda6/go/state/state_db.go#L485

kjezek: I see now. Perhaps it should be documented in this test as well, the semantics is not that obvious

kjezek (approved):

HerbertJordan (approved):

PR #934, opened 2024-06-23
Add tracing support to MPT benchmark tool

HerbertJordan:
This PR adds support for collecting [traces](https://pkg.go.dev/runtime/trace) to the MPT benchmark tool.

Tracing support can help identifying synchronization and IO issues in MPT operations.

This PR also fixes a bug: CPU profiling could not be disabled before this change.


kjezek (approved):
thanks, looks good

PR #935, opened 2024-06-23
Use atomic flag to speed up dirty node check

HerbertJordan:
This PR updates the `dirty` flag in nodes to be an atomic value. By doing so, determining whether a node is dirty can be conducted by a single atomic read operation. Before that, a lock hat to be required and released, involving multiple atomic read/write operations.

This change reduces the time required for flushing nodes. It also constitutes a preparation step for a follow-up PR introducing a  worker performing node-flushes in the background.

Impact of this change on the MPT performance (single sample each):
image

The data was collected by running
```
go run ./database/mpt/tool benchmark
```

Impact on locating dirty hashes in the CPU profile:
Before (total costs of shown operation is 0.81 seconds):
image

After (total costs of shown operation is 0.49 seconds):
image


HerbertJordan:
Replaced by #941

PR #936, opened 2024-06-23
Add worker flushing dirty nodes in the background

HerbertJordan:
This PR adds a background worker to the MPT forest that is periodically passing through the node cache, collecting dirty nodes, and flushing them to the disk.

The aim of this change is to reduce the required shutdown time of a Carmen S5 data base. The main contributor to the long shutdown delay is the need to flush out dirty nodes from the (extensive) in-memory node cache to the disk. Since the cache may contain 10s of millions of nodes, comprising several 10s of GB of data, writing this data to disk may take a substantial amount of time.

With this change, nodes are flushed gradually to the disk instead of delaying this step until shutdown time. This way, the number of dirty nodes that need to be flushed during the close of the database should be minimized.

The change was tested using the MPT tool's benchmark mode using the following command:
```
go run ./database/mpt/tool benchmark
```

Before the change, the final close of the database took ~42 seconds on the test system, retaining 10 million nodes in the cache, of which almost all have been dirty. After this change, only ~250K nodes needed to be flushed during the shutdown, reducing the shutdown time to 15 seconds.

The remaining 15 seconds are dominated by the time needed by the OS to flush out pending write operations and closing the file (~12 seconds). An attempt to perform this operation periodically as part of the background worker lead to a drastic decrease of the MPT performance (>90% throughput drop), thus this code was removed again. Furthermore, flushing pending changes at the end is the correct way of handling file-IO and its runtime does not depend on the node cache size. Thus, it is expected scale for large cache sizes.

In the benchmark stress-testing insertion performance, the addition of the background flusher reduces the throughput performance by ~15%:
image

TODO:
- [ ] test on larger systems (so far it was only tested on a single laptop)
- [ ] test performance impact on mainnet sync process

This change contributes to #938


thaarok:
@HerbertJordan I have just tried this (carmen version e35682746206) on server and after few minutes of syncing a got following panic:
```
panic: unable to store extension node with dirty hash

goroutine 69 [running]:
github.com/Fantom-foundation/Carmen/go/database/mpt.ExtensionNodeEncoderWithChildHash.Store({}, {0xc25e63b950?, 0x141dd00?, 0xdaa6?}, 0xc31468b180?)
        /home/jkalina/go/pkg/mod/github.com/!fantom-foundation/!carmen/go@v0.0.0-20240624065456-e35682746206/database/mpt/nodes.go:2247 +0x20f
github.com/Fantom-foundation/Carmen/go/backend/stock/file.(*fileStock[...]).Set(0x196c620, 0x3efbff, {{{0xc, 0xb3, 0x89, 0x94, 0x91, 0xa0, 0xec, 0x1d, ...}, ...}, ...})
        /home/jkalina/go/pkg/mod/github.com/!fantom-foundation/!carmen/go@v0.0.0-20240624065456-e35682746206/backend/stock/file/file.go:274 +0x13c
github.com/Fantom-foundation/Carmen/go/backend/stock/synced.(*syncedStock[...]).Set(0x0?, 0x0?, {{{0xc, 0xb3, 0x89, 0x94, 0x91, 0xa0, 0xec, 0x1d, ...}, ...}, ...})
        /home/jkalina/go/pkg/mod/github.com/!fantom-foundation/!carmen/go@v0.0.0-20240624065456-e35682746206/backend/stock/synced/synced.go:49 +0xf9
github.com/Fantom-foundation/Carmen/go/database/mpt.(*Forest).flushNode(0x1deca0919489b30c?, 0xf7862c2ed46dcd56?, {0x1963760?, 0xc38601bc20?})
        /home/jkalina/go/pkg/mod/github.com/!fantom-foundation/!carmen/go@v0.0.0-20240624065456-e35682746206/database/mpt/forest.go:894 +0x298
github.com/Fantom-foundation/Carmen/go/database/mpt.writeBufferSink.Write({0x1deca0919489b30c?}, 0xf7862c2ed46dcd56?, {{0xc33b9418c0}})
        /home/jkalina/go/pkg/mod/github.com/!fantom-foundation/!carmen/go@v0.0.0-20240624065456-e35682746206/database/mpt/forest.go:1037 +0x25
github.com/Fantom-foundation/Carmen/go/database/mpt.tryFlushDirtyNodes({0x1957e28, 0xc006b84000}, {0x1948460, 0xc005f61680})
        /home/jkalina/go/pkg/mod/github.com/!fantom-foundation/!carmen/go@v0.0.0-20240624065456-e35682746206/database/mpt/node_flusher.go:114 +0x2b0
github.com/Fantom-foundation/Carmen/go/database/mpt.startNodeFlusherWithConfig.func1()
        /home/jkalina/go/pkg/mod/github.com/!fantom-foundation/!carmen/go@v0.0.0-20240624065456-e35682746206/database/mpt/node_flusher.go:55 +0x157
created by github.com/Fantom-foundation/Carmen/go/database/mpt.startNodeFlusherWithConfig in goroutine 1
        /home/jkalina/go/pkg/mod/github.com/!fantom-foundation/!carmen/go@v0.0.0-20240624065456-e35682746206/database/mpt/node_flusher.go:46 +0x172
```
Do you have idea what could be the issue?

thaarok:
Shutdown after syncing 2 days of the chain has taken 33 seconds:
```
INFO [06-24|12:00:12.517] Capturing Carmen close cpuprofile        module=evm-store filename=/tmp/carmen-shutdown-1719230412
INFO [06-24|12:00:12.518] Closing State DB...                      module=evm-store
INFO [06-24|12:00:45.396] State DB closed                          module=evm-store
```
[carmen-shutdown-2days-e3568.cpuprofile](https://github.com/user-attachments/files/15955366/carmen-shutdown-2days-e3568.cpuprofile)

Data from original version to be added...

kjezek:
does not it need `nodeTransferMutex`, what if the workflow is:
* Flusher  - reads a node as non-dirty
* Main thread - evicts the node
* Main thread - puts the node back in he cache and modify
* Flusher - tries to write to disk, but the node is dirty?

thaarok:
> does not it need `nodeTransferMutex`, what if the workflow is:
> 
>  * Flusher  - reads a node as non-dirty
>  * Main thread - evicts the node
>  * Main thread - puts the node back in he cache and modify
>  * Flusher - tries to write to disk, but the node is dirty?

I don't see any issue here - the flusher just skips flushing this node and it is kept dirty for the flusher round.

thaarok:
Just confirming the `panic: unable to store extension node with dirty hash` still happens on rev ca0fa80. (Stacktrace identical with the already reported one.)

HerbertJordan:
@thaarok thanks for the feedback.

I am attempting to reproduce it in a unit test. So far, I managed to reproduce with a low probability a seemingly related issue where the final archive `Flush` operation fails. I will continue working on this today.

HerbertJordan:
@thaarok @kjezek I think I managed to reproducing it in an already preexisting unit test by ramping up the flush frequency. With this I think I understand the issue now:
- a dirty node is identified in the cache by the flusher and added to the list
- before the flusher gets the write permission, this node is released (as it gets deleted); when being released, the nodes dirty flag is reset to false, but not the dirty flags for child hashes
- when the flusher re-acquired the write lock, it was checking for a dirty hash (the node's hash) but not for the children's hash nor whether the node is now considered clean; this lead to miss the case of a node that got releases in the meanwhile

The fix for this was easy: the flusher needed to check again whether the current node was actually dirty before flushing it.

Let's run this fix and see whether this captured the issue.

kjezek (approved):
great that it got fixed

PR #939, opened 2024-06-25
Update storing of archive roots to be incremental

HerbertJordan:
This PR updates the export of roots to be incremental.

Before this change, every time an Archive was flushed or closed, the full list of roots has been dumped to the root file on the disk. In particular for large archives, covering 85 million blocks, this corresponds to writing more than 3 GiB of data to the disk.

With this change, the Archie is tracking how much roots are already stored on the disk and only writes incremental updates to the disk. As a side-effect, in consecutive `Flush` and `Close` calls, new roots are only written once.

For a benchmark stress-testing the storing of 1M root nodes in an archive, this results in a more than 90% reduction in the time required to flush an archive:
```
pkg: github.com/Fantom-foundation/Carmen/go/database/mpt
cpu: Intel(R) Core(TM) i7-6600U CPU @ 2.60GHz
                     │  before.log   │              after.log               │
                     │    sec/op     │    sec/op     vs base                │
ArchiveFlush_Roots-4   498.65m ± 17%   30.96m ± 14%  -93.79% (p=0.000 n=10)
```

This can also be seen in the CPU profiles. Before the change:
image

and after, the root hashes are no longer written during the `Close` operation since they have already been synced by the `Flush`:
image

### DIFF
--- a/go/database/mpt/archive_trie.go
+++ b/go/database/mpt/archive_trie.go
@@ -33,7 +33,7 @@ type ArchiveTrie struct {
 	head         LiveState // the current head-state
 	forest       Database  // global forest with all versions of LiveState
 	nodeSource   NodeSource
-	roots        []Root     // the roots of individual blocks indexed by block height
+	roots        rootList   // the roots of individual blocks indexed by block height
 	rootsMutex   sync.Mutex // protecting access to the roots list
 	rootFile     string     // the file storing the list of roots
 	addMutex     sync.Mutex // a mutex to make sure that at any time only one thread is adding new blocks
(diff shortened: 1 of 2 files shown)


go/database/mpt/archive_trie.go
```diff
@@ -358,19 +358,47 @@ func (a *ArchiveTrie) addError(err error) error {
 
 // ---- Reading and Writing Root Node ID Lists ----
 
-func loadRoots(filename string) ([]Root, error) {
+// rootList is a utility type managing an in-memory copy of the list of roots
+// of an archive and its synchronization with a on-disk file copy.
```
kjezek, line 362:
```suggestion
// of an archive and its synchronization with an on-disk file copy.
```

kjezek (approved):
it looks good to me

kjezek (approved):

PR #941, opened 2024-06-28
Improve efficiency of the scan for dirty nodes

HerbertJordan:
This PR applies some performance improvement on the scan for dirty nodes during flush operations. The main change is to track the dirty-state of nodes using atomic operations to facilitate checking for dirty nodes without the need of acquiring node-access permissions.

This reduces the time it takes to search for dirty nodes by more than 60%:
```
goos: linux
goarch: amd64
pkg: github.com/Fantom-foundation/Carmen/go/database/mpt
cpu: Intel(R) Core(TM) i7-6600U CPU @ 2.60GHz
                                              │  before.log  │              after.log               │
                                              │    sec/op    │    sec/op     vs base                │
NodeFlusher_CollectDirtyNodes/size=1000-4       33.796µ ± 4%   5.372µ ±  3%  -84.10% (p=0.000 n=10)
NodeFlusher_CollectDirtyNodes/size=10000-4      335.45µ ± 3%   54.32µ ±  7%  -83.81% (p=0.000 n=10)
NodeFlusher_CollectDirtyNodes/size=100000-4      3.891m ± 1%   1.462m ±  3%  -62.41% (p=0.000 n=10)
NodeFlusher_CollectDirtyNodes/size=1000000-4     39.50m ± 3%   14.70m ±  2%  -62.78% (p=0.000 n=10)
NodeFlusher_CollectDirtyNodes/size=10000000-4    399.0m ± 4%   142.6m ± 30%  -64.25% (p=0.000 n=10)
geomean                                          3.702m        978.0µ        -73.58%
```

### Notes for Reviewers
By introducing atomic operations on the `clean` flag in the `nodeBase` assignment operations on the node level like the one done in the `Forest` implementation (e.g. `*write.Get().(*AccountNode) = *node`) introduce a race condition since the reset of the `clean` flag is not atomic in this case. Thus, a new `Reset` function needed to be introduced, performing the `clean` flag reset using an atomic operation.

### DIFF
--- a/go/database/mpt/nodes.go
+++ b/go/database/mpt/nodes.go
@@ -781,9 +781,7 @@ func (n *BranchNode) setNextNode(
 		}
 		defer handle.Release()
 		newNode := handle.Get().(*BranchNode)
-		*newNode = *n
-		newNode.markDirty()
-		newNode.markMutable()
+		newNode.assign(n)
 		n = newNode
 		thisRef = &newRef
 		isClone = true
@@ -948,6 +946,16 @@ func (n *BranchNode) Reset() {
 	n.dirtyHashes = 0
 	n.embeddedChildren = 0
 	n.frozenChildren = 0
+}
+
+func (n *BranchNode) assign(other *BranchNode) {
+	n.nodeBase.markDirty()
+	n.nodeBase.markMutable()
+	n.children = other.children
+	n.hashes = other.hashes
+	n.dirtyHashes = other.dirtyHashes
+	n.embeddedChildren = other.embeddedChildren
+	n.frozenChildren = other.frozenChildren
 }
 
 func (n *BranchNode) MarkFrozen() {
@@ -1209,9 +1217,7 @@ func (n *ExtensionNode) setNextNode(
 				}
 				defer handle.Release()
 				newNode := handle.Get().(*ExtensionNode)
-				*newNode = *n
-				newNode.markDirty()
-				newNode.markMutable()
+				newNode.assign(n)
 				thisRef, n = &newRef, newNode
 				isClone = true
 			}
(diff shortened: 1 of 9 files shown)


go/database/mpt/nodes.go
```diff
@@ -473,10 +478,15 @@ func (c *nodeCheckContext) isCompatible(other *nodeCheckContext) bool {
 type nodeBase struct {
 	hash       common.Hash // the hash of this node (may be dirty)
 	hashStatus hashStatus  // indicating whether this node's hash is valid
-	clean      bool        // by default nodes are dirty (clean == false)
+	clean      int32       // by default nodes are dirty
```
kjezek, line 481: why not using `atomic.Bool`

kjezek (approved):
why is not used `atomic.Bool`. Otherwise it looks good, thanks

go/database/mpt/nodes.go
HerbertJordan: I was initially using an atomic Bool, but this caused multiple issues:
- `atomic.Bool` should not be copied and/or compared; however, we do so explicitly or implicitly in many situations; if this flag is switched to an `atomic.Bool`, `go vet` is reporting thousands of issues in our test code
- `atomic.Bool` can not be inline-initialized when defining structs; something like `ValueNode{base: nodeBase{clean: false}}` would have to be rewritten into two statements, one initializing the node and another explicitly marking it as dirty. This would touch a huge number of code locations

I switched between the `atomic.Bool` and this version several times, but in the end went for this one as the lesser of two evils.

kjezek: ok, thanks for explanation

kjezek (approved):

HerbertJordan:
This deemed unnecessary and abandoned for now. May be recovered in the future.

PR #942, opened 2024-06-28
Fill missing hashes for archive nodes when visiting path

kjezek:
This PR extends the method `CreateWitnessProof`, which iterates nodes from a root to a leaf node for a path.

This change fills children hashes (extension node, children of branch node, storage of account) back to the parents for archive node.

These hashes are missing when nodes are loaded from the archive, It caused wrong hashes computed from RLP of those nodes. It then produced a wrong witness proof.

### DIFF
--- a/go/database/mpt/hasher.go
+++ b/go/database/mpt/hasher.go
@@ -348,56 +348,9 @@ func (h ethHasher) updateHashesInternal(
 			// At this point the hashes of all children are up-to-date.
 			// They can now be transferred to the parents.
 			node := cur.handle.Get()
-			switch cur := node.(type) {
-			case *BranchNode:
-				for i := 0; i < len(cur.children); i++ {
-					if !cur.children[i].Id().IsEmpty() && cur.isChildHashDirty(byte(i)) {
-						handle, e := manager.getViewAccess(&cur.children[i])
-						if e != nil {
-							err = e
-							break
-						}
-						hash, dirty := handle.Get().GetHash()
-						if dirty {
-							panic("FATAL: detected dirty child of branch node\n")
-						}
-						cur.hashes[i] = hash
-						cur.setEmbedded(byte(i), embedded[cur.children[i].Id()])
-						handle.Release()
-					}
-				}
-				cur.clearChildHashDirtyFlags()
-			case *ExtensionNode:
-				if cur.nextHashDirty {
-					handle, e := manager.getViewAccess(&cur.next)
-					if e != nil {
-						err = e
-						break
-					}
-					hash, dirty := handle.Get().GetHash()
-					if dirty {
-						panic("FATAL: detected dirty child of extension node\n")
-					}
-					cur.nextIsEmbedded = embedded[cur.next.Id()]
-					cur.nextHash = hash
-					handle.Release()
-					cur.nextHashDirty = false
-				}
-			case *AccountNode:
-				if cur.storageHashDirty && !cur.storage.Id().IsEmpty() {
-					handle, e := manager.getViewAccess(&cur.storage)
-					if e != nil {
-						err = e
-						break
-					}
-					hash, dirty := handle.Get().GetHash()
-					if dirty {
-						panic("FATAL: detected dirty child of account node\n")
-					}
-					cur.storageHash = hash
-					handle.Release()
-					cur.storageHashDirty = false
-				}
+			if e := updateChildrenHashes(manager, node, embedded); e != nil {
+				err = e
+				break
 			}
 
 			// Test whether this node is to be embedded.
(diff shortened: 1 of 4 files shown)


go/database/mpt/hasher.go
```diff
@@ -749,6 +756,17 @@ func getEncodedPartialPathSize(numNibbles int) int {
...
 	source NodeSource,
+) (bool, error) {
+	return isHashedNodeEmbedded(node, source)
+}
+
+// isHashedNodeEmbedded determines whether the given node is an embedded node or not.
+// If information required for determining the embedded-state of the node is
+// marked dirty, this information is updated. Thus, calls to this function may
+// cause updates to the state of some nodes.
+func isHashedNodeEmbedded(
```
HerbertJordan, line 767: nit: rename "isHashedNodeEmbedded" => "isNodeEmbedded" ?

go/database/mpt/proof.go
```diff
@@ -47,8 +47,9 @@ type WitnessProof struct {
 func CreateWitnessProof(nodeSource NodeSource, root *NodeReference, address common.Address, keys ...common.Key) (WitnessProof, error) {
 	proof := proofDb{}
 	visitor := &proofExtractionVisitor{
-		nodeSource: nodeSource,
-		proof:      proof,
+		nodeSource:        nodeSource,
+		proof:             proof,
+		updateChildHashes: nodeSource.getConfig().HashStorageLocation == HashStoredWithNode,
```
HerbertJordan, line 52: This flag may be unnecessary, since each node tracks dirty child hashes already. Those dirty-flags are properly set by the decoder of the node, and in case of hashes being stored with the node, they should all be dirty.

So you could simply consider this flag to always to be `true` and rely on the dirty flags of the hashes in the nodes.

go/database/mpt/proof.go
```diff
@@ -225,6 +227,26 @@ func (p *proofExtractionVisitor) Visit(node Node, info NodeInfo) VisitResponse {
...
+			if err := updateChildrenHashesComputeEmbedded(p.nodeSource, n, n.children[:]); err != nil {
+				p.err = err
+				return VisitResponseAbort
+			}
+		case *AccountNode:
+			if err := updateChildrenHashesComputeEmbedded(p.nodeSource, n, []NodeReference{n.storage}); err != nil {
+				p.err = err
+				return VisitResponseAbort
+			}
+		}
```
HerbertJordan, line 247: I would prefer that the function `updateChildrenHashesComputeEmbedded` would handle the extraction of the list of children instead of the caller and then pass it as an argument.

go/database/mpt/proof.go
```diff
@@ -238,6 +260,30 @@ func (p *proofExtractionVisitor) Visit(node Node, info NodeInfo) VisitResponse {
 	return VisitResponseContinue
 }
 
+// updateChildrenHashesComputeEmbedded updates the hashes of the children of the given node.
+// Furthermore, it updates the embedded status of the children.
+func updateChildrenHashesComputeEmbedded(source NodeSource, node Node, children []NodeReference) error {
+	embeddedChildren := make(map[NodeId]bool, len(children))
+	for _, child := range children {
```
HerbertJordan, line 267: I think you could skip refreshing the embedded state of the child if the corresponding hash is not dirty. The dirty flag of the hash should cover both -- the hash and the embedded flag.

go/database/mpt/proof.go
```diff
@@ -238,6 +260,30 @@ func (p *proofExtractionVisitor) Visit(node Node, info NodeInfo) VisitResponse {
...
+			return err
+		}
+		embeddedChildren[child.Id()] = embedded
+	}
+
+	if err := updateChildrenHashes(source, node, embeddedChildren); err != nil {
+		return err
+	}
+
+	return nil
```
HerbertJordan, line 284:
```suggestion
	return updateChildrenHashes(source, node, embeddedChildren)
```

go/database/mpt/proof_test.go
```diff
@@ -274,6 +274,48 @@ func TestCreateWitnessProof_CannotCreateProof_FailingNodeSources(t *testing.T) {
...
+		t.Run(config.Name, func(t *testing.T) {
+			ctrl := gomock.NewController(t)
+
+			ctxt := newNiceNodeContextWithConfig(t, ctrl, config)
+
+			addressNibbles := AddressToNibblePath(address, ctxt)
+			desc := &Branch{
+				children: Children{
+					addressNibbles[0]: &Branch{
+						dirtyChildHashes: []int{1},
```
HerbertJordan, line 291: should this be the following?
```suggestion
						dirtyChildHashes: []int{addressNibbles[1]},
```

HerbertJordan (changes requested):
Looks good, but I added a few comments outlining ideas on how to reduce complexity a bit. Please have a look.

go/database/mpt/proof_test.go
kjezek: no, this marks a child dirty, this is not the nibble path

go/database/mpt/hasher.go
kjezek: ok, renamed

go/database/mpt/proof.go
kjezek: ok, good idea, updated.

go/database/mpt/proof.go
kjezek: ok, code inlined into this switch

go/database/mpt/proof.go
kjezek: ok, good idea, updated

go/database/mpt/proof.go
```diff
@@ -225,6 +225,72 @@ func (p *proofExtractionVisitor) Visit(node Node, info NodeInfo) VisitResponse {
...
+			}
+		}
+	case *AccountNode:
+		if n.storageHashDirty {
+			storageHandle, err := p.nodeSource.getViewAccess(&n.storage)
+			if err != nil {
+				p.err = err
+				return VisitResponseAbort
+			}
+			embedded, err := isNodeEmbedded(storageHandle.Get(), p.nodeSource)
```
HerbertJordan, line 281: The storage can never be embedded in the AccountNode -- see, for instance, the encoder [here](https://github.com/Fantom-foundation/Carmen/blob/8eff2394792a15d16569a683b5a70550e9003f45/go/database/mpt/hasher.go#L634-L638). 

So determining whether the storage is embedded can probably be skipped.

HerbertJordan (dismissed):

go/database/mpt/proof.go
kjezek: good catch, yes, you are right

HerbertJordan (approved):

PR #943, opened 2024-06-29
Fix synchronization issue in NodeCache

### LINKED ISSUE #940: Crash due to nil pointer deference in Test System
While doing a test run of the new background flushing feature the following panic was triggered:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xa75dd2]

goroutine 108 [running]:
github.com/Fantom-foundation/Carmen/go/database/mpt/shared.(*Shared[...]).TryGetViewHandle(0x1982160?)
        /home/herbert/go/pkg/mod/github.com/!fantom-foundation/!carmen/go@v0.0.0-20240627085016-204b50159b28/database/mpt/shared/shared.go:92 +0x12
github.com/Fantom-foundation/Carmen/go/database/mpt.tryFlushDirtyNodes.func1(0x0, 0xc145a2a000?)
        /home/herbert/go/pkg/mod/github.com/!fantom-foundation/!carmen/go@v0.0.0-20240627085016-204b50159b28/database/mpt/node_flusher.go:79 +0x2f
github.com/Fantom-foundation/Carmen/go/database/mpt.(*nodeCache).ForEach(0xc0030da740, 0xc4e6d780b0)
        /home/herbert/go/pkg/mod/github.com/!fantom-foundation/!carmen/go@v0.0.0-20240627085016-204b50159b28/database/mpt/node_cache.go:295 +0x4b
github.com/Fantom-foundation/Carmen/go/database/mpt.tryFlushDirtyNodes({0x19941a8, 0xc0030da740}, {0x1984160, 0xc003da05a0})
        /home/herbert/go/pkg/mod/github.com/!fantom-foundation/!carmen/go@v0.0.0-20240627085016-204b50159b28/database/mpt/node_flusher.go:78 +0xef
github.com/Fantom-foundation/Carmen/go/database/mpt.startNodeFlusher.func1()
        /home/herbert/go/pkg/mod/github.com/!fantom-foundation/!carmen/go@v0.0.0-20240627085016-204b50159b28/database/mpt/node_flusher.go:56 +0x157
created by github.com/Fantom-foundation/Carmen/go/database/mpt.startNodeFlusher in goroutine 1
        /home/herbert/go/pkg/mod/github.com/!fantom-foundation/!carmen/go@v0.0.0-20240627085016-204b50159b28/database/mpt/node_flusher.go:47 +0x158
exit status 2
```

HerbertJordan:
This PR fixes a synchronization issue in the node cache.

Before this PR, updates have not been correctly protected. `Get` and `ForEach` calls may observer inconsistent content, linking the wrong `NodeId` to the an incorrect `shared.Shared[Node]` instance.

The key insight of the new solution is that `tag`s are now utilized in two flavors: a stable version, in which accesses are allowed, and an in-transition version, in which no reads should be performed. This was the original intention for the synchronization mechanism, but the flaw in the implementation remained undiscovered until the integration of the background flusher.

Fixes #940


go/database/mpt/node_cache_test.go
```diff
@@ -425,3 +425,74 @@ func TestNodeCache_GetAndSetThreadSafety(t *testing.T) {
...
+										t.Errorf("for-each produced inconsistent key/value pair, wanted %v, got %v", want, got)
+									}
+								}
+							})
+						}
+					}
+				}()
+			}
+
+			for i, node := range nodes {
```
kjezek, line 474: it would make sense to use one more wait group that would start insertion of elements only after all 10 go-routines were really planned.  

It can happen at the moment (perhaps with a low probability) that no `ForEach` over the cache is executed when the main thread is too fast.

kjezek (approved):
It looks good. I would suggest to make sure that the go-routines in the test were executed. It certain condition it could happen now that the test succeeds without the go-routines being ever executed.

go/database/mpt/node_cache_test.go
HerbertJordan: Done.

kjezek (approved):
thx, looks good

PR #945, opened 2024-07-02
Add GetMemoryFootprint to public API

petr-hanzl:
This PR adds `GetMemoryFootprint` to public API via `Database` interface.


kjezek (changes requested):
I think we should return interfaces from our public API.

It means, the `carmen.MemoryFootprint` should be an interface and it should provide methods implemented by `*common.MemoryFootprint`.

go/carmen/memory_footprint.go
```diff
@@ -0,0 +1,72 @@
+package carmen
+
+import "github.com/Fantom-foundation/Carmen/go/common"
+
+type MemoryFootprint interface {
+	// GetChild returns a child of the memory footprint with the given name.
+	GetChild(name string) MemoryFootprint
+
+	// AddChild allows to attach a new value under current memory footprint.
+	AddChild(name string, childValue uintptr)
```
kjezek, line 10: I guess we do not need (want) the client to modify the memory footprint obtained from Carmen, i.e. methods `AddChild` and `SetNote` should not be in the API

kjezek (changes requested):
Can you perhaps provide an example (in the golang example format) how this API is intended to be used? 

At the moment the user gets a memory footprint, he can read `Value()` and `Total()` and possibly convert to `string`, but I do not see how he can iterate the tree of the memory footprint. 

In particular the method `GetChild` requires the name of a nested element, but the user does not know the names, i.e. I pretty much think he will get puzzled what to use as an input. 

I guess we would need some kind of a `Visitor` pattern that would drill down the trie and provide pairs of name and the size. 

In a minimal version we could provide only `String()` that prints the formatted result  without the possibility to iterate the tree, avoiding `GetChild`.  

Furthermore, I think methods that modify the footprint should be removed, I do not see why the client would need to modify the memory footprint once he gets it from Carmen

petr-hanzl:
> In particular the method `GetChild` requires the name of a nested element, but the user does not know the names, i.e. I pretty much think he will get puzzled what to use as an input.
@kjezek should we provide list of all possible children for both `LiveDB` and `Archive`?

kjezek:
> > In particular the method `GetChild` requires the name of a nested element, but the user does not know the names, i.e. I pretty much think he will get puzzled what to use as an input.
> > @kjezek should we provide list of all possible children for both `LiveDB` and `Archive`?

no, I do not think so, this API should stay generic

go/carmen/memory_footprint.go
```diff
@@ -0,0 +1,77 @@
...
+// On the date above, in accordance with the Business Source License, use of
+// this software will be governed by the GNU Lesser General Public License v3.
+
+package carmen
+
+import "github.com/Fantom-foundation/Carmen/go/common"
+
+type MemoryFootprint interface {
+	// GetChild returns a child of the memory footprint with the given name.
+	GetChild(name string) MemoryFootprint
```
kjezek, line 17: I guess this method should be removed, right? It is still hard to use when the list of names is not known.

go/carmen/memory_footprint.go
```diff
@@ -0,0 +1,77 @@
...
+
+	// ToString provides the memory footprint as a tree summary in a string
+	// The name param allows to give a name to the root of the tree.
+	ToString(name string) string
+
+	// String allow memory footprints to be used in format strings.
+	String() string
+
+	// Visit iterates the footprint tree (including all subcomponents) and applies visit onto them
+	Visit(visit func(footprint MemoryFootprint))
```
kjezek, line 33: I think the only reason why the wrapper `memoryFootprint` is needed is this recursive dependency to `MemoryFootprint`. If the callback returns e.g. a tuple `depth, name, size` the wrapper could be avoided and the `carmen` package would define only the interface.

Also the parameter `depth` would be very useful, otherwise the user does not know where he is currently in the tree.

As I said in previous comments round, it would be very useful to create an example how to use this new API - it would quickly discover all shortcomings

go/carmen/example_test.go
```diff
@@ -282,3 +282,51 @@ func ExampleDatabase_GetHistoricContext() {
...
+	if err := db.Flush(); err != nil {
+		log.Fatalf("cannot flush: %v", err)
+	}
+
+	fp := db.GetMemoryFootprint()
+
+	total := fp.Total()
+	if total <= 0 {
+		log.Fatalf("database was filled with block but memory footprint returned 0 bytes used")
+	}
```
kjezek, line 320: I would remove this assertion, this is not a unit test checking that `Total` is returning a correct value, it is an example showing the user how to use this API

go/carmen/memory_footprint.go
```diff
@@ -0,0 +1,47 @@
...
+//
+// On the date above, in accordance with the Business Source License, use of
+// this software will be governed by the GNU Lesser General Public License v3.
+
+package carmen
+
+import "github.com/Fantom-foundation/Carmen/go/common"
+
+type MemoryFootprint interface {
+	// Total provides the amount of bytes consumed by the database structure including all its subcomponents
```
kjezek, line 16:
```suggestion
	// Total provides the number of bytes consumed by the database structure including all its subcomponents.
```

go/carmen/memory_footprint.go
```diff
@@ -0,0 +1,47 @@
...
+	Total() uint64
+
+	// String allow memory footprints to be used in format strings.
+	// It returns a breakdown of all MemoryFootprints within the tree.
+	String() string
```
kjezek, line 21: I would remove this. When using the` %s` pattern, `String() `is called implicitly, i.e. it does not have to be explicitly defined in the interface. And If I am not wrong, the convention is that this method is usually not explicitly defined in interfaces.  Furthermore, there is a build-in interface `Stringer` for this method, which can be uses at places where needed. This is the beauty of Golang that you can implement interface methods, but you do not have to explicitly say it.

go/carmen/memory_footprint.go
```diff
@@ -0,0 +1,47 @@
...
+// Change Date: 2028-4-16
+//
+// On the date above, in accordance with the Business Source License, use of
+// this software will be governed by the GNU Lesser General Public License v3.
+
+package carmen
+
+import "github.com/Fantom-foundation/Carmen/go/common"
+
+type MemoryFootprint interface {
```
kjezek, line 15: This interface should also have its `memory_footprint_test.go`, even if its implementation is trivial.

go/carmen/memory_footprint.go
```diff
@@ -0,0 +1,47 @@
...
+	String() string
+}
+
+func newMemoryFootprint(fp *common.MemoryFootprint) MemoryFootprint {
+	return &memoryFootprint{
+		fp: fp,
+	}
+}
+
+func NewMemoryFootprintFromValue(value uint64) MemoryFootprint {
```
kjezek, line 30: Is this method needed? If yes, I guess  it should be private anyway, we only want provide the footprint to users, but they do not need to create their own instances.

kjezek (changes requested):
Thanks for this effort. I think now the interface is much easier to understand thanks to the fact that it is very simple. We can always extend it. 

I put some more comments, please have a look. In general it looks good.

go/carmen/memory_footprint.go
petr-hanzl: Are tests really necessary? The tests would only test already tested funcions since the interface `MemoryFootprint` only does not add any logic.

kjezek: it is the ideal situation to test each layer independently because the code can move around in the future and the dependencies between layers may change and the path that is accidentally tested now from another layer can get lost. 

We do not have this ideal situation everywhere at the moment, but we aim at improving the code all the time, and especially here it is really a few lines of code, i.e. a very low effort needed to reach this ideal state.

go/carmen/memory_footprint_test.go
```diff
@@ -0,0 +1,28 @@
...
+	if err != nil {
+		t.Fatalf("cannot open test database: %v", err)
+	}
+	fp := db.GetMemoryFootprint()
+
+	if fp.Total() <= 0 {
+		t.Error("memory footprint returned 0 for existing open database")
+	}
+
+	s := fp.(*memoryFootprint).String()
```
kjezek, line 19:
```suggestion
	s := fmt.Sprintf("%s", fp)
```

kjezek (approved):
looks good, thanks

kjezek (dismissed):

kjezek (approved):

PR #947, opened 2024-07-08
verify witness proof tool

kjezek:
This PR provides a new CLI utility to verify witness proof against a database. 

It can read either a live or archive db and extract a proof for all accounts and storage slots. The proof is then verified agains data stored in the database, and it is checked if the data matches. 

For the live db it analyses the head state, for archive a block range or a single block can be specified.


go/database/mpt/verification_proof.go
```diff
@@ -0,0 +1,277 @@
...
+// It reads the trie for each block of the archive and loads all accounts and storage slots.
+// Then, it extracts witness proofs for the same accounts and storage slots of each block.
+// Finally, the actual values from the trie are compared with the values extracted from the witness proofs.
+func VerifyProofArchiveTrie(ctx context.Context, dir string, config MptConfig, from, to int, observer VerificationObserver) (err error) {
+	trie, err := OpenArchiveTrie(dir, config, NodeCacheConfig{})
+	if err != nil {
+		return err
+	}
+	defer func() {
+		err = trie.Close()
```
petr-hanzl, line 32: err is not checked here

go/database/mpt/tool/verify_proof.go
```diff
@@ -0,0 +1,92 @@
...
+		Name:  "block-to",
+		Usage: "the ending block number to verify",
+		Value: 0,
+	}
+	blockNum = cli.IntFlag{
+		Name:  "block-num",
+		Usage: "block number to verify",
+		Value: 0,
+	}
+)
```
petr-hanzl, line 51: Do I understand correctly that user can either choose range of blocks using `--blockFrom` and `--blockTo` or a single block `--blockNum`?

If so, is that really necessary? Shouldn't we just use `--blockFrom 1 --blockTo 1` as single block? I think it removes some possible confusions.

go/database/mpt/verification_proof.go
```diff
@@ -0,0 +1,277 @@
...
+		return nil
+	}
+
+	if from < 0 {
+		from = 0
+	}
+
+	if to <= 0 || to > int(blockHeight) {
+		to = int(blockHeight) + 1
+	}
```
petr-hanzl, line 59: I think we should inform user that he chose incorrect block range and we are about to modify it. Especially in the second case, first case is logical.

go/database/mpt/tool/verify_proof.go
```diff
@@ -0,0 +1,92 @@
...
+		if context.IsSet(blockNum.Name) {
+			from = context.Int(blockNum.Name)
+			to = from + 1
+		}
+		ctx := interrupt.CancelOnInterrupt(context.Context)
+		return mpt.VerifyProofArchiveTrie(ctx, dir, info.Config, from, to, &verificationObserver{})
+	}
+
+	if info.Config.Name == mpt.S5LiveConfig.Name {
+		ctx := interrupt.CancelOnInterrupt(context.Context)
```
petr-hanzl, line 86: I believe you can move the call `ctx := interrupt.CancelOnInterrupt(context.Context)` to line 73 to avoid duplicity

go/database/mpt/verification_proof.go
```diff
@@ -0,0 +1,277 @@
...
+
+	observer.Progress(fmt.Sprintf("Verifying total %d blocks", to-from))
+	for i := uint64(from); i < uint64(to); i++ {
+		observer.Progress(fmt.Sprintf("Collecting state for block... \t %d", i))
+		// collect all accounts for this block
+		visitor := accountCollectingVisitor{accounts: accountsMap{}, ctx: ctx}
+		if err := trie.VisitTrie(i, &visitor); err != nil {
+			observer.EndVerification(err)
```
petr-hanzl, line 67: I can see call to this func in most return cases. Why dont you put it into a `defer`?
If you want to avoid the call when the app has been interrupted, you can check the interrupt again in the defer. 
```
defer func() {
   if !interrupt.IsCancelled(ctx) {
      observer.EndVerification(err)
   }
}
```

Although I dont see the reason why you want to avoid it in this case.

go/database/mpt/verification_proof.go
```diff
@@ -0,0 +1,277 @@
...
+// It reads the trie for the head block and loads all accounts and storage slots.
+// Then, it extracts witness proofs for the same accounts and storage slots.
+// Finally, the actual values from the trie are compared with the values extracted from the witness proofs.
+func VerifyProofLiveTrie(ctx context.Context, dir string, config MptConfig, observer VerificationObserver) (err error) {
+	trie, err := OpenFileLiveTrie(dir, config, NodeCacheConfig{})
+	if err != nil {
+		return err
+	}
+	defer func() {
+		err = trie.Close()
```
petr-hanzl, line 122: err is not checked here

go/database/mpt/verification_proof.go
```diff
@@ -0,0 +1,277 @@
...
+// It reads the trie for the head block and loads all accounts and storage slots.
+// Then, it extracts witness proofs for the same accounts and storage slots.
+// Finally, the actual values from the trie are compared with the values extracted from the witness proofs.
+func verifyProofLiveTrie(ctx context.Context, trie *LiveTrie, observer VerificationObserver) error {
+	observer.StartVerification()
+	observer.Progress("Collecting states... ")
+	// collect all accounts for this block
+	visitor := accountCollectingVisitor{accounts: accountsMap{}, ctx: ctx}
+	if err := trie.VisitTrie(&visitor); err != nil {
+		observer.EndVerification(err)
```
petr-hanzl, line 138: Same case as in verifyProofArchiveTrie

petr-hanzl (changes requested):

go/database/mpt/verification_proof.go
kjezek: It is assigned to the output of this method. 

I do not very like this syntax, because I myself consider it confusing, but do not know if this can be done better in golang.

It is tested here: 
https://github.com/Fantom-foundation/Carmen/pull/947/files#diff-8bb57076b2dc35d3c0b28ecd9488c8c002f0408e571713e8462c7b7e8a6d06aeR106

go/database/mpt/tool/verify_proof.go
kjezek: right, it was meant for user convenience - exactly as you write, it is a shortcut for having to write` --blockFrom 1 --blockTo 1` - this itself is confusing, because I would guess it must have ben written as `--blockFrom 1 --blockTo 2`, i.e the single param variant prevents confusion

go/database/mpt/verification_proof.go
kjezek: ok, add logging for this

go/database/mpt/tool/verify_proof.go
kjezek: ok, one line I do not consider a duplicity :) and the drawback is that an extra go-routine is started and immediately stopped when none of these IF conditions apply.  But ok, will move

go/database/mpt/verification_proof.go
petr-hanzl: Oh you are right. I always miss this.

You can do something like this to make sure the reader sees the check, but its very unnecessary.
```suggestion
		closeErr := trie.Close()
		if closeErr != nil {
		   err = errors.Join(err, closeErr)
		}
```

go/database/mpt/tool/verify_proof.go
petr-hanzl: ic. Lets leave it as it is.

go/database/mpt/verification_proof.go
kjezek: it is called in the if conditions where an error has occurred, to propagate this error. 

I could define an error variable somewhere on a method global level, assign it everywhere when there is an error, and use `defer`, but I do not see it very much readable. 

I think now the error instance is encapsulated in enclosing `if` statement, which I think is more readable.

go/database/mpt/verification_proof.go
kjezek: I think this code is rather weird - `err` is always `nil` and `errors.Join` checks for `nil` as well  - i.e. for me it would be puzzling what this whole construct is good for

go/database/mpt/verification_proof.go
petr-hanzl: I like this better but I will leave it up to your preference then.

petr-hanzl (dismissed):

go/database/mpt/verification_proof.go
```diff
@@ -0,0 +1,279 @@
...
+
+		observer.Progress(fmt.Sprintf("Collecting proof for block.. \t %d \t Accounts: %d", i, len(visitor.accounts)))
+		// collect all proofs for this block
+		rawProof := make(map[string]struct{})
+		for addr, data := range visitor.accounts {
+			if interrupt.IsCancelled(ctx) {
+				return interrupt.ErrCanceled
+			}
+
+			accountProof, err := trie.CreateWitnessProof(i, addr, maps.Keys(data.storage)...)
```
HerbertJordan, line 85: The number of storage slots of an account could be huge for some accounts. I would suggest to batch it in groups of up to 10 keys per request.

go/database/mpt/verification_proof.go
```diff
@@ -0,0 +1,279 @@
...
+	return errors.Join(
+		verifyProofLiveTrie(ctx, trie, observer),
+		trie.Close())
+}
+
+// verifyProofLiveTrie verifies the consistency of witness proofs for an live trie.
+// It reads the trie for the head block and loads all accounts and storage slots.
+// Then, it extracts witness proofs for the same accounts and storage slots.
+// Finally, the actual values from the trie are compared with the values extracted from the witness proofs.
+func verifyProofLiveTrie(ctx context.Context, trie *LiveTrie, observer VerificationObserver) error {
```
HerbertJordan, line 134: The two methods `verifyProofLiveTrie` and `verifyProofArchiveTrie` have lots of code in common. The main difference is the extra loop over the blocks and the input type.

It would be great if the per-block verification code of `verifyProofArchiveTrie` could make use of the same code as the `verifyProofLiveTrie` function.

go/database/mpt/verification_proof.go
```diff
@@ -0,0 +1,279 @@
...
+		observer.Progress(fmt.Sprintf("Verifying proof for block... \t %d", i))
+		if err := verifyProofState(ctx, rootHash, proof, visitor.accounts); err != nil {
+			observer.EndVerification(err)
+			return err
+		}
+	}
+
+	observer.EndVerification(nil)
+	return nil
+}
```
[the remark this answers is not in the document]
HerbertJordan: The same is true for storage slots. The tool should also ask for a set of random keys with zero values and make sure that valid proofs are extracted.

HerbertJordan (changes requested):
Thanks for adding this tool. I think the way it is integrated (UX) is nice, except a few minor details I remarked in the comments.

I am more concerned about the resource requirements for this tool. It looks like to verify a state the entire state + proofs for each account and slot is loaded into memory and then being verified in a single step. This is not very efficient, and does not match the main use case, where proofs for individual bits and peaces are requested from the Archive.

I think the tool should be restructured to ask for individual proofs, and then verify those, without merging them into one proof that covers everything. This is closer to the real use case, and should scale much better. Also, this could be nicely parallelized.

Furthermore, I think some proofs for empty accounts and zero-valued storage slots should be included.

go/database/mpt/tool/verify_proof.go
```diff
@@ -0,0 +1,91 @@
...
+		to := context.Int(blockTo.Name)
+		if context.IsSet(blockNum.Name) {
+			from = context.Int(blockNum.Name)
+			to = from + 1
+		}
+		return mpt.VerifyProofArchiveTrie(ctx, dir, info.Config, from, to, &verificationObserver{})
+	}
+
+	if info.Config.Name == mpt.S5LiveConfig.Name {
+		return mpt.VerifyProofLiveTrie(ctx, dir, info.Config, &verificationObserver{})
```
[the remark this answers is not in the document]
kjezek: ok, added

go/database/mpt/verification_proof.go
```diff
@@ -0,0 +1,279 @@
...
+	"fmt"
+	"github.com/Fantom-foundation/Carmen/go/common"
+	"github.com/Fantom-foundation/Carmen/go/common/interrupt"
+	"github.com/Fantom-foundation/Carmen/go/common/witness"
+	"golang.org/x/exp/maps"
+	"log"
+)
+
+// VerifyProofArchiveTrie verifies the consistency of witness proofs for an archive trie.
+// It reads the trie for each block of the archive and loads all accounts and storage slots.
```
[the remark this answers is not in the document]
kjezek: ok, fixed

go/database/mpt/verification_proof.go
```diff
@@ -0,0 +1,279 @@
...
+		observer.EndVerification(err)
+		return err
+	}
+	if empty {
+		return nil
+	}
+
+	if from < 0 {
+		from = 0
+		log.Printf("adjusting to a minimum block height: %d", from)
```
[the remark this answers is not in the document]
kjezek: ok, updated

go/database/mpt/verification_proof.go
kjezek: right, unified so that it gets `*LiveTrie` from `*ArchiveTrie` and calls the method verifying the live trie

go/database/mpt/verification_proof.go
```diff
@@ -0,0 +1,279 @@
...
+		rootHash, err := trie.GetHash(i)
+		if err != nil {
+			observer.EndVerification(err)
+			return err
+		}
+
+		observer.Progress(fmt.Sprintf("Recovering proof for block... \t %d", i))
+		// match proofs and database data
+		proof := CreateWitnessProofFromNodes(maps.Keys(rawProof))
+		observer.Progress(fmt.Sprintf("Verifying proof for block... \t %d", i))
```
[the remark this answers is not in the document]
kjezek: right, thanks for this feedback, I reworked it as you suggest

go/database/mpt/verification_proof.go
kjezek: good idea, reworked this way, now it also works better when pressing ctrl+c as previously it could take even an hour when it was processing an account with many slots.

go/database/mpt/verification_proof.go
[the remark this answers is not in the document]
kjezek: ok, I added verification of 1000 random addresses + 10 random keys within non-empty addresses at the end of verification process

go/database/mpt/tool/verify_proof.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	ctx := interrupt.CancelOnInterrupt(context.Context)
+	if info.Config.Name == mpt.S5ArchiveConfig.Name {
+		from := context.Int(blockFrom.Name)
+		to := context.Int(blockTo.Name)
+		if context.IsSet(blockNum.Name) {
+			from = context.Int(blockNum.Name)
+			to = from + 1
+		}
```
HerbertJordan, line 82: Since there are multiple ways to specify a block range, I think it would be nice of the tool to log the range of blocks that are going to be checked once the configuration flags have been parsed. This gives users immediate feedback to check that the flags had the intended effect.

go/database/mpt/verification_proof.go
```diff
@@ -0,0 +1,336 @@
...
+func VerifyProofArchiveTrie(ctx context.Context, dir string, config MptConfig, from, to int, observer VerificationObserver) error {
+	trie, err := OpenArchiveTrie(dir, config, NodeCacheConfig{})
+	if err != nil {
+		return err
+	}
+
+	observer.StartVerification()
+	err = errors.Join(
+		verifyProofArchiveTrie(ctx, trie, from, to, observer),
+		trie.Close())
```
HerbertJordan, line 38: nit: I think formatting it like this helps readability:
```suggestion
	err = errors.Join(
		verifyProofArchiveTrie(ctx, trie, from, to, observer),
		trie.Close(),
	)
```

go/database/mpt/verification_proof.go
```diff
@@ -0,0 +1,336 @@
...
+	blockHeight, empty, err := trie.GetBlockHeight()
+	if err != nil {
+		return err
+	}
+	if empty {
+		return nil
+	}
+
+	if from < 0 || from > int(blockHeight) {
+		from = 0
```
HerbertJordan, line 53: This seems odd ... if from is after the actual block height, why would it start from 0? Shouldn't it start from `blockHeight`, effectively skipping all verification steps?

go/database/mpt/verification_proof.go
```diff
@@ -0,0 +1,336 @@
...
+		observer.Progress(fmt.Sprintf("adjusting to a minimum block height: %d", from))
+	}
+
+	if to <= 0 || to > int(blockHeight) {
+		to = int(blockHeight) + 1
```
HerbertJordan, line 58: The same in reverse here. It looks surprising that when I set the `to` flag to 0 it is overwritten with the maximum block height. What if I actually only want to check block 0?

go/database/mpt/verification_proof.go
```diff
@@ -0,0 +1,336 @@
...
+func VerifyProofLiveTrie(ctx context.Context, dir string, config MptConfig, observer VerificationObserver) error {
+	trie, err := OpenFileLiveTrie(dir, config, NodeCacheConfig{})
+	if err != nil {
+		return err
+	}
+
+	observer.StartVerification()
+	err = errors.Join(
+		verifyProofLiveTrie(ctx, trie, observer),
+		trie.Close())
```
HerbertJordan, line 91:
```suggestion
	err = errors.Join(
		verifyProofLiveTrie(ctx, trie, observer),
		trie.Close(),
	)
```

go/database/mpt/verification_proof.go
```diff
@@ -0,0 +1,336 @@
...
+				percentReported = int(percent) + 10 // report step min 10%
+			}
+		}
+	}
+	// finally, that addresses and keys not present in the proof are empty
+	return verifyProofEmptyValues(ctx, trie, rootHash, maps.Keys(visitor.accounts), 1000, observer)
+}
+
+// verifyProofEmptyValues verifies that the witness proof for addresses and keys that lead to empty accounts
+// and empty slots, because they are not present in the MPT and in a generated proof.
```
HerbertJordan, line 161: nit: this sentence misses some word(s) ...

go/database/mpt/verification_proof.go
HerbertJordan, line 1: Should this file be moved to the tool directory? It seems like the provided functionality is more or less exclusively intended for the proof verification sub-command.

go/database/mpt/verification_proof.go
```diff
@@ -0,0 +1,336 @@
...
+		if got, want := proofValue, storage[key]; got != want {
+			return fmt.Errorf("storage mismatch for 0x%x, key 0x%x, got %v, want %v", addr, key, got, want)
+		}
+	}
+
+	return nil
+}
+
+// generateUnknownAddresses generates a slice of addresses that will never appear
+// in real MPT and the witness proof. Each address consists of four random
```
HerbertJordan, line 255: "never appear" is not guaranteed by this function. It is just very unlikely.

If you want to make sure that they do not appear, you should pass the trie to check that the address indeed does not exist.

Alternatively, you create `number` random addresses and where you use them, you do not assume that those accounts are empty but you look them up in the MPT. Most of them will be empty, but if by chance you do have an address that is in the MPT, you compare with the actual account state.

go/database/mpt/verification_proof.go
```diff
@@ -0,0 +1,336 @@
...
+	res := make([]common.Address, number)
+	for i := range res {
+		j := rand.Int()
+		res[i] = common.Address{byte(j), byte(j >> 8), byte(j >> 16), byte(j >> 24), 1}
+	}
+	return res
+}
+
+// generateUnknownKeys generates a slice of keys that will never appear in real
+// MPT and the witness proof. Each key consists of four random bytes, a number
```
HerbertJordan, line 269: Same as for the addresses. They are unlikely, but possible. I would suggest to use them as random keys and compare the proof-content with the MPT content as also suggested for accounts.

go/database/mpt/verification_proof.go
```diff
@@ -0,0 +1,336 @@
...
+		ranges = append(ranges, keys[i:end])
+	}
+
+	return ranges
+}
+
+type accountsMap map[common.Address]accountData
+
+type accountCollectingVisitor struct {
+	accounts       accountsMap
```
HerbertJordan, line 307: Collecting this full account map for a large state will require a large amount of memory. Basically, it brings the entire state information of all accounts and all slots into the main memory. This could be easily several 10s of GiB of data.

go/database/mpt/verification_proof.go
```diff
@@ -0,0 +1,336 @@
...
+	if hints != nil {
+		hints.Release()
+	}
+	if err != nil {
+		return err
+	}
+
+	observer.Progress("Collecting states... ")
+	// collect all accounts for this block
+	visitor := accountCollectingVisitor{accounts: accountsMap{}, ctx: ctx}
```
HerbertJordan, line 107: Running the collector here requires a lot of memory, since it is essentially loading the entire state DB into memory.

To be scalable for large DBs, the visiting of accounts should be merged with the checking of proofs. You run the visitior, and for every account you encounter, you run the proof extraction inside the visitor and check the results. Then you load the slots and check those as well (this should be Ok, although there are accounts with >1GiB of storage); Once the account-iterator is moving on, all state is abandoned.

HerbertJordan (changes requested):

go/database/mpt/tool/verify_proof.go
kjezek: ok, added log msg

go/database/mpt/verification_proof.go
kjezek: ok, I simplified that just `to > blockHeight` is checked so the `--block-to` can be omitted or set to a huge number and it is set to max block. Other variants are considered a user error, I realised that there is so many variants, e.g.  `to < from`, with` to/from` being positive, negative, etc. and it is questionable what should happen when for instance both are negative, both are above range, etc.

go/database/mpt/verification_proof.go
kjezek: fixed

go/database/mpt/verification_proof.go
kjezek: there is an issue with private methods - there is many in the `mpt` package, mostly for test, but also for the production code.

go/database/mpt/verification_proof.go
kjezek: I see your point, I wanted to save on MPT lookups as an address with 16 trailing zeros is very unlikely in real scenario, but I understand that the verification tool should be rigorous.

go/database/mpt/verification_proof.go
kjezek: fixed by skipping potentially non-empty keys/addresses

go/database/mpt/verification_proof.go
kjezek: I updated the visitor that it check each address immediately as spotted, furthermore the keys are checked by batches of 10.

go/database/mpt/verification_proof.go
kjezek: changed this way, thanks for this feedback

go/database/mpt/verification_proof.go
```diff
@@ -0,0 +1,359 @@
...
+			return fmt.Errorf("storage mismatch for 0x%x, key 0x%x, got %v, want %v", addr, key, got, want)
+		}
+	}
+
+	return nil
+}
+
+// generateUnknownAddresses generates a slice of addresses that does not appear in the input MPT.
+// if in an unlikely situation a generated address is in the trie, it is not added in the result list.
+// It means that in an unlikely situation, the result list may contain fewer addresses than the input number.
```
HerbertJordan, line 211:
```suggestion
// generateUnknownAddresses generates a slice of addresses that do not appear in the input MPT.
```

go/database/mpt/verification_proof.go
```diff
@@ -0,0 +1,359 @@
...
+func generateUnknownAddresses(trie *LiveTrie, number int) ([]common.Address, error) {
+	res := make([]common.Address, 0, number)
+	for i := 0; i < number; i++ {
```
HerbertJordan, line 214:
```suggestion
	for len(res) < number {
```

go/database/mpt/verification_proof.go
```diff
@@ -0,0 +1,359 @@
...
+
+		res = append(res, addr)
+	}
+
+	return res, nil
+}
+
+// generateUnknownKeys generates a slice of keys  that does not appear in the input MPT.
+// if in an unlikely situation a generated key is in the trie, it is not added in the result list.
+// It means that in an unlikely situation, the result list may contain fewer keys than the input number.
```
HerbertJordan, line 236:
```suggestion
// generateUnknownKeys generates a slice of keys  that do not appear in the input MPT under the given account.
```

go/database/mpt/verification_proof.go
```diff
@@ -0,0 +1,359 @@
...
+		}
+	}
+
+	return nil
+}
+
+// generateUnknownAddresses generates a slice of addresses that does not appear in the input MPT.
+// if in an unlikely situation a generated address is in the trie, it is not added in the result list.
+// It means that in an unlikely situation, the result list may contain fewer addresses than the input number.
+func generateUnknownAddresses(trie *LiveTrie, number int) ([]common.Address, error) {
```
HerbertJordan, line 212: nit: rename to "generateUnusedAddresses"

go/database/mpt/verification_proof.go
```diff
@@ -0,0 +1,359 @@
...
+		res = append(res, addr)
+	}
+
+	return res, nil
+}
+
+// generateUnknownKeys generates a slice of keys  that does not appear in the input MPT.
+// if in an unlikely situation a generated key is in the trie, it is not added in the result list.
+// It means that in an unlikely situation, the result list may contain fewer keys than the input number.
+func generateUnknownKeys(trie *LiveTrie, number int, address common.Address) ([]common.Key, error) {
```
HerbertJordan, line 237: nit: rename to "generateUnusedKeys"

go/database/mpt/verification_proof.go
```diff
@@ -0,0 +1,359 @@
...
+	res := make([]common.Key, 0, number)
+	for i := 0; i < number; i++ {
```
HerbertJordan, line 239:
```suggestion
	for len(res) < number {
```

go/database/mpt/verification_proof.go
```diff
@@ -0,0 +1,359 @@
...
+		if err := verifyAccountProof(rootHash, proof, addr, AccountInfo{}); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// verifyProofEmptyStorage verifies the consistency of witness proofs for empty storage that are not present in the trie.
+func verifyProofEmptyStorage(trie *LiveTrie, rootHash common.Hash, addr common.Address) error {
```
HerbertJordan, line 138: "Empty Storage" is a property of an account; in particular, this is the case if the storage root of an account is the empty node.

I think this function should be renamed to "verifyProofsForUnusedStorageSlots"

HerbertJordan (dismissed):
This one looks way better now, thanks for improving it!

Sorry that the review took so long.

go/database/mpt/tool/verify_proof.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	// Start profiling ...
+	cpuProfileFileName := context.String(cpuProfileFlag.Name)
+	if strings.TrimSpace(cpuProfileFileName) != "" {
+		if err := startCpuProfiler(cpuProfileFileName); err != nil {
+			return err
+		}
+		defer stopCpuProfiler()
+	}
```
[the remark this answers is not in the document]
kjezek: ok, done

go/database/mpt/verification_proof.go
```diff
@@ -0,0 +1,359 @@
...
+	"github.com/Fantom-foundation/Carmen/go/common/witness"
+	"golang.org/x/exp/maps"
+	"math/rand"
+)
+
+// VerifyProofArchiveTrie verifies the consistency of witness proofs for an archive trie.
+// It reads the trie for each block within the input range.
+// It gets account and storage slots and extracts a witness proof for each account and its storage.
+// It is checked that values in the database and the proof match.
+func VerifyProofArchiveTrie(ctx context.Context, dir string, config MptConfig, from, to int, observer VerificationObserver) error {
```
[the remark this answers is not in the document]
kjezek: ok, done

go/database/mpt/verification_proof.go
kjezek: ok, done

go/database/mpt/verification_proof.go
kjezek: ok, done

go/database/mpt/verification_proof.go
kjezek: ok, done

petr-hanzl (approved):

go/database/mpt/verification_proof.go
kjezek: I thought about it a bit more, and I could create an interface that contains selected methods from LiveTrie and ArchvieTrie, I a few would have to be made public, and the code can be moved, accepting the interface instead of the tries. This would simplify tests I think as I could more easily mock this new interface.

kjezek:
I reworked the code, and put it in its own package. It needed to re-work almost all tests, but I think they are much simpler at the moment.

kjezek:
I run experiment to verify witness proof for **8M** blocks - server with local SSD
It took: **2406** minutes - i.e **1.6** days

HerbertJordan:
> I run experiment to verify witness proof for 8M blocks - server with local SSD

For all blocks from 0 to 8M or just for block 8M?

go/database/mpt/proof/verification.go
```diff
@@ -0,0 +1,465 @@
...
+
+// verifyEmptyAccount verifies the consistency of witness proofs for empty accounts that are not present in the trie.
+func verifyEmptyAccount(trie verifiableTrie, rootHash common.Hash, numAddresses int, observer mpt.VerificationObserver) error {
+	observer.Progress(fmt.Sprintf("Veryfing %d empty addresses...", numAddresses))
+	addresses, err := generateUnusedAddresses(trie, numAddresses)
+	if err != nil {
+		return err
+	}
+	for _, addr := range addresses {
+		proof, err := trie.CreateWitnessProof(addr)
```
HerbertJordan, line 124: The `proof` produced here seems to be never checked for being consistent using its [`IsValid()`](https://github.com/Fantom-foundation/Carmen/blob/3fc9e177f6bb15f517db30573f9b50266706de04/go/common/witness/proof.go#L36) member function. Since the aim of the proof tool is mainly to make sure that the extracted proofs are correct, I think this should be added -- likely also in other places. Right?

go/database/mpt/proof/verification_test.go
```diff
@@ -0,0 +1,672 @@
...
+
+func (c *countingWhenDoneContext) Done() <-chan struct{} {
+	if c.count == c.threshold { // equality to close only once
+		close(c.done)
+	}
+	c.count++
+	return c.done
+}
+
+// errorInjectingProof is a witness.Proof that can be configured to return an error or wrong values after a certain number of calls.
```
HerbertJordan, line 495: Could the same be done with less code using a mock of [Proof](https://github.com/Fantom-foundation/Carmen/blob/3fc9e177f6bb15f517db30573f9b50266706de04/go/common/witness/proof.go#L25)?

go/database/mpt/proof/verification_test.go
```diff
@@ -0,0 +1,672 @@
...
+	if p.count >= p.threshold {
+		return common.Nonce{}, p.returnComplete, p.returnError
+	}
+	p.count++
+	*p.transientCount++
+	return p.Proof.GetNonce(root, address)
+}
+
+// errorInjectingArchiveVerifiableTrie is a verifiableArchiveTrie that can be configured to return an error after a certain number of calls.
+type errorInjectingArchiveVerifiableTrie struct {
```
HerbertJordan, line 542: Also for this type, what is missing in the mock library that makes a hand-crafted mock necessary?

HerbertJordan (changes requested):
Looks good, thanks!

One request though: I think `Proof.IsValid()` should be called for all extracted proofs in addition to checking the contained data. Wdyt?

kjezek:
> > I run experiment to verify witness proof for 8M blocks - server with local SSD
> 
> For all blocks from 0 to 8M or just for block 8M?

Actually it was live db for the recent blockchain:  beyond `80M`, `171400000` accounts, I wrote it wrong in the original post

go/database/mpt/proof/verification.go
kjezek: ok, added

go/database/mpt/proof/verification_test.go
kjezek: yes, switched to use mocks. Not sure the result is very much better, but I think it is ok

go/database/mpt/proof/verification_test.go
kjezek: yes, switched to use mocks. Not sure the result is very much better, but I think it is ok

go/database/mpt/proof/verification_test.go
```diff
@@ -270,61 +332,239 @@ func TestVerification_FailingLiveTrie(t *testing.T) {
...
+			var count int
+			threshold := 1000
+			errorInjectingProofMock := witness.NewMockProof(ctrl)
+			errorInjectingProofMock.EXPECT().GetBalance(gomock.Any(), gomock.Any()).DoAndReturn(func(root common.Hash, addr common.Address) (amount.Amount, bool, error) {
+				if count >= threshold {
+					return amount.Amount{}, test.returnComplete, nil
+				}
+				count++
+				return amount.New(1), true, nil
+			}).AnyTimes()
```
HerbertJordan, line 525: I would have thought something like
```
gomock.InOrder(
   errorInjectingProofMock.EXPECT().GetBalance(...).Times(theshold).Return(amount.New(1), true, nil),
   errorInjectingProofMock.EXPECT().GetBalance(...).AnyTimes().Return(amount.Amount{}, test.returnComplete, nil),
)
```
would have done the trick here. But this depends on whether for this test here it is essential that all expectations share a common counter.

HerbertJordan (approved):

go/database/mpt/proof/verification_test.go
kjezek: do I understand this idea correctly that I would need to manually check `threshold` for each function, and set it accordingly for each method?

HerbertJordan (approved):

Issue #948, opened 2024-07-09
Background Node Flusher is causing random MPT corruption


HerbertJordan:
The background node flusher introduced in #936 is seems to be causing a sporadic, random corruption ion the MPT. The cause is yet unknown, but in stress-test it could be demonstrated that the issue can be reliably reproduced when the node flusher is enabled, and the issue is not present when the flusher is disabled.

After first experimentation, it looks like the likelihood of this problem to show up seems to influenced by:
 - the node cache size: the larger the cache, the less likely the error
 - the flusher frequency: the more the flusher is idle the less frequent the error was observed; disabling the flusher seems to fully eliminate the problem

HerbertJordan:
After two days of debugging the following sequence of steps has been identified to cause this issue:
- a node `N1` with ID `I` is created and added to the cache
- the node `N1` is no longer used and released; but the node stays in the cache
- after some time, `N1` gets evicted from the cache; if it is dirty or currently locked, it gets added to the write buffer [here](https://github.com/Fantom-foundation/Carmen/blob/102aeea2a9cefe434f4832324f2f7c7aeae719d0/go/database/mpt/forest.go#L875)
- a new node `N2` is created for ID `I` and added to the cache
- the new node `N2` gets flushed by the background flusher and is marked as clean
- after some time `N2` gets evicted from the node cache; since it is clean, it is not forwarded to the write buffer
- when the node `I` is to be loaded, the following steps happen:
   - there is no node with ID `I` in the cache (checked [here](https://github.com/Fantom-foundation/Carmen/blob/102aeea2a9cefe434f4832324f2f7c7aeae719d0/go/database/mpt/forest.go#L660))
   - node `N1` associated to ID `I` is still in the write buffer; => this node gets recovered [here](https://github.com/Fantom-foundation/Carmen/blob/102aeea2a9cefe434f4832324f2f7c7aeae719d0/go/database/mpt/forest.go#L680)

Thus, the out-dated reincarnation of `N1` is reinstated, replacing the more recent reincarnation of `N2`. This is corrupting the MPT, leading to the observed failures.

PR #950, opened 2024-07-10
Make flush interval configurable

HerbertJordan:
This PR adds add support for configuring the flush period of the background flusher when creating an MPT State instance.

This is required in order to implement a stress test utility running the database at a higher flush rate than its default rate.

This PR contributes to the investigation of issue #948

**Note for reviewers:**
The main change of this PR is that the former stand-alone `cacheCapacity` parameter is wrapped into a `NodeCacheConfig` struct, extended by the `backgroundFlushPeriod` parameter and requested as a configuration struct in the factory interface. All other changes in the PR are mere adaptations to this interface signature change.


kjezek (approved):
looks good, thanks

PR #952, opened 2024-07-10
Recover nodes from write buffer

HerbertJordan:
This PR fixes a bug in the forest which can lead to the accidental creation of multiple node instances associated with the same ID. 

For a detailed discussion of the addressed issue see #948.


kjezek (dismissed):
great  that this was found

kjezek (approved):

Issue #953, opened 2024-07-10
Write Buffer encounters dirty hashes in stress test


HerbertJordan:
After applying #952 Carmen's new MPT stress test encounters a new issue. Nodes with dirty hashes reach the write buffer, as seen in the following failure message:
```
panic: unable to store branch node with dirty hash

goroutine 14 [running]:
github.com/Fantom-foundation/Carmen/go/database/mpt.BranchNodeEncoderWithChildHashes.Store({}, {0xc00570a280?, 0xaa3a80?, 0x262?}, 0xc00570e408?)
        /home/herbert/coding/carmen/go/database/mpt/nodes.go:2163 +0x209
github.com/Fantom-foundation/Carmen/go/backend/stock/file.(*fileStock[...]).Set(_, _, {{{0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...}, ...}, ...})
        /home/herbert/coding/carmen/go/backend/stock/file/file.go:274 +0x13d
github.com/Fantom-foundation/Carmen/go/backend/stock/synced.(*syncedStock[...]).Set(_, _, {{{0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...}, ...}, ...})
        /home/herbert/coding/carmen/go/backend/stock/synced/synced.go:49 +0xe2
github.com/Fantom-foundation/Carmen/go/database/mpt.(*Forest).flushNode(0xc00014c5a0?, 0x1bdee?, {0xc98410?, 0xc003632400?})
        /home/herbert/coding/carmen/go/database/mpt/forest.go:925 +0x1e5
github.com/Fantom-foundation/Carmen/go/database/mpt.writeBufferSink.Write({0xae3480?}, 0xc00011e600?, {{0xc0033dc140}})
        /home/herbert/coding/carmen/go/database/mpt/forest.go:1070 +0x25
github.com/Fantom-foundation/Carmen/go/database/mpt.(*writeBuffer).emptyBuffer(0xc000160000)
        /home/herbert/coding/carmen/go/database/mpt/write_buffer.go:208 +0x354
github.com/Fantom-foundation/Carmen/go/database/mpt.makeWriteBuffer.func1()
        /home/herbert/coding/carmen/go/database/mpt/write_buffer.go:104 +0xe5
created by github.com/Fantom-foundation/Carmen/go/database/mpt.makeWriteBuffer in goroutine 1
[...]

PR #954, opened 2024-07-10
Avoid nodes to be recovered from write buffer before being written

HerbertJordan:
This PR fixes a synchronization issue in the write buffer discovered by issue #953.

Before this change, the following sequence of events would have been possible (A and B are different goroutines):
- (A) a node in the write buffer is selected for being flushed
- (B) the node is restored, write protection is acquired, the node modified, and write protection abandoned
- (A) write protection on the node is acquired and the node gets flushed

In the last step, the node may be in a state where hashes are invalid. In such a situation, the node can not be flushed.

This change prevents the node from being restored between its selection for being flushed and the acquisition of write permissions.


kjezek (approved):
good catch, thx

PR #956, opened 2024-07-11
removes obsolete go/Jenkinsfile + added stress test to MPT

kjezek:
This PR consolidates Jenkinsfile scripts


Jenkinsfile
```diff
@@ -55,6 +55,12 @@ pipeline {
             }
         }
 
+        stage('Run Mpt Go Stress Test') {
+            steps {
+                sh 'cd go && go run ./database/mpt/tool stress-test --num-blocks 2000'
+            }
+        }
+
```
HerbertJordan, line 63: Since we are already here, maybe we should also move the C++ code format check up. It is not very user friendly to run through all the Go tests, waiting 30+ minutes, to get then a failed test because some formatting is off.

HerbertJordan (approved):

Jenkinsfile
kjezek: ok, good point

PR #958, opened 2024-07-15
allows for cancelling witness proof creation

kjezek:
This PR allows for the witness proof creation to be cancelled via a context. 

It is useful when running verification of the witness proof, so far this step could not be interrupted. 
In the future this may be also useful for instance when a proof is generated via an RPC request and this request is interrupted, the proof generation can be interrupted as well  - though more adaptaions would be needed to reach this functionality.


kjezek:
I think this will not be merged until it is really required. For the verification tool, the proofs will be created by batches of keys

PR #962, opened 2024-07-17
Add explicit tests verifying that non-released StateDB instances are implicitly released by the GC

HerbertJordan:
This PR adds a test automatically testing and verifying a property that has been manually tested before.

### DIFF
--- a/go/state/gostate/go_state_test.go
+++ b/go/state/gostate/go_state_test.go
@@ -14,6 +14,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"runtime"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


kjezek (changes requested):
I somehow cannot see why this is tested. I think forgetting to call `db.Release()` is a mistake. If it is not called, and the `db` is set `nil`, it is released, but the instance of `db` = `nonCommittableStateDB` contains `*stateDB`, and I guess this instance is still not released, because it is held by the `Pool`. Hence, I am not really sure what is tested here.

kjezek (approved):
as discussed in person, I did not fully understand this test at the beginning. It is clarified

HerbertJordan:
Closed since this test is not achieving what it is intended to do.

PR #966, opened 2024-07-19
Allow export LiveDB genesis from Archive.

petr-hanzl:
This PR allows exporting `LiveDB Genesis` from `Archive`. 
It also removes returned error from `GetCodes()` func which was always nil.

To use this simply use `--block` with block > 0 and pass archive as first arg.


go/database/mpt/io/live_test.go
```diff
@@ -287,3 +282,64 @@ func TestCheckEmptyDirectory_FailsIfDirectoryContainsADirectory(t *testing.T) {
...
+
+func TestIO_ExportLiveFromArchive(t *testing.T) {
+	// Create a small Archive from which we export LiveDB genesis.
+	sourceDir := t.TempDir()
+	source, err := mpt.OpenArchiveTrie(sourceDir, mpt.S5ArchiveConfig, mpt.NodeCacheConfig{Capacity: 1024})
+	if err != nil {
+		t.Fatalf("failed to create archive: %v", err)
+	}
+	_ = fillTestBlocksIntoArchive(t, source)
+	exportedBlockHeight := uint64(2)
```
kjezek, line 294: should be `const`

go/database/mpt/io/live_test.go
```diff
@@ -287,3 +282,64 @@ func TestCheckEmptyDirectory_FailsIfDirectoryContainsADirectory(t *testing.T) {
...
+		t.Fatalf("verification of imported DB failed: %v", err)
+	}
+
+	db, err := mpt.OpenGoFileState(targetDir, mpt.S5LiveConfig, mpt.NodeCacheConfig{Capacity: 1024})
+	if err != nil {
+		t.Fatalf("failed to open recovered DB: %v", err)
+	}
+	defer db.Close()
+
+	exists, err := db.Exists(common.Address{1})
```
kjezek, line 326: even if there are only two addresses, I would prefer to do this in a loop, the final code would be more compact

go/database/mpt/tool/export.go
```diff
@@ -52,15 +53,25 @@ func doExport(context *cli.Context) error {
...
 	// check the type of target database
 	mptInfo, err := io.CheckMptDirectoryAndGetInfo(dir)
 	if err != nil {
 		return err
 	}
 
-	export := io.Export
-	if mptInfo.Mode == mpt.Immutable {
-		export = io.ExportArchive
+	hasGivenBlock := blkNumber > 0
```
kjezek, line 64: what if I want to export block "0"

go/database/mpt/tool/export.go
```diff
@@ -52,15 +53,25 @@ func doExport(context *cli.Context) error {
...
+
+	export := io.ExportArchive
+	if hasGivenBlock && mptInfo.Mode == mpt.Immutable {
```
kjezek, line 67: I think more readable would be just to extend the original condition 

```
	export := io.Export
	if mptInfo.Mode == mpt.Immutable {
                if context.IsSet(targetBlockFlag.Name) {
                        export = io.ExportFromArchive      
                } else {
		      export = io.ExportArchive
                }
	}

```

go/database/mpt/io/live.go
```diff
@@ -74,6 +74,31 @@ func Export(ctx context.Context, directory string, out io.Writer) error {
 	}
 	defer db.Close()
 
+	return exportLive(ctx, db, out)
+}
+
+func ExportFromArchive(ctx context.Context, directory string, out io.Writer, block uint64) error {
```
kjezek, line 80: I think names `ExportArchive` and `ExportFromArchive` are too similar, I think this new method should be named perhaps  `ExportBlock`, `ExportBlockHeight`, `ExportBlockFromArchive`, etc.

kjezek (changes requested):
Thanks for implementing this much needed feature. 

I put some comments next to the code. 

Furthermore, my main concern is introduction of the new `exportable` interface.  I think it is not making the code more cleaner and easy to understand because: 
1. it does not serve properly the main purpose of interfaces - polymorphism: 2 out of 3 implementing classes do not use the parameter `block` which makes it not clear what the classes have in common and why they share the same interface if they differ in parameters. 
2. there is already a level of abstraction of live and archive db (perhaps even too extensive at the moment), while adding yet another, one purpose, layer makes it even more complicated to comprehend. 

I think the export could be much simpler implemented by opening the Archive at the requested block, instead of the head state as it is supported now, and just export this view of the Archive. In particular implement a method similar to this:
```

func OpenArchiveState(directory string, block uint64, config MptConfig, cacheConfig NodeCacheConfig) (*MptState, error) {
	lock, err := openStateDirectory(directory)
	if err != nil {
		return nil, err
	}
	trie, err := OpenArchiveTrie(directory, config, cacheConfig)
	if err != nil {
		return nil, err
	}
	view, err := trie.getView(block)
	if err != nil {
		return nil, errors.Join(err, trie.Close(), lock.Release())
	}
	return newMptState(directory, lock, view)
}

```

then I believe that the resulting `*MptState` can be used for the export the same way as if it was a `LiveDB`, i.e. basically the same code after this line: https://github.com/Fantom-foundation/Carmen/blob/main/go/database/mpt/io/live.go#L78

go/database/mpt/archive_trie.go
```diff
@@ -328,6 +328,17 @@ func (a *ArchiveTrie) Flush() error {
 	)
 }
 
+func (a *ArchiveTrie) VisitTrie(visitor NodeVisitor, block uint64) error {
```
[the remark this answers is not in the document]
kjezek: Can you also make the `block` parameter the first one in the method header - it will be consistent with other methods.

go/database/mpt/archive_trie.go
```diff
@@ -205,7 +205,7 @@ func (a *ArchiveTrie) GetCode(block uint64, account common.Address) (code []byte
 	return a.head.GetCodeForHash(info.CodeHash), nil
 }
 
-func (a *ArchiveTrie) GetCodes() (map[common.Hash][]byte, error) {
+func (a *ArchiveTrie) GetCodes() map[common.Hash][]byte {
```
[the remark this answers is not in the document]
petr-hanzl: The error actually cannot be created this and it easifies the code inside `immutable` and. I would like to include this but I can create first PR with this change, merge it into main and then use it here if you feel like its necessary...

go/database/mpt/io/live_test.go
petr-hanzl: Correct me if Im wrong but `address(1)` should be inside and `address(2)` should not be inside meaning the loop does not make sense.

go/database/mpt/archive_trie.go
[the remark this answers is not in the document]
kjezek: ok, lets keep this change, since this is not an API method I think it does not matter much if it is changed or not. I just wanted to stress that it is always better to have only related changes in PRs.

go/database/mpt/io/live.go
petr-hanzl: Im not really a fan of these names but I dont see a better alternative. So lets use these I guess.

petr-hanzl: How about `ExportHead` instead of `ExportBlock`, then keep `ExportArchive` and use `ExportBlockFromArchive`?

go/database/mpt/io/live_test.go
kjezek: got it now. I did not fully understand this test, because it relies on data inserted in `fillTestBlocksIntoArchive`. I would suggest to create and test data in this test - have a loop that inserts N blocks, a few addresses each, export a block, and check that all addresses except the exported ones are not present in the newly imported db, and the imported ones are present.  It would be more visible what is happening. 

Other tests using data from `fillTestBlocksIntoArchive` do not directly test addresses created by this method. But since this method does, it would be more readable to create and test the data one next to another.

go/database/mpt/io/archive.go
```diff
@@ -83,7 +83,7 @@ func ExportArchive(ctx context.Context, directory string, out io.Writer) error {
 	}
 
 	// Write out codes.
-	codes, err := archive.GetCodes()
+	codes := archive.GetCodes()
 	if err != nil {
 		return fmt.Errorf("failed to retrieve codes: %v", err)
 	}
```
kjezek, line 89: I guess this test is not needed anymore

go/database/mpt/archive_trie.go
```diff
@@ -328,6 +328,19 @@ func (a *ArchiveTrie) Flush() error {
...
+	a.rootsMutex.Lock()
+	if uint64(a.roots.length()) < block {
+		a.rootsMutex.Unlock()
+		return fmt.Errorf("block %d is not present", block)
+
+	}
+
+	ref := a.roots.get(block).NodeRef
+	a.rootsMutex.Unlock()
+	return a.forest.VisitTrie(&ref, visitor)
```
kjezek, line 341:
```suggestion
	view, err := a.getView(block)
	if err != nil {
		return err
	}
	return view.VisitTrie(visitor)
```

kjezek: getting the view this way is more compact but also already checks for errors, which would be otherwise missing.

go/database/mpt/io/live.go
```diff
@@ -65,7 +65,7 @@ func Export(ctx context.Context, directory string, out io.Writer) error {
 	}
 
 	if info.Config.Name != mpt.S5LiveConfig.Name {
-		return fmt.Errorf("can only support export of LiveDB instances, found %v in directory", info.Mode)
+		return fmt.Errorf("can only support exportLive of LiveDB instances, found %v in directory", info.Mode)
```
kjezek, line 68:
```suggestion
		return fmt.Errorf("can only support export of LiveDB instances, found %v in directory", info.Mode)
```

go/database/mpt/tool/export.go
```diff
@@ -60,7 +62,13 @@ func doExport(context *cli.Context) error {
 
 	export := io.Export
 	if mptInfo.Mode == mpt.Immutable {
-		export = io.ExportArchive
+		if cliCtx.IsSet(targetBlockFlag.Name) {
+			blkNumber := cliCtx.Uint64(targetBlockFlag.Name)
+			cliCtx.Context = context.WithValue(cliCtx.Context, "block", blkNumber)
```
kjezek, line 67: I think great idea to pass the parameter this way, much better then before

go/database/mpt/io/live.go
kjezek: not sure what would be called `ExportHead` ?  Anyway: `ExportFromArchive -> ExportBlockFromArchive` sounds good to me.

go/database/mpt/io/exportable.go
```diff
@@ -0,0 +1,46 @@
...
+// this software will be governed by the GNU Lesser General Public License v3.
+
+package io
+
+import (
+	"github.com/Fantom-foundation/Carmen/go/common"
+	"github.com/Fantom-foundation/Carmen/go/database/mpt"
+)
+
+type exportable interface {
```
kjezek, line 18: I do not think this interface name is very well fitting. Having a look at the interface name and its method, what suggests  the methods` Visit() `+ getters compose something exportable? Exportable interface I would normally expect methods like `export`, `convert`, `send`, etc. 

I would say this interface - judging from its methods - is `TrieVisitor`, or even more generic `StateVisitor`, or back more specific `MptStateVisitor`. Wdyf?

kjezek: Also the interface should be documented

go/database/mpt/io/live.go
```diff
@@ -74,6 +74,32 @@ func Export(ctx context.Context, directory string, out io.Writer) error {
 	}
 	defer db.Close()
 
+	return exportLive(ctx, db, out)
+}
+
+func ExportFromArchive(ctx context.Context, directory string, out io.Writer) error {
```
kjezek, line 80: did not we discussed before renaming this method?

go/database/mpt/io/live.go
```diff
@@ -74,6 +74,32 @@ func Export(ctx context.Context, directory string, out io.Writer) error {
...
+	}
+
+	archive, err := mpt.OpenArchiveTrie(directory, info.Config, mpt.NodeCacheConfig{})
+	if err != nil {
+		return err
+	}
+
+	defer archive.Close()
+
+	block := ctx.Value("block")
```
kjezek, line 97: I thought at the beginning that it is a good idea to pass this parameter via context, but seeing the final version, I do not see any reason now, because this method can just have a parameter `block` in its header right?  I did not like the unification with other methods, where `block` was not used, but I cannot see why only this method could not have this parameter. 

If you want to have the same header with other methods to use method reference in `export.go,` yo can always wrap it there in another method. 

```
m := func(ctx context.Context, directory string) error {
    return ExportFromArchive(ctx, directory, out, block)
} 
```

go/database/mpt/io/live_test.go
```diff
@@ -287,3 +282,65 @@ func TestCheckEmptyDirectory_FailsIfDirectoryContainsADirectory(t *testing.T) {
 		t.Errorf("unexpected error: %v", err)
 	}
 }
+
+func TestIO_ExportLiveFromArchive(t *testing.T) {
```
kjezek, line 286: `...ExportLive...` is a bit misleading, because you export one state from archive, the live db is the always changing head state, not an already sealed state in the archive

go/database/mpt/io/live_test.go
```diff
@@ -287,3 +282,65 @@ func TestCheckEmptyDirectory_FailsIfDirectoryContainsADirectory(t *testing.T) {
...
+
+	exists, err = db.Exists(common.Address{2})
+	if err != nil {
+		t.Fatalf("cannot find whether account exists: %v", err)
+	}
+	if exists {
+		t.Fatal("restored DB contains account 2 added after exported block")
+	}
+
+	if got, err := db.GetHash(); err != nil || got != hash {
```
kjezek, line 343: would be better to test `err` in its own condition.

go/database/mpt/tool/export.go
kjezek: not think actually this at the moment, see the other comment

kjezek (changes requested):
Thank for more effort invested. 

I think generally it looks good, the structure is good. 
There are still unresolved comments and ideas from previous round. Please address them as well. 

I think the test for export should be more comprehensive - it is better to create M blocks with N addresses and storage slots each, and then export (perhaps even block by block) end check that addresses are correctly present or not-present in the recovered live db. 

Relying the test only on two addresses is too shallow, it can miss some corner cases etc. The test should cover situations like some addresses were new in a block, some were deleted and some were updated. Similarly for storage keys and codes.

go/database/mpt/io/live.go
petr-hanzl: `ExportHead` would be `ExportLive` - exporting live genesis from `LiveDB`

go/database/mpt/io/live.go
petr-hanzl: I dont really like your the func `m` you suggested. I will just call the functions inside a switch and then pass the error to the `Join`

kjezek: yes sure, it is fine. I was suggesting this solution because I thought you wanted to have all three methods with the same header (otherwise I do not understand why the `block` param was added to other methods), but sure calling all three methods independently in a switch statement is the most straightforward and readable approach.

go/database/mpt/io/live.go
kjezek: no, dont change this, it is unrelated to this PR, lets stay just with adding the method `ExportBlockFromArchive`

petr-hanzl:
@kjezek 
> I think the test for export should be more comprehensive - it is better to create M blocks with N addresses and storage slots each, and then export (perhaps even block by block) end check that addresses are correctly present or not-present in the recovered live db.
> 
> Relying the test only on two addresses is too shallow, it can miss some corner cases etc. The test should cover situations like some addresses were new in a block, some were deleted and some were updated. Similarly for storage keys and codes.

In that case the tests for all of the export functions should be rewritten to this style. They always use couple accounts and check them for existence altogether with the hash.

go/database/mpt/io/live.go
petr-hanzl: ic. I thought you wanted to rename all of the `Export...` funcs. Understood now.

kjezek:
> @kjezek
> 
> > I think the test for export should be more comprehensive - it is better to create M blocks with N addresses and storage slots each, and then export (perhaps even block by block) end check that addresses are correctly present or not-present in the recovered live db.
> > Relying the test only on two addresses is too shallow, it can miss some corner cases etc. The test should cover situations like some addresses were new in a block, some were deleted and some were updated. Similarly for storage keys and codes.
> 
> In that case the tests for all of the export functions should be rewritten to this style. They always use couple accounts and check them for existence altogether with the hash.

Why this test cannot be extended according to review without touching other tests?

kjezek (changes requested):
Thanks, I think we are almost done, though the tests need a bit more work.

1. test of `VisitTrie` tests that the visitor may be called but also does not have to be called, it will need to test that correct data are passed through
2.  test of export - please see my previous comment. I think testing only two addresses - one is present while the other one is not, is not enough robust,

go/database/mpt/archive_trie_test.go
```diff
@@ -1690,6 +1684,41 @@ func TestArchiveTrie_FailingLiveStateUpdate_InvalidatesArchive(t *testing.T) {
 	}
 }
 
+func TestArchiveTrie_VisitTrie(t *testing.T) {
+	for _, config := range allMptConfigs {
+		ctrl := gomock.NewController(t)
+		nodeVisitor := NewMockNodeVisitor(ctrl)
+		nodeVisitor.EXPECT().Visit(gomock.Any(), gomock.Any()).AnyTimes()
```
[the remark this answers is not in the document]
petr-hanzl: I thought `AnyTime()` means atleast once. Fixed.

Anyhow I believe testing whether correct data is passed would mean that the `Forest` is tested instead of the `ArchiveTrie`. I think only call to the `Visitor` should be tested, what data is being passed should be concern of `forest_test.go`

https://github.com/Fantom-foundation/Carmen/blob/deaa6b416d8c69c1ba9ef99a66f35165399f0a12/go/database/mpt/forest.go#L454

go/database/mpt/archive_trie_test.go
```diff
@@ -1690,6 +1684,42 @@ func TestArchiveTrie_FailingLiveStateUpdate_InvalidatesArchive(t *testing.T) {
 	}
 }
 
+func TestArchiveTrie_VisitTrie(t *testing.T) {
+	for _, config := range allMptConfigs {
+
+		t.Run(config.Name, func(t *testing.T) {
+			ctrl := gomock.NewController(t)
+			nodeVisitor := NewMockNodeVisitor(ctrl)
+			nodeVisitor.EXPECT().Visit(gomock.Any(), gomock.Any()).MinTimes(1)
```
kjezek, line 1693: as I wrote previously, it should also test here that correct addresses and keys were propagated to the visitor.

also it should test that for block 0 or 1 the visitor is not called at all

go/database/mpt/io/live.go
```diff
@@ -74,6 +74,32 @@ func Export(ctx context.Context, directory string, out io.Writer) error {
...
+// ExportBlockFromArchive exports LiveDB genesis for a single given block from an Archive.
+// Note: block must be <= of Archive block height.
+func ExportBlockFromArchive(ctx context.Context, directory string, out io.Writer, block uint64) error {
+	info, err := CheckMptDirectoryAndGetInfo(directory)
+	if err != nil {
+		return fmt.Errorf("error in input directory: %v", err)
+	}
+
+	if info.Config.Name != mpt.S5ArchiveConfig.Name {
+		return fmt.Errorf("can only support exportLive of S5 Archive instances, found %v in directory", info.Config.Name)
```
kjezek, line 89:
```suggestion
		return fmt.Errorf("can only support export of S5 Archive instances, found %v in directory", info.Config.Name)
```

go/database/mpt/io/live_test.go
```diff
@@ -287,3 +282,100 @@ func TestCheckEmptyDirectory_FailsIfDirectoryContainsADirectory(t *testing.T) {
...
+			newAddr := common.AddressFromNumber(i + j)
+
+			update.CreatedAccounts = append(update.CreatedAccounts, newAddr)
+			update.Balances = append(update.Balances, common.BalanceUpdate{Account: newAddr, Balance: amount.New(u)})
+			update.Nonces = append(update.Nonces, common.NonceUpdate{Account: newAddr, Nonce: common.ToNonce(u)})
+			update.Codes = append(update.Codes, common.CodeUpdate{Account: newAddr, Code: code})
+			update.Slots = append(update.Slots, common.SlotUpdate{Account: newAddr, Key: common.Key{1}, Value: common.Value{1}})
+		}
+
+		if i%2 != 0 {
```
kjezek, line 317: I somehow cannot fully understand this algorithm.  If I am not wrong the code above re-creates two addresses and creates a new one in each block:

```
block0:  addr0, adder, addr2 
blcok1: addr0, addr1, addr2, address

```
in `block1` - `addr1` and `addr2` were re-created (old keys removed, created new ones, but with the same values, i.e. no-one can really see the difference.  `addr3` is new.

Next you either update or delete address which matches current block. In this case `addr1` - if it gets updated, it is set with the same values, right? I.e. no one can really see the difference.  If it gets deleted, it has some merit - a real change is happening.  

Maybe I could not fully understand the algorithm, but all in all I think a much simpler version would do the job:

In each loop `i`,  keep creating still the same address `Address{j}`, but for account properties and key values set always `i` or `i+1`.  This way, you will have a new version of the same accounts in each block. 

You can have a condition that one of these three addresses is deleted in each block, not created. 

that it is

go/database/mpt/io/mpt_state_visitor.go
```diff
@@ -0,0 +1,47 @@
...
+// this software will be governed by the GNU Lesser General Public License v3.
+
+package io
+
+import (
+	"github.com/Fantom-foundation/Carmen/go/common"
+	"github.com/Fantom-foundation/Carmen/go/database/mpt"
+)
+
+// mptStateVisitor is a wrapper around Trie which determines functions necessary to export a LiveDB genesis.
```
kjezek, line 18: this is not a correct documentation, saying that  
1. is a wrapper around Trie 
2. is necessary for export 

is oriented to one implementation of this interface. But the interface is generic and the documentation should be generic as well.

... mptStateVisitor is an interface for Tries that allows for visiting the Trie nodes and furthermore getting its properties such as a root hash and contract codes.  ...

kjezek (changes requested):
Thanks for the update. I think it is much better, but the two tests need a bit of more effort. Please have a look at my comments.

go/database/mpt/archive_trie_test.go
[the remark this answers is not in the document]
kjezek: no, you test as a black-box. You insert accounts and keys, and you test that the visitor returns you nodes representing these accounts and keys. You do not assume anything about forest and layers underneath.

go/database/mpt/io/live_test.go
petr-hanzl: Im fully aware what it does. I just wanted to make the test a bit random and I think it serves its purpose. Do we still want to simplify it...?

kjezek: If you confirm that I understand the algorithm right, then how is it "a bit random" if two third of changes actually do  a "change" in a way that the result is still the same? 

Yes, I think the algorithm I propose is much more straightforward and performs actually more changes than the original one. I prefer it.

HerbertJordan (dismissed):
Looks good, thanks.

We should integrate this new export capability into some Jenkins test pipeline as a follow-up. It would be nice to check that the data exported from a LiveDB is equivalent to the data extracted from an Archive. To do so, an MD5 checksum of the respective files could be compared.

go/database/mpt/archive_trie_test.go
```diff
@@ -1690,6 +1684,107 @@ func TestArchiveTrie_FailingLiveStateUpdate_InvalidatesArchive(t *testing.T) {
...
+			}
+			found = false
+
+			// reset visitor
+			nodeVisitor = NewMockNodeVisitor(ctrl)
+			nodeVisitor.EXPECT().Visit(gomock.Any(), gomock.Any()).Do(func(node Node, info NodeInfo) {
+				switch node.(type) {
+				case *AccountNode:
+					found = true
+					n := node.(*AccountNode)
```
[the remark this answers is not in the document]
petr-hanzl: Thats pretty useful! I appreciate it. Thanks!

go/database/mpt/archive_trie_test.go
```diff
@@ -1690,6 +1684,105 @@ func TestArchiveTrie_FailingLiveStateUpdate_InvalidatesArchive(t *testing.T) {
 	}
 }
 
+func TestArchiveTrie_VisitTrie_CorrectDataIsVisited(t *testing.T) {
+	for _, config := range allMptConfigs {
+		t.Run(config.Name, func(t *testing.T) {
+			ctrl := gomock.NewController(t)
+
+			archive, err := OpenArchiveTrie(t.TempDir(), config, NodeCacheConfig{Capacity: 1024})
```
kjezek, line 1692: This test could be also implemented using TableDriven testing: https://go.dev/wiki/TableDrivenTests
where the subject of customisation would be `NodeVisitor` because at the moment the test contains two blocks of code that are the same, just the mock is different.

kjezek (approved):
thanks for all this effort. It looks good to me. I put some advice in one of the tests.

go/database/mpt/archive_trie_test.go
petr-hanzl: Good point. I will rework the test.

kjezek (changes requested):
Thanks for this update. The test is done as I meant, but there is one seemingly detail that must be changed: the tests may run in parallel. i.e. the `found` variable cannot be global. It must be wrapped in the "`tests`" struct for instance.

kjezek (approved):
looks good!

PR #967, opened 2024-07-20
Fix conversion of cache size to cache capacity

HerbertJordan:
This PR fixes an issue introduced by #950 related to the definition of very small node cache sizes.

Before #950, if a node cache size of less than a single node was specified, the MPT implementation would increase it to a minimum value of `2_000` nodes.

After #950, if a node cache size of less than a single node is specified, it got converted to a `capacity` definition of `0`, which lead to the utilization of the default cache size of `10_000_000` nodes. This lead to memory issue problems in Sonic tests.

With this fix, the old behavior was recovered.


kjezek (approved):
thanks for debugging and fixing this

PR #971, opened 2024-07-23
adds extraction features of witness proofs

kjezek:
This PR extends export capabilities of the WitnessProof. It was implemented on a request for easier integration to the RPC API.  It was detected that following features are missing for integration to the RPC
* `WitnessProof` contains MPT nodes for the whole proof in one structure, but the RPC returns two sets: MPT nodes for an account and MPT nodes for storage keys. 
* RPC API provides hash of the storage root from an account.

While all the data are encoded already in the `WitnessProof` and could be resolved at the RPC level, it was decided to help with integration by providing new extraction features by Carmen. 

In particular a method to extract sub-proof of a proof was made publicly available, and one method to extract storage nodes + storage root was added. 

As a result, account nodes, storage nodes and the storage root can be obtained by following sequence of method calls: 
```
	if err := db.QueryBlock(0, func(ctxt carmen.HistoricBlockContext) error {
		proof, err := ctxt.GetProof(carmen.Address{1}, carmen.Key{1})
		if err != nil {
			...
		}

                 accountProof, accountComplete := proof.Extract(rootHash, carmen.Address{1})
                 storageElements, storageRoot, storageComplete := proof.GetStorageElements(rootHash, carmen.Address{1}, carmen.Key{1})

		return nil
	}); err != nil {
		...
	}
```

see full example: https://github.com/Fantom-foundation/Carmen/pull/971/files#diff-982625a1f991b05fd70b53dfcf459242ea40d9c3bc73ba2bd7998a5a617fa725R413 

Notes about the design:
* the method `Extract` returns `WitnessProof` and this proof can be further queried by methods of `WitnessProof`, or the nodes can be obtained by calling `proof.GetElements()`
* the method `GetStorageElements` returns the nodes and the hash directly, because such a partial proof cannot be further queried by carmen API - i.e. there is no method in the API that would accept a "storage only" proof and the storage root. This method is purely for export by the RPC.

### DIFF
--- a/go/carmen/example_test.go
+++ b/go/carmen/example_test.go
@@ -410,6 +410,83 @@ func ExampleHistoricBlockContext_GetProof() {
 	//Storage slot value of key 0x0900000000000000000000000000000000000000000000000000000000000000 at block: 9 and address: 0x0900000000000000000000000000000000000000 is 0x0900000000000000000000000000000000000000000000000000000000000000
 }
 
+func ExampleWitnessProof_GetStorageElements() {
+	dir, err := os.MkdirTemp("", "carmen_db_*")
+	if err != nil {
+		log.Fatalf("cannot create temporary directory: %v", err)
+	}
+	db, err := carmen.OpenDatabase(dir, carmen.GetCarmenGoS5WithArchiveConfiguration(), nil)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// ------- Prepare the database -------
+
+	// Add only one block with one address and storage
+	if err := db.AddBlock(0, func(context carmen.HeadBlockContext) error {
+		if err := context.RunTransaction(func(context carmen.TransactionContext) error {
+			context.CreateAccount(carmen.Address{1})
+			context.AddBalance(carmen.Address{1}, carmen.NewAmount(1))
+			context.SetState(carmen.Address{1}, carmen.Key{1}, carmen.Value{1})
+			return nil
+		}); err != nil {
+			log.Fatalf("cannot create transaction: %v", err)
+		}
+		return nil
+	}); err != nil {
+		log.Fatalf("cannot add block: %v", err)
+	}
+
+	// block wait until the archive is in sync
+	if err := db.Flush(); err != nil {
+		log.Fatalf("cannot flush: %v", err)
+	}
+
+	// ------- Query witness proofs from a block -------
+
+	var elements []string
+	// proof each address and key from each block, and merge all in one proof
+	if err := db.QueryBlock(0, func(ctxt carmen.HistoricBlockContext) error {
+		proof, err := ctxt.GetProof(carmen.Address{1}, carmen.Key{1})
+		if err != nil {
+			log.Fatalf("cannot create witness proof: %v", err)
+		}
+
+		// store proof
+		elements = proof.GetElements()
+		return nil
+	}); err != nil {
+		log.Fatalf("cannot query block: %v", err)
+	}
+
+	var rootHash carmen.Hash
+	if err := db.QueryHistoricState(0, func(ctxt carmen.QueryContext) {
+		rootHash = ctxt.GetStateHash()
+	}); err != nil {
+		log.Fatalf("cannot query block: %v", err)
+	}
+
+	// ------- Close the database - no more needed -------
+	if err := db.Close(); err != nil {
+		log.Fatalf("cannot close db: %v", err)
+	}
+	if err := os.RemoveAll(dir); err != nil {
+		log.Fatalf("cannot remove dir: %v", err)
+	}
+
+	// ------- Recover Proof and Split into Account and Storage Proof  -------
+
+	proof := carmen.CreateWitnessProofFromNodes(elements...)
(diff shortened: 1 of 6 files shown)


go/common/witness/proof.go
```diff
@@ -23,13 +23,30 @@ import (
...
 	// IsValid checks that this proof is self-consistent. If the result is true, the proof can be used
 	// for extracting verified information. If false, the proof is corrupted and should be discarded.
 	IsValid() bool
 
 	// GetElements returns serialised elements of the witness proof.
 	GetElements() []string
 
+	// GetStorageElements returns serialised elements of the witness proof for a given account
+	// and selected storage locations from this proof.
+	// The resulting proof contains only a proof for the storage part of the account.
```
HerbertJordan, line 42: nit: the documentation states that the first result value is a proof, but it is a `[]string` instead. Should the result be replaced by a proof?

go/database/mpt/proof_test.go
```diff
@@ -753,12 +753,183 @@ func TestWitnessProof_Extract_EmbeddedNode_In_Proof(t *testing.T) {
...
+	t.Run("Extract storage", func(t *testing.T) {
+		storageElements, storageHash, complete := proof.GetStorageElements(hash, address, key)
+		if !complete {
+			t.Fatalf("proof is not complete")
+		}
+		storageProof := CreateWitnessProofFromNodes(storageElements)
+		if !storageProof.IsValid() {
+			t.Fatalf("storage proof is not valid")
+		}
+
```
HerbertJordan, line 835: Maybe it should be tested that
- the list of `storageElements` is not empty
- one of the `storageElements` has a hash that is equivalent to the `storageHash`

go/carmen/example_test.go
```diff
@@ -410,6 +410,83 @@ func ExampleHistoricBlockContext_GetProof() {
...
+
+	proof := carmen.CreateWitnessProofFromNodes(elements...)
+	accountProof, accountComplete := proof.Extract(rootHash, carmen.Address{1})
+	storageElements, storageRoot, storageComplete := proof.GetStorageElements(rootHash, carmen.Address{1}, carmen.Key{1})
+
+	fmt.Printf("Account proof is complete: %v and has elements: %d elements\n", accountComplete, len(accountProof.GetElements()))
+	fmt.Printf("Storage proof is complete: %v and has elements: %d elements and root: %v\n", storageComplete, len(storageElements), storageRoot)
+
+	// Output: Account proof is complete: true and has elements: 1 elements
+	// Storage proof is complete: true and has elements: 1 elements and root: [46 124 116 253 41 173 162 163 11 134 9 5 150 41 208 193 156 91 227 104 90 168 232 46 92 214 249 73 33 194 217 13]
```
HerbertJordan, line 487:
```suggestion
	fmt.Printf("Account proof is complete: %v and has %d elements\n", accountComplete, len(accountProof.GetElements()))
	fmt.Printf("Storage proof is complete: %v and has %d elements and root: %v\n", storageComplete, len(storageElements), storageRoot)

	// Output: Account proof is complete: true and has 1 elements
	// Storage proof is complete: true and has 1 elements and root: [46 124 116 253 41 173 162 163 11 134 9 5 150 41 208 193 156 91 227 104 90 168 232 46 92 214 249 73 33 194 217 13]
```

HerbertJordan (dismissed):

go/common/witness/proof.go
kjezek: thx, fixed doc. I did not want to return the Proof interface - since it would contain only storage nodes, all getter of the Proof interface would be useless because we do not support proving storage only.

go/carmen/proof.go
```diff
@@ -30,13 +30,30 @@ func CreateWitnessProofFromNodes(elements ...string) WitnessProof {
...
 	// IsValid checks that this proof is self-consistent. If the result is true, the proof can be used
 	// for extracting verified information. If false, the proof is corrupted and should be discarded.
 	IsValid() bool
 
 	// GetElements returns serialised elements of the witness proof.
 	GetElements() []string
 
+	// GetStorageElements returns serialised elements of the witness proof for a given account
+	// and selected storage locations from this proof.
+	// The resulting proof contains only a proof for the storage part of the account.
```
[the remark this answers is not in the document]
kjezek: fixed

go/carmen/proof.go
```diff
@@ -82,10 +99,30 @@ type witnessProof struct {
 	proof witness.Proof
 }
 
+func (w witnessProof) Extract(root Hash, address Address, keys ...Key) (WitnessProof, bool) {
+	commonKeys := make([]common.Key, len(keys))
+	for i, k := range keys {
+		commonKeys[i] = common.Key(k)
+	}
+
+	proof, err := w.proof.Extract(common.Hash(root), common.Address(address), commonKeys...)
```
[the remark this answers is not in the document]
kjezek: oh, right :)

go/database/mpt/proof_test.go
kjezek: good idea, will add

thaarok (approved):
Seems this provides all I need

HerbertJordan (approved):

PR #975, opened 2024-07-29
Make CreateWitnessProof accessible through StateDB interface

thaarok:
The whole RPC layer of Sonic currently use StateDB interface and migration to Carmen facade is not in sight. To avoid the need of passing the unwrapped State reference into the RPC, I want to add `CreateWitnessProof` method into the NonCommittableStateDB interface.

### DIFF
--- a/go/state/state_db.go
+++ b/go/state/state_db.go
@@ -14,6 +14,7 @@ package state
 import (
 	"errors"
 	"fmt"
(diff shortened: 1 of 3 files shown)


HerbertJordan (dismissed):
Looks good, thanks for fixing this. Please merge #976 into this PR before merging this one.

HerbertJordan (approved):

PR #973, opened 2024-07-24
Add checkpoint support to stock implementations

HerbertJordan:
This PR adds checkpointing support to all stock implementations.

While implementing the tests it was realized that the `Restore` operation of a checkpoint can not be a member function of the targeted data structure itself. This would require the ability to open a corrupted data structure to be able to restore it. Supporting this would be very complicated, and it is not needed.

Instead, the `Restore` functionality was moved into its own interface, independent of the data structure it is targeting. This way, a corrupted data structure can be restored before attempting to reopen the data structure.

This is the second step toward supporting checkpointing of archives (see #946)

PR #978, opened 2024-07-30
Add generic utilities for Reading/Writing JSON files

HerbertJordan:
This PR adds two generic utility functions to simplify the reading and writing of small JSON files.

The need for those files was identified in #973 and #977.

### DIFF
--- a/go/backend/utils/json_file_io.go
+++ b/go/backend/utils/json_file_io.go
@@ -1,0 +1,39 @@
+// Copyright (c) 2024 Fantom Foundation
+//
+// Use of this software is governed by the Business Source License included
+// in the LICENSE file and at fantom.foundation/bsl11.
+//
+// Change Date: 2028-4-16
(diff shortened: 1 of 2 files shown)


go/backend/utils/json_file_io.go
```diff
@@ -0,0 +1,29 @@
...
+	var zero T
+	data, err := os.ReadFile(file)
+	if err != nil {
+		return zero, err
+	}
+	var meta T
+	if err := json.Unmarshal(data, &meta); err != nil {
+		return zero, err
+	}
+	return meta, nil
```
petr-hanzl, line 29:
```suggestion
	var t T
	if err := json.Unmarshal(data, &t); err != nil {
		return zero, err
	}
	return t, nil
```

petr-hanzl (approved):
Thanks for splitting the PRs

PR #980, opened 2024-07-31
Add export to public API

petr-hanzl:
This PR adds option to export LiveDB genesis for certain block from Archive while LiveDB is syncing.


go/carmen/carmen.go
```diff
@@ -198,6 +200,11 @@ type HistoricBlockContext interface {
 	// otherwise, the proof is returned.
 	GetProof(address Address, keys ...Key) (WitnessProof, error)
 
+	// CreateLiveDBGenesis writes a LiveDB genesis file for given block into out.
+	// The genesis file is exported from the archive for the
+	// given block and is used to sync LiveDB to that block.
+	CreateLiveDBGenesis(out io.Writer) (Hash, error)
```
HerbertJordan, line 206: I think the term `Genesis` should not be in this function name. While it is true that what gets exported is potentially part of a genesis file, it is only a small part of it, and hardly the defining element.

I would suggest to rename this function to something like`Export()`. Wdyt?

go/backend/archive/archive.go
```diff
@@ -60,6 +65,9 @@ type Archive interface {
 	// CreateWitnessProof creates a witness proof for the given account and keys.
 	CreateWitnessProof(block uint64, address common.Address, keys ...common.Key) (witness.Proof, error)
 
+	// CreateLiveDBGenesis writes a LiveDB genesis for given block into out.
+	CreateLiveDBGenesis(block uint64, out io.Writer) (common.Hash, error)
```
HerbertJordan, line 69: As in the Carmen interface, I would argue that what is exported is not a `Genesis`.

go/carmen/carmen.go
HerbertJordan, line 206: This function will eventually also need a `context.Context` to support canceling the export, right?

go/carmen/example_test.go
```diff
@@ -515,3 +518,82 @@ func ExampleDatabase_GetMemoryFootprint() {
...
+
+	// Export genesis
+	var rootHash carmen.Hash
+	b := bytes.NewBuffer(nil)
+	if err = db.QueryBlock(uint64(1), func(ctxt carmen.HistoricBlockContext) error {
+		rootHash, err = ctxt.CreateLiveDBGenesis(b)
+		if err != nil {
+			log.Fatalf("cannot create genesis: %v", err)
+		}
+		return nil
```
HerbertJordan, line 559:
```suggestion
		return err
```

go/carmen/example_test.go
```diff
@@ -515,3 +518,82 @@ func ExampleDatabase_GetMemoryFootprint() {
...
+	}); err != nil {
+		log.Fatalf("cannot query block: %v", err)
```
HerbertJordan, line 561:
```suggestion
		log.Fatalf("failed to export block: %v", err)
```

go/common/genesis/genesis.go
```diff
@@ -8,7 +8,7 @@
 // On the date above, in accordance with the Business Source License, use of
 // this software will be governed by the GNU Lesser General Public License v3.
 
-package io
+package genesis
```
HerbertJordan, line 11: Not a real fan of the `genesis` term due to its much wider (and to some extend different) meaning in the context of a block chain.

If this package is needed (see other comments) we should think about a better name.

go/common/genesis/genesis.go
HerbertJordan, line 1: Why do those formally private constants, types, and functions have to be made public?

go/common/genesis/genesis.go
```diff
@@ -38,7 +51,7 @@ func writeCodes(codes map[common.Hash][]byte, out io.Writer) error {
 	return nil
 }
 
-func readCode(in io.Reader) ([]byte, error) {
+func ReadCode(in io.Reader) ([]byte, error) {
```
HerbertJordan, line 54: This function used to be a utility function for a very specific purpose. By making it public and putting it in a common package, much more documentation would be required to describe what this function is supposed to do.

go/database/mpt/export.go
HerbertJordan, line 1: There is no need to move this code from [io/live.go](https://github.com/Fantom-foundation/Carmen/blob/main/go/database/mpt/io/live.go#L114) to this location. Keeping it where it was would eliminate the need for the new `genesis` package, which's purpose is not well defined.

go/database/mpt/io/live.go
```diff
@@ -127,49 +83,8 @@ func ExportBlockFromArchive(ctx context.Context, directory string, out io.Writer
...
-	}
-
-	// Write out all accounts and values.
-	visitor := exportVisitor{out: out, ctx: ctx}
-	if err := db.Visit(&visitor); err != nil || visitor.err != nil {
-		return fmt.Errorf("failed exporting content: %w", errors.Join(err, visitor.err))
-	}
-
-	return nil
+	_, err = mpt.ExportLive(ctx, mpt.NewExportableArchiveTrie(archive, block), out)
```
HerbertJordan, line 86: Why is this moved to the `mpt` package?

go/state/cppstate/cpp_state.go
```diff
@@ -285,6 +287,10 @@ func (cs *CppState) CreateWitnessProof(address common.Address, keys ...common.Ke
 	panic("not implemented")
 }
 
+func (cs *CppState) CreateLiveDBGenesis(out io.Writer) (common.Hash, error) {
+	panic("not implemented")
```
HerbertJordan, line 291: Should return an error, not panic.

go/state/gostate/go_state.go
```diff
@@ -368,6 +370,10 @@ func (s *GoState) GetProof() (backend.Proof, error) {
 	return backend.GetComposedProof(proofs), nil
 }
 
+func (s *GoState) CreateLiveDBGenesis(io.Writer) (common.Hash, error) {
+	panic("not implemented")
```
HerbertJordan, line 374: Prefer errors over `panic`.

HerbertJordan (changes requested):
I think some of the code movements of export functions are adding additional complexity that should be avoided (in particular the need for the `common/genesis` package).

Before this PR, exporting a dump for a single world state from an archive was already introduced by #966, and it is working. It is surprising that the need make this public available through the Carmen facade causes so much refactoring of the already existing export infrastructure.

Furthermore, I think we should avoid the term `Genesis` altogether since what we are exporting is not a `Genesis` dataset, merely a small part of it.

go/carmen/carmen.go
petr-hanzl: You are right, this is only part of the whole `genesis` file. I will remove the term from the whole PR and simply use `Export()`.

go/database/mpt/io/live.go
```diff
@@ -54,9 +54,16 @@ const (
...
+	return exportableArchiveTrie{
+		trie:  trie,
+		block: block,
+	}
+}
+
+// MptStateVisitor is an interface for Tries that allows for visiting the Trie nodes
 // and furthermore getting its properties such as a root hash and contract codes.
-type mptStateVisitor interface {
+type MptStateVisitor interface {
```
HerbertJordan, line 66: There is not really a need to make this interface public.

If this interface is going to be public, I have higher quality expectations on its name.
The thing this interfaces defined is not a `Visitor`, so it shout not be named this way.

How about `VisitableState` or `ExportableState`?

HerbertJordan (changes requested):
Some minor requests for improvements; see coments.

go/database/mpt/io/live.go
```diff
@@ -54,9 +54,16 @@ const (
 	EthereumHash = HashType(0)
 )
 
-// mptStateVisitor is an interface for Tries that allows for visiting the Trie nodes
+func NewExportableArchiveTrie(trie *mpt.ArchiveTrie, block uint64) MptStateVisitor {
```
HerbertJordan, line 57: Needs documentation.

go/database/mpt/io/live.go
petr-hanzl: made private

go/carmen/example_test.go
```diff
@@ -515,3 +519,81 @@ func ExampleDatabase_GetMemoryFootprint() {
...
+	}
+
+	if err := db.Close(); err != nil {
+		log.Fatalf("cannot close db: %v", err)
+	}
+	if err := os.RemoveAll(dir); err != nil {
+		log.Fatalf("cannot remove dir: %v", err)
+	}
+
+	importedDbPath := filepath.Join(os.TempDir(), "imported_carmen")
```
HerbertJordan, line 571: This can lead to collisions with already existing copies of previous runs. Better to use
```
os.MkdirTemp("", "carmen_db_*")
```
like in the first line of this test.

HerbertJordan (dismissed):

HerbertJordan (approved):

Issue #946, opened 2024-07-03
Add support for recovering Archive data after a crash


HerbertJordan:
Carmen's data consistency policy defines any Database that was not properly shut down as corrupted. This is fine for LiveDB instances, where a recovery from an external source takes less than an hour, but for Archives this recovery takes 6-8 days.

To mitigate this issue, a recovery procedure for incorrectly closed Archives shall be provided.

PR #985, opened 2024-07-31
Add checkpoint support to root lists

HerbertJordan:
This PR adds checkpoint and recovery support to root lists. This is step 3 towards supporting Archive checkpoints and quick recovery in Carmen.

This is part of #946.


petr-hanzl (approved):
Thanks for adding another checkpoint support!

PR #986, opened 2024-07-31
Add db-heal test script

petr-hanzl:
This PR adds a test script which tests new feature `db-heal`
To use this script please define `Aida paths` on top of the file.

TODO: 
- [ ] Allow Aida to reopen database created by interrupted `aida-vm-sdb`
- [ ] Add looping

### DIFF
--- a/scripts/db-heal-test.sh
+++ b/scripts/db-heal-test.sh
@@ -1,0 +1,70 @@
+#!/bin/bash
+
+###########################
+#--- Dynamic variables ---#
+###########################
+
+# Aida paths
+aida_path=''
+aida_db_path=''
+tmp_path=''
+
+# Block variables
+sync_block=20000
+kill_block=11000
+restore_block=10000
+final_block=21000
+
+##########################
+#--- Static variables ---#
+##########################
+
+carmen_root=$(cd ../go && pwd)
+
+
+################
+#--- Script ---#
+################
+
+COMMAND="./build/aida-vm-sdb substate --db-tmp "$tmp_path" --carmen-schema 5 --db-impl carmen --aida-db "$aida_db_path" --no-heartbeat-logging --track-progress --archive --archive-variant s5 0 "$sync_block""
+
+# Run the command in the background and redirect stdout and stderr to a log file
+LOG_FILE="$(pwd)/output.log"
+current=$(pwd)
+
+cd $aida_path
+$COMMAND &> "$LOG_FILE" &
+COMMAND_PID=$!
+cd $current
+
+echo "Starting aida-vm-sdb with interrupt."
+
+# Function to monitor the log file
+monitor_log() {
+  tail -F "$LOG_FILE" | while read LINE; do
+    echo "$LINE" | grep -q "block $kill_block"
+    if [ $? -eq 0 ]; then
+      echo "Interrupting."
+      kill $COMMAND_PID
+      exit 0
+    fi
+  done
+}
+
+# Start monitoring the log file
+monitor_log
+
+# Wait for the command to complete
+wait $COMMAND_PID
+
+# Find working dir
+working_dir=$(ls -td "$tmp_path"/*/ | head -1)
+
+(cd $carmen_root && go run ./database/mpt/tool reset --force-unlock "$working_dir"archive "$restore_block")
+
+echo "Restoration complete. Syncing to block $final_block."
+
+final_first=$((restore+1))
+(cd $aida_path && ./build/aida-vm-sdb substate --db-tmp "$tmp_path" --carmen-schema 5 --db-impl carmen --aida-db "$aida_db_path" --no-heartbeat-logging --track-progress --archive --archive-variant s5 --db-src "$working_dir" "$final_first" "$final_block")
+
+echo "Sync complete to block $final_block. Final db path: $(ls -td "$tmp_path"/*/ | head -1)."


scripts/db-heal-test.sh
```diff
@@ -0,0 +1,70 @@
...
+# Start monitoring the log file
+monitor_log
+
+# Wait for the command to complete
+wait $COMMAND_PID
+
+# Find working dir
+working_dir=$(ls -td "$tmp_path"/*/ | head -1)
+
+(cd $carmen_root && go run ./database/mpt/tool reset --force-unlock "$working_dir"archive "$restore_block")
```
HerbertJordan, line 63: The restoration step should also needs the export of a LiveDB genesis from the restored block at the restored block height + an import into the LiveDB of the working directory.

HerbertJordan (changes requested):
Thanks for the script, I will try it on my dev system.

scripts/db-heal-test.sh
```diff
@@ -0,0 +1,70 @@
...
+##########################
+
+carmen_root=$(cd ../go && pwd)
+
+
+################
+#--- Script ---#
+################
+
+COMMAND="./build/aida-vm-sdb substate --db-tmp "$tmp_path" --carmen-schema 5 --db-impl carmen --aida-db "$aida_db_path" --no-heartbeat-logging --track-progress --archive --archive-variant s5 0 "$sync_block""
```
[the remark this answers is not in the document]
petr-hanzl: I was thinking about adding it, but first command (the interrupted one) actually does not make sense to add validation, does it?

Anyway, since it only slows it down by a forgettable time, we can just add it.

scripts/db-heal-test.sh
```diff
@@ -0,0 +1,70 @@
+#!/bin/bash
+
```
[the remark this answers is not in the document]
petr-hanzl: Using this actually makes the script not `kill` the process on found block. I will not add it and find a better way or fix it.

scripts/db-heal-test.sh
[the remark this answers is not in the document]
HerbertJordan: Not sure, does `--validate` also enable `--validate-state-hash` ? The latter, I think, would be the strongest check that we got it right.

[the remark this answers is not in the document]
petr-hanzl: `--validate` enables every validation

scripts/db-heal-test.sh
```diff
@@ -0,0 +1,127 @@
...
+  (cd $carmen_root && go run ./database/mpt/tool import-live-db "$genesis" "$live")
+
+  echo "Iteration "$i"/"$number_of_iterations""
+  first_block=$((restore_block + 1))
+  last_block=$((last_block + 1000))
+  restore_block=$((restore_block + 1000))
+  kill_block=$((kill_block + 1000))
+
+  echo "Syncing to block "$last_block"..."
+  command="./build/aida-vm-sdb substate --validate --db-tmp "$tmp_path" --carmen-schema 5 --db-impl carmen --aida-db "$aida_db_path" --no-heartbeat-logging --track-progress --archive --archive-variant s5 --archive-query-rate 200 --carmen-cp-interval 200 --db-src "$working_dir" --skip-priming "$first_block" "$last_block""
```
kjezek, line 100: use `go run` ?

scripts/db-heal-test.sh
```diff
@@ -0,0 +1,127 @@
...
+  (cd $carmen_root && go run ./database/mpt/tool export --block "$restore_block" "$archive" "$genesis")
+
+  # Restore LiveDB
+  echo "Export complete. Applying LiveDB genesis."
+  rm -rf "$live"
+  (cd $carmen_root && go run ./database/mpt/tool import-live-db "$genesis" "$live")
+
+  echo "Iteration "$i"/"$number_of_iterations""
+  first_block=$((restore_block + 1))
+  last_block=$((last_block + 1000))
```
kjezek, line 95: would be nice to have the windows size (1000) also as a variable

scripts/db-heal-test.sh
```diff
@@ -0,0 +1,127 @@
...
+
+  # Re-find working dir - Aida copies db-src
+  working_dir=$(ls -td "$tmp_path"/*/ | head -1)
+  archive="${working_dir}archive"
+  live="${working_dir}live"
+
+done
+
+rm $log_file
+
```
kjezek, line 126: I guess all remaining tmp files should be deleted here. The last iteration was not deleted right?

scripts/db-heal-test.sh
```diff
@@ -0,0 +1,127 @@
+#!/bin/bash
+
```
kjezek, line 2: a documentation how to use the script should be added at the beginning of the script. For example, it is not clear where the script should be run (in Aida, in Carmen, ...) and what the input parameters are.

scripts/db-heal-test.sh
petr-hanzl: im not sure what you mean by that

scripts/db-heal-test.sh
petr-hanzl: good catch

scripts/db-heal-test.sh
kjezek: if I understand correctly, it synchronises next 1000 blocks in each iteration. I meant that the number of 1000 could be configurable, i.e. be a global variable

petr-hanzl: ic. Sure

scripts/db-heal-test.sh
```diff
@@ -0,0 +1,127 @@
...
+################
+#--- Script ---#
+################
+
+# Run the command in the background and redirect stdout and stderr to a log file
+log_file="$(pwd)/output.log"
+current=$(pwd)
+
+# First iteration has different command
+cmd="./build/aida-vm-sdb substate --validate --db-tmp "$tmp_path" --carmen-schema 5 --db-impl carmen --aida-db "$aida_db_path" --no-heartbeat-logging --track-progress --archive --archive-variant s5 --archive-query-rate 200 --carmen-cp-interval 200 "$first_block" "$last_block""
```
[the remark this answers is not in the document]
petr-hanzl: using `go run` seems to have issues with terminating the process. Nor `terminate` nor `kill` worked very well, I will keep the binary as it 100% works.

kjezek (commented):
I think it would make this experiment stronger to make the distance between last block and the block to restore variable. 

For instance, the window size is 1000 blocks, and the number of iterations is the same, you can try to restore blocks:
```
[last - 1000, last - 999, ... ,last - 2, last - 1]
```
(or customise this interval based on check-point rate)

otherwise, I do not see too many reasons why the experiment runs in a loop - well it may happen that the block 800 can be restored and the block 1800 cannot, but it is unlikely considering the implementation is tested for functionality by unit tests, and thus basic functionality is expected to work.

petr-hanzl:
> I think it would make this experiment stronger to make the distance between last block and the block to restore variable.
> 
> For instance, the window size is 1000 blocks, and the number of iterations is the same, you can try to restore blocks:
> 
> ```
> [last - 1000, last - 999, ... ,last - 2, last - 1]
> ```
> 
> (or customise this interval based on check-point rate)
> 
> otherwise, I do not see too many reasons why the experiment runs in a loop - well it may happen that the block 800 can be restored and the block 1800 cannot, but it is unlikely considering the implementation is tested for functionality by unit tests, and thus basic functionality is expected to work.

That is not possible at this moment. We would need to create checkpoint every block which I think is not what we really want to do in this test. Instead of doing one block after we can do one checkpoint after another 

ie. Lets say checkpoint is every 100 blocks. Then we can do
```
[last - 900, last - 800, ... , last - 100]
```

Although I feel like this adds unnecessary complexity to the test.
wdyt? @kjezek

kjezek:
> > I think it would make this experiment stronger to make the distance between last block and the block to restore variable.
> > For instance, the window size is 1000 blocks, and the number of iterations is the same, you can try to restore blocks:
> > ```
> > [last - 1000, last - 999, ... ,last - 2, last - 1]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > (or customise this interval based on check-point rate)
> > otherwise, I do not see too many reasons why the experiment runs in a loop - well it may happen that the block 800 can be restored and the block 1800 cannot, but it is unlikely considering the implementation is tested for functionality by unit tests, and thus basic functionality is expected to work.
> 
> That is not possible at this moment. We would need to create checkpoint every block which I think is not what we really want to do in this test. Instead of doing one block after we can do one checkpoint after another
> 
> ie. Lets say checkpoint is every 100 blocks. Then we can do
> 
> ```
> [last - 900, last - 800, ... , last - 100]
> ```
> 
> Although I feel like this adds unnecessary complexity to the test. wdyt? @kjezek

right, it does not (should not) checkpoint every block, but the window should be rather narrow to stress-test checkpoint creation.  Why do you think it is complicated? it would need to compute the block to `reset` to e.g. as `lastBlock - (windowSize - i * 100)` that's it. 

As I wrote above, having still the same distance makes the experiment less valuable because it tests still the same even if it runs million times, while I guess we want to test that the checkpoint is far in the past as well as very recent.

scripts/db-heal-test.py
```diff
@@ -0,0 +1,217 @@
+import os
+import subprocess
+import tempfile
+import time
+import shutil
+from pathlib import Path
+import sys
+import flag
+
```
kjezek, line 9: the script should have been documented: what it does, how to run it, what prerequisites must be met (python packages installation, etc) .

kjezek:
I run the script, but it stuck: 

```
python3 ./scripts/db-heal-test.py  --aida '/Users/kjezek/fantom/Aida' --tmp /Volumes/ext2tb --aidadb /Volumes/ext2tb/aida-db 
Creating database with aida-vm-sdb...
```

In noticed there was runtime crash in Aida. In this case the script should terminate, perhaps print aida log.

kjezek:
> > > I think it would make this experiment stronger to make the distance between last block and the block to restore variable.
> > > For instance, the window size is 1000 blocks, and the number of iterations is the same, you can try to restore blocks:
> > > ```
> > > [last - 1000, last - 999, ... ,last - 2, last - 1]
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > (or customise this interval based on check-point rate)
> > > otherwise, I do not see too many reasons why the experiment runs in a loop - well it may happen that the block 800 can be restored and the block 1800 cannot, but it is unlikely considering the implementation is tested for functionality by unit tests, and thus basic functionality is expected to work.
> > 
> > 
> > That is not possible at this moment. We would need to create checkpoint every block which I think is not what we really want to do in this test. Instead of doing one block after we can do one checkpoint after another
> > ie. Lets say checkpoint is every 100 blocks. Then we can do
> > ```
> > [last - 900, last - 800, ... , last - 100]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Although I feel like this adds unnecessary complexity to the test. wdyt? @kjezek
> 
> right, it does not (should not) checkpoint every block, but the window should be rather narrow to stress-test checkpoint creation. Why do you think it is complicated? it would need to compute the block to `reset` to e.g. as `lastBlock - (windowSize - i * 100)` that's it.
> 
> As I wrote above, having still the same distance makes the experiment less valuable because it tests still the same even if it runs million times, while I guess we want to test that the checkpoint is far in the past as well as very recent.

Alternative (perhaps better) idea: at the moment, always 900's block is killed, and 800's block is reverted to. It means there is always 100 blocks difference. I think it would be interesting to make this gap smaller in each iteration up to a critical point = the checkpoint is 1 block old.

kjezek:
Progressing further, I got this error:

```
> python3 ./scripts/db-heal-test.py  --aida '/Users/kjezek/fantom/Aida' --tmp /Volumes/ext2tb --aidadb /Volumes/ext2tb/aida-db
Creating database with aida-vm-sdb...
Interrupting.
Testing db created, starting loop.

Traceback (most recent call last):
  File "/Users/kjezek/fantom/Carmen/./scripts/db-heal-test.py", line 108, in <module>
    c = open(carmen_log_file, 'r+')
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '/Users/kjezek/fantom/Carmen/carmen.log'

```

scripts/db-heal-test.py
```diff
@@ -0,0 +1,231 @@
...
+checkpoint_granularity = checkpoint_granularity_flag.value
+
+# Mark first checkpoint
+latest_checkpoint = checkpoint_granularity
+
+if aida_path == "":
+    print("please set Aida using --aida")
+    exit(1)
+if aida_db_path == "":
+    print("please set AidaDB using --aidadb")
```
kjezek, line 35: this took me some time:  I would somehow naturally expect `--aida-db`, but here it says to use `--aidadb`, actually it works with `--aida_db`

I would suggest to unify to  `--aida-db`
* without separator in the middle it is difficult to read
* underscore I think is not very commonly used as an argument.

kjezek: also for other parameters, I think underscore does not seem natural being used as a program parameter.

scripts/db-heal-test.py
```diff
@@ -0,0 +1,231 @@
...
+current_dir = Path.cwd()
+
+print("Your settings:")
+print(f"\tNumber of iterations: {number_of_iterations}.")
+print(f"\tWindow size: {window} seconds.")
+print(f"\tCheckpoint granularity: {checkpoint_granularity} blocks.")
+
+
+# Function which stops process after given sleep_time.
+def terminate_process_after(sleep_time: int, checkpoint: int):
```
kjezek, line 63: I think the last checkpoint and the time window when to kill the process should be tracked in two separate threads.

At the moment when I configure CP every 1M block and kill time 5s, the program is not interrupted every 5s - it is interrupted after each 1M blocks (close after this block heigh is done as the timeout is long overdue)

We wanted to test the 
1. a program is killed at arbitrary time  - i.e. may be a fixed window of X seconds
2. right when the program is killed, we check what is the last checkpoint and recover it

current solution IMHO still depends (waits for) a checkpoint being done and only after that it recovers. 

not sure if it is the best way to get the last checkpoint by parsing the log, ideally Carmen should tell what is the last CP, but not sure we have this option at the moment.

scripts/db-heal-test.py
```diff
@@ -0,0 +1,231 @@
...
+            if start > 0 and time.time() - start >= sleep_time:
+                print("Interrupting...")
+                process.terminate()
+                return checkpoint
+            if f"block {checkpoint + checkpoint_granularity}" in line:
+                # First checkpoint was found, we should start timer as it means block processing is running.
+                if start == 0.0:
+                    start = time.time()
+                checkpoint = checkpoint + checkpoint_granularity
+                print(f"Found new checkpoint {checkpoint}.")
```
kjezek, line 80: this was a bit annoying to read the checkpoint number in the log, when the experiment runs long and CP is set short, it floods the console

scripts/db-heal-test.py
```diff
@@ -0,0 +1,231 @@
...
+    if result.returncode != 0:
+        # Print carmen logs if interation fails
+        text = c.read()
+        print(text)
+        sys.exit(1)
+
+    # Export genesis to restore LiveDB
+    genesis = Path(tmp_path) / 'test_genesis.dat'
+
+    print(f"Restoration complete. Exporting LiveDB genesis block {latest_checkpoint}.")
```
kjezek, line 151: I think it is good to have experiment that continues sync after db was healed, but I think it would be worth having a separate script (pipeline?) for this. 

When I want to run experiment that creates a many CPs and kills the app often, verifying the reset works.  I think continuing with export/import/sync adds too much extra time

kjezek (changes requested):
Thanks for this script, unfortunately I think it has still a few flaws.

1. I am getting following error:
```
kjezek@Kamils-MacBook-Pro:/Users/kjezek/fantom/Carmen$ python3 scripts/db-heal-test.py --aida /Users/kjezek/fantom/Aida --aida_db /Volumes/ext2tb/aida-db --cp_granularity 1000000 --window 30
tmp not set - using default /var/folders/xw/04df617n7rsfr5__h58dvzpr0000gp/T
Your settings:
	Number of iterations: 1000.
	Window size: 30 seconds.
	Checkpoint granularity: 1000000 blocks.
Creating database with aida-vm-sdb...
Found new checkpoint 2000000.
Interrupting...
Testing db created, starting loop.
Traceback (most recent call last):
  File "/Users/kjezek/fantom/Carmen/scripts/db-heal-test.py", line 137, in <module>
    result = subprocess.run(
             ^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.12/3.12.5/Frameworks/Python.framework/Versions/3.12/lib/python3.12/subprocess.py", line 548, in run
    with Popen(*popenargs, **kwargs) as process:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.12/3.12.5/Frameworks/Python.framework/Versions/3.12/lib/python3.12/subprocess.py", line 1026, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/opt/homebrew/Cellar/python@3.12/3.12.5/Frameworks/Python.framework/Versions/3.12/lib/python3.12/subprocess.py", line 1955, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: '/Users/kjezek/fantom/go'
```

I could be probably able to debug what is wrong, but the script should be documented, how to use it, i.e. I do not try to investigate at the moment.

2. I do not think the loop to kill the process is what we exactly wanted. We need to (1) kill the process at arbitrary time, (2) read what is the last CP, (3) recover this checkpoint. At the moment the loop still waits until the CP is done, then it potentially waits more, and only then it kills.  See my comment inline as well

3. Thinking more of the script, I think it is good to test the reset/export/import/sync chain works, but perhaps we should put it in another script (perhaps just a pipeline), because there are two use-cases mixed-up in current script: (1) functionality of checkpoint - it needs to be tested by many tight loops, (2) functionality of the whole toolchain - this may be one or a few iterations once functionality of (1) is granted.

scripts/db-heal-test.py
petr-hanzl: unfortunately python (or this library) does not allow dashes in flag names.

scripts/db-heal-test.py
petr-hanzl: Yes it does wait until the first CP is done, but thats the only way to actually know Aida started block processing. 
Then it waits some time chosen by `--window` and then it terminates Aidas process. I thought the aim of this script was to be able to terminate the program at the time checkpoint is being created, which does not mean it needs to kill it at the first checkpoint.

Also why would we even want to kill the process without having the checkpoint? 

With `--cp_granularity` you choose how often (in blocks) Carmen creates checkpoints. It also choses how often Aida logs progress, so that it logs the block number at the time when Carmen creates checkpoint (tldr `carmen-checkpoint-interval` == `aida-tracker-interval`)...
If you chose 1M blocks it makes sense the script wont kill the process in 5s because there is no way Aida could sync to 1M in that time window and I think this is correct behaviour so that you do not chose weird combination because the script would do nothing.

scripts/db-heal-test.py
petr-hanzl: I know, this is somewhat debug print. This will be removed once we finish discussions about how we want to make this work.

scripts/db-heal-test.py
petr-hanzl: Without the export/import/sync you actually dont test the whole `heal` feature. I think it should stay in here. Thats actually what we discussed with @HerbertJordan when I got assigned this task.

scripts/db-heal-test.py
kjezek: could you use this library: https://docs.python.org/dev/library/argparse.html#dest

it seems to support this.  But most importantly, it seems to be standard, and already installed with python. The current library I had to install with some hack bypassing pip sandboxing

scripts/db-heal-test.py
kjezek: The key message is that you do not kill at arbitrary time, but always wait after the CP is ready (and potentially then it wats a bit more).

if you want to measure time only after there is at least one CP, then yes, I agree it is a good idea - but it cannot break what we want to reach - kill at arbitrary (random) time

scripts/db-heal-test.py
kjezek: I said split it in two scripts/pipelines as they can run at different pace, when it is split, we loose nothing

scripts/db-heal-test.py
```diff
@@ -0,0 +1,231 @@
+import os
+import subprocess
+import tempfile
+import time
+import shutil
+from pathlib import Path
+import sys
+import flag
+
+# --- Dynamic variables --- #
```
HerbertJordan, line 10: "Dynamic variables" aka "Parameters"?

scripts/db-heal-test.py
```diff
@@ -0,0 +1,231 @@
...
+aida_path_flag = flag.string("aida", "", "Path to Aida root.")
+aida_db_path_flag = flag.string("aida_db", "", "Path to AidaDB.")
+tmp_path_flag = flag.string("tmp", "", "Path to tmp dir.")
+
+number_of_iterations_flag = flag.int("iter", 1000, "Number of iterations.")
+window_flag = flag.int("window", 120, "How often will program get terminated (in seconds).")
```
HerbertJordan, line 16:
```suggestion
window_flag = flag.int("window", 120, "Delay between start of sync process and forced termination (in seconds).")
```

scripts/db-heal-test.py
```diff
@@ -0,0 +1,231 @@
...
+checkpoint_granularity_flag = flag.int("cp_granularity", 2000, "How often will Carmen create checkpoints.")
```
HerbertJordan, line 17:
```suggestion
checkpoint_granularity_flag = flag.int("cp_granularity", 2000, "How often will Carmen create checkpoints (in blocks)")
```

scripts/db-heal-test.py
```diff
@@ -0,0 +1,231 @@
...
+flag.parse()
```
HerbertJordan, line 18: These default parameters look very relaxed. Only one iteration every 2 minutes is not a high frequency.

Also, 1000 iterations with 2 minutes sync periods take at least 1.3 days.

Why not checkpoint after each block, or after each 10 blocks, and kill after 5 seconds?

scripts/db-heal-test.py
```diff
@@ -0,0 +1,231 @@
...
+# --- Script --- #
+
+# Log file path from which we read output to find kill_block
+aida_log_file = Path.cwd() / 'aida.log'
+carmen_log_file = Path.cwd() / 'carmen.log'
+current_dir = Path.cwd()
+
+print("Your settings:")
+print(f"\tNumber of iterations: {number_of_iterations}.")
+print(f"\tWindow size: {window} seconds.")
```
HerbertJordan, line 58:
```suggestion
print(f"\tSync time before kill: {window} seconds.")
```

scripts/db-heal-test.py
```diff
@@ -0,0 +1,231 @@
...
+                    start = time.time()
+                checkpoint = checkpoint + checkpoint_granularity
+                print(f"Found new checkpoint {checkpoint}.")
+            # If process ends with error (return code 1) or either 'fail' or 'exit status' occurs in line exit script
+            if process.poll() == 1 or any(s in line for s in ["exit status", "fail"]):
+                print("Error occurred - printing output.log:")
+                with open(aida_log_file, 'r') as l:
+                    text = l.read()
+                    print(text)
+                os.chdir(current_dir)
```
HerbertJordan, line 87: This line seems to be not part of the responsibility of this function ... changing the current directory in case of an error should probably be added before line 116 and 217, before `sys.exit(1)` is called.

scripts/db-heal-test.py
```diff
@@ -0,0 +1,231 @@
...
+    result = subprocess.run(
+        ['go', 'run', './database/mpt/tool', 'import-live-db', str(genesis), str(live)],
+        stdout=c,
+        stderr=c,
+        cwd=carmen_root)
+    if result.returncode != 0:
+        # Print carmen logs if interation fails
+        text = c.read()
+        print(text)
+        sys.exit(1)
```
HerbertJordan, line 188: nit: this block running a command and handling the error code is present several times; maybe it should be factory out into a utility function

scripts/db-heal-test.py
```diff
@@ -0,0 +1,231 @@
...
+        shutil.rmtree(last_working_dir, ignore_errors=True)
+
+    genesis.unlink(missing_ok=True)
+# Clear anything leftover
+
+print(f"Clearing last database {working_dir} and log files.")
+aida_log_file.unlink(missing_ok=True)
+carmen_log_file.unlink(missing_ok=True)
+shutil.rmtree(working_dir, ignore_errors=True)
+
```
HerbertJordan, line 230: nit: It is always nice to have a final message like "All tests passed" or "Success", just to know that the process just not crashed with something unforeseeable.

HerbertJordan (approved):
Looks good to me, but I think the parameters need to be further tweaked.

The first goal is to basically create checkpoints all the time while syncing the DB to maximize the chance of interrupting the checkpoint creation. The other goal is to do as many iterations per minute as possible.

scripts/db-heal-test.py
HerbertJordan: @kjezek what would the first script do instead of continuing with the export/import/sync?

scripts/db-heal-test.py
petr-hanzl: I normally run this script with checkpoint every block and kill after 1s. These defaults are archaic from earlier versions. Thanks for noticing.

scripts/db-heal-test.py
kjezek: I thought it could only do the db reset, possibly db reset +  export genesis. 
The idea was to speed-up this test, which stress-test checkpoint creation.

scripts/db-heal-test.py
```diff
@@ -0,0 +1,238 @@
...
+                print("Error occurred - printing output.log:")
+                with open(aida_log_file, 'r') as l:
+                    text = l.read()
+                    print(text)
+                return -1
+
+
+# First iteration command
+cmd = [
+    './build/aida-vm-sdb', 'substate', '--validate',
```
kjezek, line 106: as I mention previously, it would make more sense to me to configure the script directly to run `aida-vm-sdb` - why it has to be the Aida workspace as it appears in git?

kjezek: if the script requires Aida workspace, it would make more sense to run the command using `go run` so the user does not have to build Aida before

kjezek (changes requested):
Thanks for the script. Reading the code, it looks good to me.

Unfortunately I am not able to run it. When I do not provide required arguments, it crashes - it used to show some useful help in previous iteration, it seems to be gone now. Even when I provide compulsory parameters, it still crashes. 

Generally, the script misses proper documentation. 

Here is what I tried to run, and results I got:

```
 python3 scripts/db-heal-test.py 
Your settings:
	Number of iterations: 1000.
	Sync time before kill: 5 seconds.
	Checkpoint granularity: 10 blocks.
Traceback (most recent call last):
  File "/Users/kjezek/fantom/Carmen/scripts/db-heal-test.py", line 114, in <module>
    os.chdir(aida_path)
TypeError: chdir: path should be string, bytes, os.PathLike or integer, not NoneType



python3 scripts/db-heal-test.py --aida /Users/kjezek/fantom/Aida 
Your settings:
	Number of iterations: 1000.
	Sync time before kill: 5 seconds.
	Checkpoint granularity: 10 blocks.
Traceback (most recent call last):
  File "/Users/kjezek/fantom/Carmen/scripts/db-heal-test.py", line 116, in <module>
    process = subprocess.Popen(cmd, stdout=f, stderr=subprocess.STDOUT)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.12/3.12.5/Frameworks/Python.framework/Versions/3.12/lib/python3.12/subprocess.py", line 1026, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/opt/homebrew/Cellar/python@3.12/3.12.5/Frameworks/Python.framework/Versions/3.12/lib/python3.12/subprocess.py", line 1885, in _execute_child
    self.pid = _fork_exec(
               ^^^^^^^^^^^
TypeError: expected str, bytes or os.PathLike object, not NoneType



python3 scripts/db-heal-test.py --aida /Users/kjezek/fantom/Aida --aida-db /Volumes/ext2tb/aida-db 
Your settings:
	Number of iterations: 1000.
	Sync time before kill: 5 seconds.
	Checkpoint granularity: 10 blocks.
Traceback (most recent call last):
  File "/Users/kjezek/fantom/Carmen/scripts/db-heal-test.py", line 116, in <module>
    process = subprocess.Popen(cmd, stdout=f, stderr=subprocess.STDOUT)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.12/3.12.5/Frameworks/Python.framework/Versions/3.12/lib/python3.12/subprocess.py", line 1026, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/opt/homebrew/Cellar/python@3.12/3.12.5/Frameworks/Python.framework/Versions/3.12/lib/python3.12/subprocess.py", line 1885, in _execute_child
    self.pid = _fork_exec(
               ^^^^^^^^^^^
TypeError: expected str, bytes or os.PathLike object, not NoneType

```

scripts/db-heal-test.py
cabrador: As written in [this](https://github.com/Fantom-foundation/Carmen/pull/986#discussion_r1727097445) comment, using `go run` had issues...

scripts/db-heal-test.py
```diff
@@ -0,0 +1,238 @@
...
+        stderr=c,
+        cwd=carmen_root)
+    check_program_failure(result.returncode, c)
+
+    # Restore LiveDB
+    print("Export complete. Applying LiveDB genesis.")
+    try:
+        shutil.rmtree(live)
+        print("Live directory removed successfully.")
+    except FileNotFoundError:
```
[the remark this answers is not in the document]
cabrador: I guess we can remove that, there was an issue with catching these errors so I added try/catch block to specifically log these.

scripts/db-heal-test.py
```diff
@@ -0,0 +1,238 @@
...
+    latest_checkpoint = terminate_process_after(window, latest_checkpoint)
+
+    # Wait for the command to complete
+    process.wait()
+
+    if latest_checkpoint == -1:
+        os.chdir(current_dir)
+        sys.exit(1)
+
+    if last_working_dir:
```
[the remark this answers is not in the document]
cabrador: The reason why I delete it every iteration is because if the test fails, all of the data is kept there. This way only the latest db is prevailed

scripts/db-heal-test.py
```diff
@@ -0,0 +1,238 @@
...
+
+for i in range(1, number_of_iterations + 1):
+    last_working_dir = working_dir
+
+    # Find working dir - Aida copies db-src
+    working_dir = max(Path(tmp_path).iterdir(), key=os.path.getmtime)
+    archive = working_dir / 'archive'
+    live = working_dir / 'live'
+
+    # Dumb carmen's logs into a file to avoid spamming
```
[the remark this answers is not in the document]
cabrador: haha good catch

scripts/db-heal-test.py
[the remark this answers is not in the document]
kjezek: but I guess it is because you terminate the loop (the whole script) by `sys.exit()`. If you `break` the loop, it would reach end of the script, where the temp directory can be deleted.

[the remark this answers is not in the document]
cabrador: I feet like there could be scenario where we would like to keep the broken db.

scripts/db-heal-test.py
cabrador: Discussed in DMs.

scripts/db-heal-test.py
HerbertJordan: In this case we would have to always start from an empty archive in each iteration?

HerbertJordan (commented):
Looks good to me, but I have not had the chance yet to test-run it. Will approve once I managed to do so.

scripts/db-heal-test.py
kjezek: I really insists on having some documentation here. I had again hard time to start the script, it printed many errors first, then I realised it must be run from the `script` directory - this is nowhere documented, how can one know it?

Furthermore, running the script from the root directory is IMHO more intuitive than running from any subdirectories. It is more straightforward to integrate in further scripts like pipelines - one just checkouts the repository and executes the script, no further `cd` is needed.

kjezek (changes requested):
I was successful to run the test, but

1. do the proper documentation - I have already asked for this 3-times. Please next time do the documentation as the first step, not the last one, so the reviewers do not have to struggle each iteration.
2. the mistake I was doing was that I run the script from the carmen root - I think it should work this way, it is (1) more intuitive, (2) better for further integration into scripts - as only checkout + execution is needed when running e.g. from a pipeline, no further `cd` is required.   

Furthermore, the script prints each iteration this: 

```
Iteration 27/1000
Syncing restarted...
Interrupting...
Forcing unlock of directory /var/folders/xw/04df617n7rsfr5__h58dvzpr0000gp/T/db-heal-test/state_db_tmp_2996455141/archive ...
Resetting archive in /var/folders/xw/04df617n7rsfr5__h58dvzpr0000gp/T/db-heal-test/state_db_tmp_2996455141/archive to block 15180 ...
block 15180 is beyond the last checkpoint height of 15177
exit status 1

Error occured during reset - looking for checkpoint using 'info' command...
Found on block 15177.
Success! Using block 15177
Restoration complete. Exporting LiveDB genesis block 15177.
Export complete. Applying LiveDB genesis.

```

I think the problem is that the script tries to source the checkpoint from two sources: (1) Aida log, (2) carmen.  The source (1) always fails - this is odd, I do not know what is the problem, but there is something obviously not right


Also relates to a comment I had at one of the previous iterations: why is the checkout process driven by Aida log? It would be much simpler, if the script (1) just kills the process at configured time (starting the timer when first checkpoint is ready), (2) reads from carmen what the last checkpoint is, (3) and resets to this checkpoint.

I have never understood why the checkpoint is parsed from the log, and it turn out not functioning anyway - i.e. it did not clarify my doubts.

scripts/db-heal-test.py
```diff
@@ -0,0 +1,263 @@
...
+
+# --- Script --- #
+
+# Create working dir which gets deleted after the run
+working_dir = os.path.join(tmp_path, 'db-heal-test')
+if os.path.exists(working_dir):
+    shutil.rmtree(working_dir)
+os.makedirs(working_dir)
+
+# Log file path from which we read output to find kill_block
```
kjezek, line 64: I think these files should be stored also in `tmp`, this way they will get cleared-up. At the moment they persist in current dir.

scripts/db-heal-test.py
```diff
@@ -0,0 +1,263 @@
...
+aida_log_file = Path.cwd() / 'aida.log'
+carmen_log_file = Path.cwd() / 'carmen.log'
+genesis = os.path.join(working_dir, 'test_genesis.dat')
+
+current_dir = Path.cwd()
+carmen_root = os.path.join(current_dir, 'go')
+if "scripts" in str(current_dir):
```
kjezek, line 71: I guess it is here to allow for running the script either from the `script` directory or from a parent directory.  There is endless options where the users may try to run the script from. I think it does not matter much what is the correct way (albeit I think intuitive is the root dir), but it should be documented what users are supposed to do.

scripts/db-heal-test.py
```diff
@@ -0,0 +1,263 @@
...
+    log = os.path.join(working_dir, 'carmen-info.log')
+    cp: str
+    with open(log, 'w') as cl:
+        r = subprocess.run(
+            ['go', 'run', './database/mpt/tool', 'info', str(archive)],
+            stdout=cl,
+            stderr=cl,
+            cwd=carmen_root)
+        if has_program_failed(r.returncode, cl):
+            shutil.rmtree(log)
```
kjezek, line 128: if all logs were in temp dir, you would not have to delete them here and there, and you would just remove all at the end

kjezek (changes requested):
thanks for all iterations, I think the script looks much more elegant at the moment. I was able to run it without hassle. 

Please document the file at the header and also add the licence header. 

I further put a few recomendations

scripts/db-heal-test.py
cabrador: The thing is that I need to look for Carmens root. Thats why I only allow these two options. I dont like the idea of asking user to pass Carmen root via flag.

scripts/db-heal-test.py
cabrador: It actually is within the temp dir. This is archaic code. Removed

scripts/db-heal-test.py
```diff
@@ -0,0 +1,284 @@
...
+            cwd=carmen_root)
+        if has_program_failed(r.returncode, cl):
+            return -1
+
+    with open(log, 'r') as cl:
+        info_checkpoint = cl.readlines()[-1]
+        # Return last word which is the block number
+        cp = info_checkpoint.split()[-1]
+
+    os.remove(log)
```
kjezek, line 146: I guess you do not have to remove the log here, right? It will get removed together with the whole `tmp` dir

scripts/db-heal-test.py
```diff
@@ -0,0 +1,284 @@
...
+cmd = [
+    './build/aida-vm-sdb', 'substate', '--validate',
+    '--db-tmp', working_dir, '--carmen-schema', '5', '--db-impl', 'carmen',
+    '--aida-db', aida_db_path, '--no-heartbeat-logging', '--track-progress',
+    '--archive', '--archive-variant', 's5', '--archive-query-rate', '200',
+    '--carmen-checkpoint-interval', str(checkpoint_granularity), '--tracker-granularity',
+    str(checkpoint_granularity), str(first_block), str(last_block)
+]
+
+os.chdir(aida_path)
```
kjezek, line 171: do you really need to `chdir` here and there? It looks to me that all paths are absolute to the temp dir.

kjezek: also holds for other places in the script

scripts/db-heal-test.py
```diff
@@ -0,0 +1,284 @@
...
+import sys
+
+# The script is used for testing the db-heal feature in loop.
+# 1) It kills data processing hence makes database corrupted.
+# 2) Resets the Archive to block returned by info command.
+# 3) Exports and imports LiveDB from Archive.
+# The script needs to be run either in Carmen root or Carmen/scripts
+# as it looks for Carmen commands and uses it.
+# It also requires path to Aida root with BUILT Aida and AidaDb.
+# Example: python3 scripts/db-heal-test.py --aida /Path/To/Aida --aida-db /Path/To/Aida/Db
```
kjezek, line 17:
```suggestion
# This script performs an experiment to recover database after a severe failure. It runs synchronisation of the 
# blockchain having the archive enabled. At a configured time, the process running the synchronisation is killed,
# simulating a failure. When this happens, the script reads what is the last checkpoint of the archive and recovers 
# the database to this checkpoint. Synchronisation continues from this checkpoint to verify that the database was 
# recovered successfully. 
#
# To stress test the process, the script runs in a loop with frequent checkpoint creation and a tight window to kill 
# the process. 
#
# The script needs python with default packages installed.  It needs Aida being build in a separate directory and it
# further needs the Aida database. These two directories must be configured. 
#
# Example: python3 scripts/db-heal-test.py --aida /Path/To/Aida --aida-db /Path/To/Aida/Db
```

kjezek: This was not a good description. The text should really help the reader to understand what is the intention of the script. Would you understand from the sentence: `The script is used for testing the db-heal feature in loop.` what is the script doing if you were not an author of the script.  Even in Carmen we do not have any "db-heal feature", i.e. it is not a common term even in Carmen. We do have a set of features that together recovers/heal the database.  

It is much easier to understand the code if the documentation tells what is the intention of the code, it was missing completely in the documentation.

scripts/db-heal-test.py
```diff
@@ -0,0 +1,284 @@
+import argparse
```
kjezek, line 1: license header

kjezek (changes requested):
Please add license header.

Otherwise I think the script is ok to be merged. I found a few places where I think the script can be simplified. In particular, It looks to me that `chdir` into directories is not necessary when absolute paths are used. Deleting some selected files is also redundant when the whole directory is deleted anyway at the end. 

I think also that allowing the user to run the script from two directories is not necessary, once it is documented where to run, the user (and Jenkins) is just obliged to follow the requirement.

kjezek (approved):
thx, it looks much better now

scripts/db-heal-test.py
```diff
@@ -0,0 +1,285 @@
...
+
+    # Wait for the command to complete
+    process.wait()
+
+    if has_failed:
+        break
+
+# Clear anything leftover
+print(f"Clearing work directory {working_dir}.")
+shutil.rmtree(working_dir, ignore_errors=True)
```
HerbertJordan, line 278: To simplify debugging it would be good to only clear the data if the run was a success. Keeping the working directory in case of failures enables the inspection of the failure state.

If you do not like to keep the data around in any potential failure state, maybe there can be an optional `--keep-data` flag

cabrador: The script now does not delete any data from `working-dir` if it fails.

kjezek (approved):

scripts/db-heal-test.py
```diff
@@ -0,0 +1,298 @@
...
+genesis = os.path.join(working_dir, 'test_genesis.dat')
+
+print("Your settings:")
+print(f"\tNumber of iterations: {number_of_iterations}.")
+print(f"\tSync time before kill: {window} seconds.")
+print(f"\tCheckpoint granularity: {checkpoint_granularity} blocks.")
+
+
+# Function which checks programs return code, if program failed, log is printed and True is returned.
+def has_program_failed(code, log):
```
HerbertJordan, line 108: nit: `code` => `return_code` or `result`

scripts/db-heal-test.py
```diff
@@ -0,0 +1,298 @@
...
+    if code != 0:
+        log.close()
+        with open(carmen_log_file, 'r') as l:
```
HerbertJordan, line 111: The log file path should be a parameter or taken from `log` if possible; at one point in the program the file is actually named `carmen-info.log` not `carmen.log` as defined in the global.

scripts/db-heal-test.py
```diff
@@ -0,0 +1,298 @@
...
+            text = l.read()
+            print(text)
+        return True
+    return False
+
+
+# Function which checks every line added to aida_log_file and behaves accordingly to the line.
+def check_aida_log(sleep_time: int, checkpoint: int):
+    start = 0.0
+    with open(aida_log_file, 'r') as f:
```
HerbertJordan, line 121: nit: like for the `carmen_log_file` used in the function above, the `aida_log_file` should also be a parameter of the function

scripts/db-heal-test.py
```diff
@@ -0,0 +1,298 @@
...
+        while True:
+            line = f.readline()
+            if not line:
+                time.sleep(0.1)
+                continue
+            if start > 0 and time.time() - start >= sleep_time:
+                print("Interrupting...")
+                process.terminate()
```
HerbertJordan, line 129: Python's dynamic scoping makes it difficult to identify the `process` that is actually terminated here => it should be a parameter.

scripts/db-heal-test.py
```diff
@@ -0,0 +1,298 @@
...
+            stdout=cl,
+            stderr=cl,
+            cwd=carmen_path)
+        if has_program_failed(r.returncode, cl):
+            return -1
+
+    with open(log, 'r') as cl:
+        info_checkpoint = cl.readlines()[-1]
+        # Return last word which is the block number
+        cp = info_checkpoint.split()[-1]
```
HerbertJordan, line 160: Assuming that the checkpoint block is the last word in the last line is rather brittle. While debugging I added debug messages to some code, that added output to the `info` command, and the process failed.

I would suggest to use a regex over the entire output to identify something like "Checkpoint: ([\d])" and then use the match to extract the number.

HerbertJordan (changes requested):
Looks good, thanks for developing this.

I have one request to improve resilience towards future changes and a few code style comments. Please have a look.

HerbertJordan (approved):

HerbertJordan (approved):

PR #988, opened 2024-08-01
Add recovery support for Archive

HerbertJordan:
This PR integrates checkpoint and recovery support into MPT archives. The main additions are:
- the periodic creation of checkpoints when adding blocks to an Archive
- the support for restoring arbitrary block heights preceding the last checkpoint
- the support for restoring corrupted archives using the MPT tool

This PR contributes to #946 

How to review this PR:
- the main additions are in the `database/mpt/archive_trie.go` file and the associated test file
- the feature of forcing the removal of a lock got added to `database/mpt/directory_lock.go` which is independent from the rest (if requested, it can be moved into its own PR)
- the command line support for restoring archives is added to `database/mpt/tool/reset.go` (could also be moved into its own PR if requested)
- all other files needed to be modified due to the addition of the `ArchiveConfig` struct to the factory function of the MPT archive

### DIFF
--- a/go/database/mpt/archive_trie.go
+++ b/go/database/mpt/archive_trie.go
@@ -119,7 +119,7 @@ func OpenArchiveTrie(
 
 	checkpointInterval := archiveConfig.CheckpointInterval
 	if checkpointInterval <= 0 {
-		checkpointInterval = 10_000
+		checkpointInterval = 1000
 	}
 
 	return &ArchiveTrie{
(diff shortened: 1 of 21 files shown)


go/database/mpt/archive_trie.go
```diff
@@ -352,6 +412,80 @@ func (a *ArchiveTrie) Close() error {
...
+		return fmt.Errorf("failed to get checkpoint height: %v", err)
+	}
+	if block > uint64(checkpointHeight) {
+		return fmt.Errorf("block %d is beyond the last checkpoint height of %d", block, checkpointHeight)
+	}
+
+	// Mark this directory as dirty at least for the duration of the recovery.
+	if err := markDirty(directory); err != nil {
+		return fmt.Errorf("failed to mark directory %s as dirty: %w", directory, err)
+	}
```
petr-hanzl, line 460: Shouldnt it already be marked as `dirty`, since at this point, we are restoring it?

petr-hanzl (dismissed):

go/database/mpt/archive_trie.go
HerbertJordan: This can not be known. There is nothing that checks whether the given directory covers indeed a corrupted Archive. It could be a healthy Archive that should just be reverted to an earlier state.

petr-hanzl (approved):

PR #989, opened 2024-08-01
Specify import errors.

petr-hanzl:
This PR specifies errors when running incorrect command+genesis combination.

### DIFF
--- a/go/database/mpt/io/archive.go
+++ b/go/database/mpt/io/archive.go
@@ -227,7 +227,11 @@ func importArchive(liveDbDir, archiveDbDir string, in io.Reader) (err error) {
 	if _, err := io.ReadFull(in, buffer); err != nil {
 		return err
 	} else if !bytes.Equal(buffer, archiveMagicNumber) {
-		return fmt.Errorf("invalid format, wrong magic number")
+		// Specify error if incorrect genesis is passed
+		if bytes.Contains(buffer, stateMagicNumber) {
+			return fmt.Errorf("incorrect genesis+command combination\n your genesis is meant to be used with import-live")
+		}
+		return errors.New("invalid format, unknown magic number")
 	}
 
 	// Check the version number.
--- a/go/database/mpt/io/archive.go
+++ b/go/database/mpt/io/archive.go
@@ -227,9 +227,9 @@ func importArchive(liveDbDir, archiveDbDir string, in io.Reader) (err error) {
 	if _, err := io.ReadFull(in, buffer); err != nil {
 		return err
 	} else if !bytes.Equal(buffer, archiveMagicNumber) {
-		// Specify error if incorrect genesis is passed
+		// Provide an explicit warning to the user if instead of an archive dump a live-db dump was provided
 		if bytes.Contains(buffer, stateMagicNumber) {
-			return fmt.Errorf("incorrect genesis+command combination\n your genesis is meant to be used with import-live")
+			return fmt.Errorf("incorrect input data format; use the `import` or `import-live-db` sub-command with this type of data")
 		}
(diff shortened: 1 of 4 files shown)


go/database/mpt/io/archive.go
```diff
@@ -227,7 +227,11 @@ func importArchive(liveDbDir, archiveDbDir string, in io.Reader) (err error) {
 	if _, err := io.ReadFull(in, buffer); err != nil {
 		return err
 	} else if !bytes.Equal(buffer, archiveMagicNumber) {
-		return fmt.Errorf("invalid format, wrong magic number")
+		// Specify error if incorrect genesis is passed
+		if bytes.Contains(buffer, stateMagicNumber) {
+			return fmt.Errorf("incorrect genesis+command combination\n your genesis is meant to be used with import-live")
+		}
```
HerbertJordan, line 233: I would like to avoid the term `genesis` here, to reduce the confusion with what is considered a `genesis` file in the full Sonic client. What we are dealing with here is only a small part of it, and so far I think we are not referring to it in the code as `genesis`.

Suggestion:
```suggestion
		// Provide an explicit warning to the user if instead of an archive dump a live-db dump was provided
		if bytes.Contains(buffer, stateMagicNumber) {
			return fmt.Errorf("incorrect input data format; use the `import` or `import-live-db` sub-command with this type of data")
		}
```

go/database/mpt/io/live.go
```diff
@@ -229,7 +229,11 @@ func runImport(directory string, in io.Reader, config mpt.MptConfig) (root mpt.N
 	if _, err := io.ReadFull(in, buffer); err != nil {
 		return root, hash, err
 	} else if !bytes.Equal(buffer, stateMagicNumber) {
-		return root, hash, fmt.Errorf("invalid format, wrong magic number")
+		// Specify error if incorrect genesis is passed
+		if bytes.Contains(buffer, archiveMagicNumber[:len(stateMagicNumber)]) {
+			return root, hash, fmt.Errorf("incorrect genesis+command combination\n your genesis is meant to by used with either import or import-archive")
```
HerbertJordan, line 234:
```suggestion
		// Provide an explicit warning to the user if instead of a live state dump an archive dump was provided
		if bytes.Contains(buffer, archiveMagicNumber[:len(stateMagicNumber)]) {
			return root, hash, fmt.Errorf("incorrect input data format\n use the `import-archive` sub-command  with this type of data")
```

HerbertJordan (changes requested):
I would like to avoid the term `genesis` in the Database context. We can discuss the reasons offline if you have questions.

go/database/mpt/io/live.go
```diff
@@ -229,7 +229,11 @@ func runImport(directory string, in io.Reader, config mpt.MptConfig) (root mpt.N
 	if _, err := io.ReadFull(in, buffer); err != nil {
 		return root, hash, err
 	} else if !bytes.Equal(buffer, stateMagicNumber) {
-		return root, hash, fmt.Errorf("invalid format, wrong magic number")
+		// Provide an explicit warning to the user if instead of a live state dump an archive dump was provided
+		if bytes.Contains(buffer, archiveMagicNumber[:len(stateMagicNumber)]) {
+			return root, hash, fmt.Errorf("incorrect input data format\n use the `import-archive` sub-command  with this type of data")
```
HerbertJordan, line 234: Not sure a new line `\n` in an error string is a good idea ... formatting should be left to the UI part.

HerbertJordan (dismissed):

go/database/mpt/io/live.go
petr-hanzl: removed.

HerbertJordan (approved):

PR #996, opened 2024-08-06
Eliminate ineffective cache caching in Archive block state export

HerbertJordan:
This PR improves the block-state exporting performance from an Archive. Before this fix, exporting block 10M from an Archive including 10M blocks took 5.5 minutes. With this change, this time got reduced to 2 minutes.

In the pprof profile the issue was clearly visible:
image

After this fix, this part of the overall processing is no longer visible in the profile.

In the old version of this code, the full code map was copied for each individual lookup, as the function
```
func (e exportableArchiveTrie) GetCodeForHash(hash common.Hash) []byte
```
accepts `e` by value. Thus, the fetched code map is dropped after each call.

Furthermore, there is no need to cache codes in the first place, as those are also cached by the Archive itself.

This is part of #997

### DIFF
--- a/go/database/mpt/archive_trie.go
+++ b/go/database/mpt/archive_trie.go
@@ -292,7 +292,11 @@ func (a *ArchiveTrie) GetCode(block uint64, account common.Address) (code []byte
 	if err != nil {
 		return nil, a.addError(err)
 	}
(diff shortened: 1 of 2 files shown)


thaarok (approved):

petr-hanzl (approved):
Good catch! Thanks for improving the export.

PR #999, opened 2024-08-14
Pre-allocate slice for Archive roots

HerbertJordan:
This PR adds a minor performance fix for opening Archives. Instead of gradually growing the root slice, it is allocated once while loading root node data from the disk.

This reduced loading time for the command
```
go run ./database/mpt/tool info /mnt/sonicdb/mainnet/carmen/archive
```
for an Archive with 88M blocks from 20 to 10 seconds.

The reduction can be seen in the CPU profile. Before:
image

After:
image


kjezek (approved):
thanks for this improvement

PR #1001, opened 2024-08-22
Kjezek/improve export tools ux

kjezek:
This PR improves user experience with MPT export tools. It 

1. adds progress tracking into the output
2. enables ctrl+c at some places, where it was missing - in particular export of codes of liveDB

Furthermore, this PR introduces a simple logging utility that unifies output for export tools.

### DIFF
--- a/go/database/mpt/io/log.go
+++ b/go/database/mpt/io/log.go
@@ -1,0 +1,72 @@
+// Copyright (c) 2024 Fantom Foundation
+//
+// Use of this software is governed by the Business Source License included
+// in the LICENSE file and at fantom.foundation/bsl11.
+//
+// Change Date: 2028-4-16
+//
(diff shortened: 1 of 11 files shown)


go/database/mpt/io/log.go
petr-hanzl, line 1: Should we move `log` to under `common` package into `log` package? It might be usable somewhere else in the future...?

petr-hanzl (approved):

go/database/mpt/io/log.go
kjezek: I would probably wait with this. At the moment the implementation is the simplest possible, i.e. I would wait with promoting it to a generally usable library. But once we need more functionality and more places to adopt it, then yes, I would move it there.

petr-hanzl (approved):

PR #1003, opened 2024-08-23
extends log output for import tools

kjezek:
this PR extends log output of database import tools. It now prints either the number of processed Accounts for liveDB or the number of processed blocks for Archive

### DIFF
--- a/go/database/mpt/io/archive.go
+++ b/go/database/mpt/io/archive.go
@@ -203,29 +203,29 @@ func ExportArchive(ctx context.Context, logger *Log, directory string, out io.Wr
 	return archive.Close()
 }
 
-func ImportArchive(directory string, in io.Reader) error {
+func ImportArchive(logger *Log, directory string, in io.Reader) error {
 	// check that the destination directory is an empty directory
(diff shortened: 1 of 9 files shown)


go/database/mpt/io/archive.go
```diff
@@ -267,6 +267,7 @@ func importArchive(liveDbDir, archiveDbDir string, in io.Reader) (err error) {
 	}()
 
 	// Restore the archive from the input file.
+	progress := logger.NewProgressTracker("processed %d blocks, rate: %f blocks/s", 1_000_000)
```
petr-hanzl, line 270: nit: progress -> tracker?
Also isnt 1m blocks a bit too shallow? I think 100k is just enough

kjezek: when importing current blockchain it prints about 80 lines, if the window was set to 100k, it would print 800 lines. What is better, I do not know ...

petr-hanzl: I guess it does not really matter. I will leave the decision on you.

petr-hanzl (dismissed):

petr-hanzl (approved):

PR #1004, opened 2024-08-27
Introduce method to export a LiveDB dump from a running Archive

kjezek:
this PR introduces a method to export a livedb view from an archive, that is being currently online, accepting new blocks. This method accesses the archive independently on the main program by do not checking if the directory is locked, and allowing for accessing directly the directory.

### DIFF
--- a/go/database/mpt/io/live.go
+++ b/go/database/mpt/io/live.go
@@ -54,15 +54,6 @@ const (
 	EthereumHash = HashType(0)
 )
 
-// NewExportableArchiveTrie allows visiting mpt.ArchiveTrie at given block
-// and getting its properties such as Code Hashes or Root Hash.
-func NewExportableArchiveTrie(trie *mpt.ArchiveTrie, block uint64) mptStateVisitor {
-	return exportableArchiveTrie{
-		trie:  trie,
-		block: block,
-	}
-}
-
 // mptStateVisitor is an interface for Tries that allows for visiting the Trie nodes
 // and furthermore getting its properties such as a root hash and contract codes.
 type mptStateVisitor interface {
@@ -135,6 +126,30 @@ func ExportBlockFromArchive(ctx context.Context, logger *Log, directory string, 
 
 	defer archive.Close()
 	_, err = ExportLive(ctx, logger, exportableArchiveTrie{trie: archive, block: block}, out)
+	return err
+}
+
+// ExportBlockFromOnlineArchive exports a LiveDB dump for a single given block from an Archive.
+// This method exports from an online archive, i.e, an archive that is being updated with new blocks.
+// To ensure the exported data is up-to-date, this method flushes the archive to disk before exporting.
+// Expected usage is, for instance, the creation of database dumps once in many blocks to backup the state.
+func ExportBlockFromOnlineArchive(ctx context.Context, logger *Log, archive *mpt.ArchiveTrie, out io.Writer, block uint64) error {
+	logger.Printf("exporting block %d from online archive", block)
+	defer func() {
+		logger.Printf("exported block %d from online archive", block)
+	}()
+
+	logger.Printf("flushing archive")
+	// before doing anything, flush the archive to ensure the data is up-to-date
+	if err := archive.Flush(); err != nil {
+		return err
+	}
+
+	logger.Printf("exporting")
+	_, err := ExportLive(ctx, logger, exportableArchiveTrie{
+		trie:  archive,
+		block: block,
+	}, out)
 	return err
 }
 
--- a/go/database/mpt/io/live_test.go
+++ b/go/database/mpt/io/live_test.go
@@ -291,17 +291,17 @@ func TestIO_ExportBlockFromArchive(t *testing.T) {
 		t.Fatalf("failed to create archive: %v", err)
 	}
 	const (
-		M = 10
-		N = 3
+		Blocks   = 10
+		Accounts = 3
 	)
 
 	var expectedHashes []common.Hash
 
-	for i := 0; i < M; i++ {
+	for i := 0; i < Blocks; i++ {
 		code := []byte{1, 2, 3, byte(i)}
 		u := uint64(i)
 		update := common.Update{}
-		for j := 0; j < N; j++ {
+		for j := 0; j < Accounts; j++ {
 			newAddr := common.AddressFromNumber(j)
 
 			update.CreatedAccounts = append(update.CreatedAccounts, newAddr)
@@ -327,7 +327,7 @@ func TestIO_ExportBlockFromArchive(t *testing.T) {
 		t.Fatalf("failed to close archive archive: %v", err)
 	}
 
-	for i := 0; i < M; i++ {
+	for i := 0; i < Blocks; i++ {
 		// Export live database from archive.
 		buffer := new(bytes.Buffer)
 		if err := ExportBlockFromArchive(context.Background(), NewLog(), sourceDir, buffer, uint64(i)); err != nil {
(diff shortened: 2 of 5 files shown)


HerbertJordan (approved):
Looks good, thanks for implementing this!

go/database/mpt/io/live.go
```diff
@@ -138,6 +130,33 @@ func ExportBlockFromArchive(ctx context.Context, logger *Log, directory string,
 	return err
 }
 
+// ExportBlockFromOnlineArchive exports LiveDB view for a single given block from an Archive.
+// This method exports from the online archive, i.e, an archive that is being updated with new blocks.
+// To ensure the exported data are up-to-date, this method flushes archive to disk before exporting.
+// For this reason, this method should not be called often, as it would hammer performance.
+// Expected usage is, for instance, the creation of database dump once in many blocks to backup the state.
+func ExportBlockFromOnlineArchive(ctx context.Context, archive *mpt.ArchiveTrie, out io.Writer, block uint64) error {
+	logger := NewLog()
```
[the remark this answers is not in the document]
kjezek: good idea, changed this way. Furthermore the logger was changed that it can be nil

go/database/mpt/archive_trie.go
```diff
@@ -549,6 +551,11 @@ func (a *ArchiveTrie) addError(err error) error {
 	return a.archiveError
 }
 
+// Directory returns the directory where the archive is stored on disk.
+func (a *ArchiveTrie) Directory() string {
```
petr-hanzl, line 555:
```suggestion
// GetDirectory returns the path to directory where the archive is stored on disk.
func (a *ArchiveTrie) GetDirectory() string {
```

go/database/mpt/io/log.go
```diff
@@ -30,9 +30,11 @@ func NewLog() *Log {
 
 // Print logs a message that includes the time elapsed since the start of the program.
 func (l *Log) Print(msg string) {
-	now := time.Now()
-	t := uint64(now.Sub(l.start).Seconds())
-	l.logger.Printf("[t=%4d:%02d] - %s\n", t/60, t%60, msg)
+	if l != nil {
```
petr-hanzl, line 33: How can l be nil?

go/state/gostate/archive_state.go
```diff
@@ -155,14 +155,19 @@ func (s *ArchiveState) Export(ctx context.Context, out io.Writer) (common.Hash,
 		return common.Hash{}, state.ExportNotSupported
 	}
 
-	exportableTrie := mptio.NewExportableArchiveTrie(trie, s.block)
-	rootHash, err := mptio.ExportLive(ctx, mptio.NewLog(), exportableTrie, out)
+	err := mptio.ExportBlockFromOnlineArchive(ctx, nil, trie, out, s.block)
```
petr-hanzl, line 158: You can return the hash from this func and remove it from getting on line 164.

go/database/mpt/archive_trie.go
kjezek: It is idiomatic (albeit not followed everywhere) to name getters without `Get..()`  in golang: https://go.dev/doc/effective_go#Getters

go/state/gostate/archive_state.go
kjezek: I am not fan of the parent method design at the first place.  I think API should provide the simplest (most straightforward) use-cases.  I want export - there is a method to export, I need a root hash - there is a method to get it (we already have it),

but having a method that "exports a database and as a side effect gives hash of the original database" is too complicated - it servers two purposes. 

As a consequence, there will be clients that must ignore one of the results because it does not serve their needs.  

I.e. I do not want to corrupt API of `ExportBlockFromOnlineArchive` to fit this particular callsite

go/database/mpt/io/log.go
petr-hanzl: Yeah, I think the interface is what we normally use to create a 'Nil' instance of something.

HerbertJordan (approved):

go/database/mpt/io/log.go
```diff
@@ -30,9 +30,11 @@ func NewLog() *Log {
...
+		now := time.Now()
+		t := uint64(now.Sub(l.start).Seconds())
+		l.logger.Printf("[t=%4d:%02d] - %s\n", t/60, t%60, msg)
+	}
```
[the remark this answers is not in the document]
kjezek: ok, have no strong opinion about it

go/database/mpt/io/live.go
```diff
@@ -138,6 +130,32 @@ func ExportBlockFromArchive(ctx context.Context, logger *Log, directory string,
 	return err
 }
 
+// ExportBlockFromOnlineArchive exports LiveDB view for a single given block from an Archive.
+// This method exports from the online archive, i.e, an archive that is being updated with new blocks.
+// To ensure the exported data are up-to-date, this method flushes archive to disk before exporting.
+// For this reason, this method should not be called often, as it would hammer performance.
+// Expected usage is, for instance, the creation of database dump once in many blocks to backup the state.
```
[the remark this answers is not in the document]
kjezek: ok, I removed this part:  `For this reason, this method should not be called often, as it  ...`

go/state/gostate/archive_state.go
```diff
@@ -155,14 +155,19 @@ func (s *ArchiveState) Export(ctx context.Context, out io.Writer) (common.Hash,
...
 	}
 
-	return rootHash, s.archiveError
+	rootHash, err := trie.GetHash(s.block)
+	if err != nil {
+		s.archiveError = errors.Join(s.archiveError, err)
+		return common.Hash{}, s.archiveError
+	}
+
+	return rootHash, nil
```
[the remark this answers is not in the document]
kjezek: I think in will be always nil (I changed this method a bit)

go/database/mpt/io/live_test.go
```diff
@@ -366,6 +366,81 @@ func TestIO_ExportBlockFromArchive(t *testing.T) {
...
+	}
+	defer func() {
+		if err := archive.Close(); err != nil {
+			t.Fatalf("failed to close archive archive: %v", err)
+		}
+	}()
+
+	const (
+		M = 10
+		N = 3
```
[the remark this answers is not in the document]
kjezek: ok, renamed to Blocks, and Accounts.

go/database/mpt/io/live.go
```diff
@@ -75,8 +66,9 @@ type mptStateVisitor interface {
 }
 
 type exportableArchiveTrie struct {
-	trie  *mpt.ArchiveTrie
-	block uint64
+	directory string
+	trie      *mpt.ArchiveTrie
```
HerbertJordan, line 70: If the `ArchiveTrie` now contains the `directory` and it can be obtained from it using its `Directory()` member function, storing the `directory` explicitly here is not really necessary, right?

go/database/mpt/io/live.go
```diff
@@ -138,6 +130,31 @@ func ExportBlockFromArchive(ctx context.Context, logger *Log, directory string,
...
+
+	logger.Printf("flushing archive")
+	// before doing anything, flush the archive to ensure the data is up-to-date
+	if err := archive.Flush(); err != nil {
+		return err
+	}
+
+	logger.Printf("exporting")
+	_, err := ExportLive(ctx, logger, exportableArchiveTrie{
+		directory: archive.Directory(),
```
HerbertJordan, line 151: It seems like the `directory` field is nowhere used but here.

go/database/mpt/io/live.go
```diff
@@ -138,6 +130,31 @@ func ExportBlockFromArchive(ctx context.Context, logger *Log, directory string,
 	return err
 }
 
+// ExportBlockFromOnlineArchive exports LiveDB view for a single given block from an Archive.
+// This method exports from the online archive, i.e, an archive that is being updated with new blocks.
+// To ensure the exported data are up-to-date, this method flushes archive to disk before exporting.
+// Expected usage is, for instance, the creation of database dump once in many blocks to backup the state.
```
HerbertJordan, line 136:
```suggestion
// ExportBlockFromOnlineArchive exports a LiveDB dump for a single given block from an Archive.
// This method exports from an online archive, i.e, an archive that is being updated with new blocks.
// To ensure the exported data is up-to-date, this method flushes the archive to disk before exporting.
// Expected usage is, for instance, the creation of database dumps once in many blocks to backup the state.
```

go/state/gostate/archive_state.go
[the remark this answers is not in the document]
HerbertJordan: It would just be defensive coding; It would not hurt to keep returning `s.archiveError`, even if it is always `nil`. Somebody may change something in this function in the future and overlook that in the end `nil` is returned.

HerbertJordan (dismissed):

go/database/mpt/io/live.go
kjezek: Anyway, I noticed that this directory is needed in another PR, so I will not include this change here

go/state/gostate/archive_state.go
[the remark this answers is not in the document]
kjezek: ok, makes sense, will change it this way

HerbertJordan (approved):

PR #1006, opened 2024-08-27
Use parallelized tree traversal for LiveDB dump from Archives

kjezek:
This PR introduces parallel reading from archive.


go/database/mpt/io/parallel_visit.go
```diff
@@ -0,0 +1,429 @@
...
+	const NumWorker = 16
+
+	// open all sources for the number of workers and track possible errors
+	sources := make([]NodeSource, 0, NumWorker)
+	openingErrors := make([]error, 0, NumWorker)
+	for i := 0; i < NumWorker; i++ {
+		source, err := sourceFactory.Open()
+		sources = append(sources, source)
+		openingErrors = append(openingErrors, err)
+	}
```
HerbertJordan, line 90: Why did you factor out the source creation from the individual workers? Creating them at the beginning of each worker seemed to be less code and complexity; also, closing them at the end of the worker's go-routine seems to be less code ...

go/database/mpt/io/parallel_visit.go
```diff
@@ -0,0 +1,429 @@
...
+
+	defer func() {
+		// only log errors when closing the sources
+		// as data are opened parallel to the main program
+		// and thus the error should not invalidate the result
+		for _, source := range sources {
+			if source != nil {
+				if err := source.Close(); err != nil {
```
HerbertJordan, line 98: You are not blocking for all of workers you are creating below to have finished before closing the sources. Thus, you may close a source while a worker is still actively using it, which may not be thread safe and cause problems.

go/database/mpt/io/parallel_visit.go
```diff
@@ -0,0 +1,429 @@
...
+					log.Printf("failed to close source: %v", err)
```
HerbertJordan, line 99: These errors get lost; they should be reported as part of the result of this call.

go/database/mpt/io/parallel_visit.go
```diff
@@ -0,0 +1,429 @@
...
+		responsesMutex.Unlock()
+
+		if res.err != nil {
+			return res.err
+		}
+
+		switch visitor.Visit(res.node, mpt.NodeInfo{Id: cur}) {
+		case mpt.VisitResponseAbort:
+			return nil
+		case mpt.VisitResponsePrune:
```
HerbertJordan, line 200: I do not think that this algorithm is supporting visitors deciding to prune the visiting process. This should be removed and documented -- or even the type of visitor accepted by this function should be changed.

The reason is: if a visitor decides to "Prune" a part of the tree, the workers do not know about this. They will still fetch all the nodes of that part of the tree, and those nodes are not consumed. This can easily end up in a OOM crash.

go/database/mpt/io/parallel_visit.go
```diff
@@ -0,0 +1,429 @@
...
+					stack = append(stack, id)
+				}
+			}
+		}
+	}
+
+	return nil
+}
+
+type position struct {
```
HerbertJordan, line 230: nit: needs a line of documentation; something like "position addresses a node within a tree by listing the path from the root node to the respective node" ...

go/database/mpt/io/parallel_visit.go
```diff
@@ -0,0 +1,429 @@
...
+}
+
+func (f *nodeSourceFactory) Open() (NodeSource, error) {
+	accounts, err := openSource[mpt.AccountNode](f.directory, "accounts", mpt.AccountNodeWithPathLengthEncoderWithNodeHash{})
+	if err != nil {
+		return nil, err
+	}
+	branches, err := openSource[mpt.BranchNode](f.directory, "branches", mpt.BranchNodeEncoderWithNodeHash{})
+	if err != nil {
+		return nil, err
```
HerbertJordan, line 335: If opening the `branches` failed, `accounts` should be closed. The same for the sources below.

go/database/mpt/io/parallel_visit.go
```diff
@@ -0,0 +1,429 @@
...
+}
+
+type source[T any] struct {
+	file    utils.OsFile
+	encoder stock.ValueEncoder[T]
+	buffer  []byte
+}
+
+func openSource[T any](directory, name string, encoder stock.ValueEncoder[T]) (source[T], error) {
+	file, err := os.Open(fmt.Sprintf("%s/%s/values.dat", directory, name))
```
HerbertJordan, line 399: If we want to play it safe, the `openSource` function should check whether the directory contains a `stock` with the proper value and index type. This would be easiest by reusing code from the file stock.

Probably this code for the "source" should be moved to the file stock package and renamed to something like a "StockReader" or something like that ...

go/database/mpt/state.go
```diff
@@ -345,7 +345,7 @@ func (s *MptState) Apply(block uint64, update common.Update) (archiveUpdateHints
 	return hints, err
 }
 
-func (s *MptState) Visit(visitor NodeVisitor) error {
+func (s *MptState) Visit(visitor NodeVisitor, _ bool) error {
```
HerbertJordan, line 348: This extra bool (likely the "cutAtAccount") should not be ignored. Actually, it should not be added at all, since the "NodeVisitor" tells the `MptState` what to include and what not by returning visitor responses like "abort", "continue" or "prune".

Instead of updating the this function, a wrapper will be required, customizing the visitor to "prune" when reaching an account, defined by the second flag.

go/state/gostate/archive_state.go
```diff
@@ -155,14 +155,21 @@ func (s *ArchiveState) Export(ctx context.Context, out io.Writer) (common.Hash,
 		return common.Hash{}, state.ExportNotSupported
 	}
 
-	exportableTrie := mptio.NewExportableArchiveTrie(trie, s.block)
-	rootHash, err := mptio.ExportLive(ctx, mptio.NewLog(), exportableTrie, out)
+	err := mptio.ExportBlockFromOnlineArchive(ctx, trie, out, s.block)
+	if err != nil {
+		// export opens the trie parallel to the current program,
+		// possible error is sent to the caller but should not corrupt
+		// the database on error.
```
HerbertJordan, line 162: It might ... if the error happened during the initial flush ... would this be recorded by the flush?

go/state/gostate/archive_state.go
```diff
@@ -155,14 +155,21 @@ func (s *ArchiveState) Export(ctx context.Context, out io.Writer) (common.Hash,
...
+	}
+
+	rootHash, err := trie.GetHash(s.block)
 	if err != nil {
 		s.archiveError = errors.Join(s.archiveError, err)
 		return common.Hash{}, s.archiveError
 	}
 
-	return rootHash, s.archiveError
+	return rootHash, nil
```
HerbertJordan, line 172: Is the error really `nil` in this case? I think it should remain to be the `s.archiveError` ...

HerbertJordan (changes requested):
This is a large PR and I think it would be best to factor out some elements. In particular the heap could be added as a separate PR. Another thing that could be factored out into its own PR could be the implementation of the node source. Wdyt?

go/state/gostate/archive_state.go
kjezek: good point. It seems it was not checked, I also found a few places where I think error should be stored as well but was not. I will add it to the other PR to simplify this one

go/state/gostate/archive_state.go
kjezek: I think it is - if it was not nil, this method would return at the very beginning.

go/state/gostate/archive_state.go
kjezek: I created a new PR #1007 for this. I think this check was missing in Flush. when it is added I think we can keep the code here as described in the comment

go/database/mpt/io/live.go
```diff
@@ -134,7 +131,38 @@ func ExportBlockFromArchive(ctx context.Context, logger *Log, directory string,
...
+		directory: directory,
+		trie:      archive,
+		block:     block,
+	}, out)
+	return err
+}
+
+// ExportBlockFromOnlineArchive exports LiveDB view for a single given block from an Archive.
+// This method exports from the online archive, i.e, an archive that is being updated with new blocks.
+// To ensure the exported data are up-to-date, this method flushes archive to disk before exporting.
```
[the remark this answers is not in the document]
kjezek: will get fixed in the other PR

go/database/mpt/io/live.go
```diff
@@ -134,7 +131,38 @@ func ExportBlockFromArchive(ctx context.Context, logger *Log, directory string,
...
+// For this reason, this method should not be called often, as it would hammer performance.
+// Expected usage is, for instance, the creation of database dump once in many blocks to backup the state.
+func ExportBlockFromOnlineArchive(ctx context.Context, archive *mpt.ArchiveTrie, out io.Writer, block uint64) error {
+	logger := NewLog()
+	logger.Printf("exporting block %d from online archive", block)
```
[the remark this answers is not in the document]
kjezek: will get fixed in the other PR and rebased

go/database/mpt/io/live.go
```diff
@@ -54,33 +54,30 @@ const (
...
 
-func (e exportableArchiveTrie) Visit(visitor mpt.NodeVisitor) error {
-	return e.trie.VisitTrie(e.block, visitor)
+func (e exportableArchiveTrie) Visit(visitor mpt.NodeVisitor, cutAtAccounts bool) error {
+	root, err := e.trie.GetBlockRoot(e.block)
+	if err != nil {
+		return err
+	}
+
+	return visitAll(&nodeSourceFactory{directory: e.directory}, root, visitor, cutAtAccounts)
```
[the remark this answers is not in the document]
kjezek: I did this change to be able to test the code. When The factory was initialised inside the method, I could not mock it and could not inject errors

go/database/mpt/io/live.go
```diff
@@ -54,33 +54,30 @@ const (
...
-		block: block,
-	}
-}
-
 // mptStateVisitor is an interface for Tries that allows for visiting the Trie nodes
 // and furthermore getting its properties such as a root hash and contract codes.
 type mptStateVisitor interface {
 	// Visit allows for traverse the whole trie.
-	Visit(visitor mpt.NodeVisitor) error
+	Visit(visitor mpt.NodeVisitor, cutAtAccounts bool) error
```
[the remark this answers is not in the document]
kjezek: what about `pruneStorage` it sounds to fit terms we have been using already?

go/database/mpt/io/parallel_visit.go
kjezek: anyway, I changed it to open/close inside the go-routine, it needs some synchronisation to propagate errors, hope it is better, but still a bit complicated

go/database/mpt/io/parallel_visit.go
```diff
@@ -0,0 +1,429 @@
...
+	"sync"
+	"sync/atomic"
+)
+
+//go:generate mockgen -source parallel_visit.go -destination parallel_visit_mocks.go -package io
+
+// NodeSourceFactory is a factory for NodeSource instances.
+// It provides access to nodes side to another infrastructure
+// that already accesses the nodes.
+type NodeSourceFactory interface {
```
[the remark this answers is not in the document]
kjezek: ok, will make private

go/database/mpt/io/parallel_visit.go
kjezek: reverted back to open/close in the go-routine and sending errors via a channel

go/database/mpt/io/parallel_visit.go
kjezek: reverted back to open/close in the go-routine and sending errors via a channel

go/database/mpt/io/parallel_visit.go
kjezek: ok, will add

go/database/mpt/io/parallel_visit.go
kjezek: good point, fixed

go/database/mpt/io/parallel_visit.go
kjezek: reworked this way

go/database/mpt/io/parallel_visit.go
```diff
@@ -0,0 +1,429 @@
...
+	Open() (NodeSource, error)
+}
+
+// NodeSource is a source of nodes.
+// It provides access to nodes by their ids.
+type NodeSource interface {
```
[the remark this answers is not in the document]
kjezek: changed

go/database/mpt/io/parallel_visit.go
kjezek: ok, I created a new type of visitor. Also I needed to create a visitor to wrap the `cutAtAccounts` (`pruneStorage`) flag, i.e. it is a bit complex at the moment. Let me know what do you think.

go/state/gostate/archive_state.go
kjezek: oh, right. it must have leaked during merge

go/database/mpt/io/live.go
```diff
@@ -58,20 +58,28 @@ const (
...
 
-func (e exportableArchiveTrie) Visit(visitor mpt.NodeVisitor) error {
-	return e.trie.VisitTrie(e.block, visitor)
+func (e exportableArchiveTrie) Visit(visitor mpt.NodeVisitor, pruneStorage bool) error {
+	root, err := e.trie.GetBlockRoot(e.block)
+	if err != nil {
+		return err
+	}
+
+	return visitAll(&stockNodeSourceFactory{directory: e.trie.Directory()}, root, &noResponseMptNodeVisitor{visitor}, pruneStorage)
```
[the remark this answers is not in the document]
kjezek: as I wrote previously, I can do that, but then I cannot (or I do not how) test that a file cannot be opened or closed. Especially testing to `close` the file is tricky, I can easily test it by mocking  `stockNodeSourceFactory`, but when this factory is inside, I cannot mock it.

go/database/mpt/io/parallel_visit.go
```diff
@@ -0,0 +1,471 @@
...
+
+// noResponseMptNodeVisitor is a noResponseNodeVisitor implementation that routes to the mpt.NodeVisitor,
+// but ignores the results of each call to Visit.
+type noResponseMptNodeVisitor struct {
+	parent mpt.NodeVisitor
+}
+
+func (v *noResponseMptNodeVisitor) Visit(node mpt.Node, info mpt.NodeInfo) {
+	v.parent.Visit(node, info)
+}
```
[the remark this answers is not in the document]
kjezek: ok, makes sense

go/database/mpt/io/parallel_visit_test.go
```diff
@@ -0,0 +1,395 @@
...
+			t.Fatalf("failed to flush archive: %v", err)
+		}
+
+		blocks, _, err := trie.GetBlockHeight()
+		if err != nil {
+			t.Fatalf("failed to get block height: %v", err)
+		}
+
+		// iterate all nodes in the trie for all blocks
+		// and compare that the loaded nodes match the nodes in the trie
```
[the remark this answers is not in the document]
kjezek: ok, removed

go/database/mpt/io/parallel_visit_test.go
```diff
@@ -0,0 +1,395 @@
...
+					t.Parallel()
+					var position int
+					ctrl := gomock.NewController(t)
+					visitor := NewMocknoResponseNodeVisitor(ctrl)
+					visitor.EXPECT().Visit(gomock.Any(), gomock.Any()).DoAndReturn(func(_ mpt.Node, info mpt.NodeInfo) {
+						if got, want := info.Id, nodes[position]; got != want {
+							t.Errorf("expected node %v, got %v", want, got)
+						}
+						position++
+					}).AnyTimes()
```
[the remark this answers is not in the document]
kjezek: good idea, thanks

go/database/mpt/io/live.go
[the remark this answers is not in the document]
kjezek: ok done as discussed

go/state/gostate/archive_state.go
```diff
@@ -14,12 +14,12 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/Fantom-foundation/Carmen/go/database/mpt"
+	mptio "github.com/Fantom-foundation/Carmen/go/database/mpt/io"
```
HerbertJordan, line 18: nit: this looks like an unrelated change in for this PR; maybe this file change should be removed

go/database/mpt/io/parallel_visit.go
```diff
@@ -0,0 +1,462 @@
...
+	directory string,
+	root mpt.NodeId,
+	visitor noResponseNodeVisitor,
+	pruneStorage bool,
+) error {
+	return visitAllSources(&stockNodeSourceFactory{directory}, root, visitor, pruneStorage)
+}
+
+// visitAllSources is an internal implementation of visitAll that allows to provide a custom node source factory.
+func visitAllSources(
```
HerbertJordan, line 44: nit: this sounds a bit like sources are visited ... how about `visitAllWithSources`?

go/database/mpt/io/parallel_visit.go
```diff
@@ -0,0 +1,462 @@
...
+	}
+	if p.parent == nil {
+		return fmt.Sprintf("%d", p.pos)
+	}
+	return fmt.Sprintf("%s.%d", p.parent.String(), p.pos)
+}
+
+// Child creates a new position that is a child of the current position.
+// The new position is one step deeper than the current one.
+func (p *position) Child(step byte) *position {
```
HerbertJordan, line 291: nit: this was my mistake already in the prototype -- I guess like all other members of the position, this one should also be  private

HerbertJordan (approved):
Looks good, thanks for getting this done.

Once merged we should re-run the export experiment (I think I still have the setup active somewhere) and see whether the result is still correct and we retained the needed performance.

go/state/gostate/archive_state.go
kjezek: oh, right, was updated by my IDE

go/database/mpt/io/parallel_visit.go
kjezek: ok, will change

go/database/mpt/io/parallel_visit.go
kjezek: ok, changed

kjezek:
I tried to export block 80M from the Archive using this tool - the server with local SSH

It took **193min** -> **3.2h**

kjezek:
the resulting file has **38GB**

HerbertJordan (approved):

PR #1007, opened 2024-08-29
extends tracking of error states

kjezek:
this PR extends tracking of erroneous state of archive and forest. It seems it was forgotten to add this check at some newer code.


HerbertJordan (approved):
Thanks for having an eye on this and fixing it!

PR #1008, opened 2024-08-30
Add generic heap container

kjezek:
This PR introduces heap type


HerbertJordan (approved):
I just assume it is a copy of the one I copied from Serse.

kjezek:
> I just assume it is a copy of the one I copied from Serse.

yes, it is a copy

PR #1009, opened 2024-08-30
Introduces a read-only stock

kjezek:
This PR introduces a read-only stock

### DIFF
--- a/go/backend/stock/file/readonly.go
+++ b/go/backend/stock/file/readonly.go
@@ -1,0 +1,71 @@
+// Copyright (c) 2024 Fantom Foundation
+//
+// Use of this software is governed by the Business Source License included
+// in the LICENSE file and at fantom.foundation/bsl11.
+//
+// Change Date: 2028-4-16
+//
+// On the date above, in accordance with the Business Source License, use of
+// this software will be governed by the GNU Lesser General Public License v3.
+
+package file
+
+import (
+	"fmt"
+	"github.com/Fantom-foundation/Carmen/go/backend/stock"
+	"github.com/Fantom-foundation/Carmen/go/backend/utils"
+	"os"
+)
+
+type readonly[I stock.Index, V any] struct {
+	file    utils.OsFile
+	encoder stock.ValueEncoder[V]
+	buffer  []byte
+}
+
+// OpenReadOnlyStock opens a stock for reading only.
+// Multiple read-only stocks may be opened for the same directory at any time. However,
+// there are no guarantees against concurrent modifications introduced by Stock instances
+// accessing the same underlying files. It is the task of the user of this API to make sure that
+// no elements accessed by this reader are modified concurrently to guarantee consistency.
+func OpenReadOnlyStock[I stock.Index, V any](directory string, encoder stock.ValueEncoder[V]) (stock.ReadOnly[I, V], error) {
+	if err := VerifyStock[I](directory, encoder); err != nil {
+		return &readonly[I, V]{}, err
+	}
+
+	return openReadOnlyStock[I, V](directory, encoder)
+}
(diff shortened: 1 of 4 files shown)


go/backend/stock/stock.go
```diff
@@ -33,19 +48,14 @@ import (
...
 
-	// Get retrieves a value associated to an index value. If there is no such
-	// element, an undefined value is returned. Stock implementations are not
-	// required to retain information on valid and deleted indexes to
-	// efficiently distinguish valid from invalid accesses. The life-cycle of
-	// IDs is expected to be managed by the client code.
-	Get(I) (V, error)
-
-	// Updates the value associated ot the given index. The given index must be
+	// Set updates the value associated ot the given index. The given index must be
```
HerbertJordan, line 58:
```suggestion
	// Set updates the value associated to the given index. The given index must be
```

go/backend/stock/stock.go
```diff
@@ -21,6 +22,20 @@ import (
...
+// ReadOnly is a read-only interface to a Stock.
+// It allows for retrieving values associated with index values,
+// while it can neither create nor delete values.
+type ReadOnly[I Index, V any] interface {
+	// Get retrieves a value associated to an index value. If there is no such
+	// element, an undefined value is returned. Stock implementations are not
+	// required to retain information on valid and deleted indexes to
+	// efficiently distinguish valid from invalid accesses. The life-cycle of
+	// IDs is expected to be managed by the client code.
+	Get(I) (V, error)
```
HerbertJordan, line 34: nit: the documentation string is matching the `Stock` use case and would need to be adjusted here and there for the `ReadOnly` stock. It might be simpler to not do the embedding and define an independent interface.

go/backend/stock/file/readonly.go
```diff
@@ -0,0 +1,67 @@
...
+	"os"
+)
+
+type readonly[I stock.Index, V any] struct {
+	file    utils.OsFile
+	encoder stock.ValueEncoder[V]
+	buffer  []byte
+}
+
+// OpenReadOnlyStock opens a stock for reading only.
```
HerbertJordan, line 26:
```suggestion
// OpenReadOnlyStock opens a stock for reading only.
// Multiple read-only stocks may be opened for the same directory at any time. However,
// there are no guarantees against concurrent modifications introduced by Stock instances
// accessing the same underlying files. It is the task of the user of this API to make sure that
// no elements accessed by this reader are modified concurrently to guarantee consistency.
```

go/backend/stock/file/readonly_test.go
```diff
@@ -0,0 +1,151 @@
...
+	"go.uber.org/mock/gomock"
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+func TestReadOnlyFile_Get(t *testing.T) {
+	const Items = 3300
+	directory := t.TempDir()
+	s, err := openInitFileStock(directory, Items)
```
HerbertJordan, line 27: nit: wherever this function is defined, it should be renamed to `openAndInitFileStock` or `initFileStock` ; the way it is now looks like it is opening some special "InitFileStock" ...

HerbertJordan (approved):
Looks good, thanks!

go/backend/stock/stock.go
kjezek: ok, done this way. I basically removed the second part of the documentation, which was unrelated.

go/backend/stock/file/readonly_test.go
kjezek: ok, I picked `initFileStock`

HerbertJordan (approved):
Thanks for the updates, looks good!

PR #1012, opened 2024-09-05
Introduce VisitAccount, GetAccountInfo, CreateWitnessProof methods for tries

kjezek:
This PR introduces `VisitAccount`, `GetAccountInfo` and `CreateWitnessProof` method for tries. This method will be later needed for #947 to get nodes exclusively for accounts.

### DIFF
--- a/go/database/mpt/live_trie.go
+++ b/go/database/mpt/live_trie.go
@@ -15,6 +15,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/Fantom-foundation/Carmen/go/common/witness"
 	"os"
 	"unsafe"
 
@@ -156,6 +157,32 @@ func (s *LiveTrie) setHashes(hashes *NodeHashes) error {
 
 func (s *LiveTrie) VisitTrie(visitor NodeVisitor) error {
 	return s.forest.VisitTrie(&s.root, visitor)
+}
+
+// VisitAccountStorage visits the storage nodes of an account with the given address.
+// The visited nodes do not contain the account node itself.
+func (s *LiveTrie) VisitAccountStorage(address common.Address, visitor NodeVisitor) error {
+	var innerError error
+	// this visitor finds the account node with the given address
+	accountVisitor := MakeVisitor(func(node Node, info NodeInfo) VisitResponse {
+		switch n := node.(type) {
+		case *AccountNode:
+			if n.Address() == address {
+				// and then it sends the storage node to the visitor
+				_, innerError = n.visitStorage(s.forest, 0, visitor)
+			}
+			return VisitResponseAbort
+		}
+
+		return VisitResponseContinue
+	})
+
+	_, err := VisitPathToAccount(s.forest, &s.root, address, accountVisitor)
+	return errors.Join(innerError, err)
+}
+
+func (s *LiveTrie) CreateWitnessProof(addr common.Address, keys ...common.Key) (witness.Proof, error) {
+	return CreateWitnessProof(s.forest, &s.root, addr, keys...)
 }
 
 func (s *LiveTrie) Flush() error {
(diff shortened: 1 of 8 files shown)


go/database/mpt/live_trie.go
```diff
@@ -158,6 +159,32 @@ func (s *LiveTrie) VisitTrie(visitor NodeVisitor) error {
 	return s.forest.VisitTrie(&s.root, visitor)
 }
 
+// VisitAccount visits the storage nodes of an account with the given address.
+// The visited nodes contain the account node and all storage nodes of the account.
+func (s *LiveTrie) VisitAccount(address common.Address, visitor NodeVisitor) error {
```
HerbertJordan, line 164: Based on the description I would suggest to rename this function to `VisitAccountStorage` since it is the storage that is visited, not the account.

go/database/mpt/live_trie.go
```diff
@@ -158,6 +159,32 @@ func (s *LiveTrie) VisitTrie(visitor NodeVisitor) error {
...
+	var innerError error
+	// this visitor finds the account node with the given address
+	accountVisitor := MakeVisitor(func(node Node, info NodeInfo) VisitResponse {
+		switch n := node.(type) {
+		case *AccountNode:
+			if n.Address() == address {
+				// and then it sends the storage node to the visitor
+				_, innerError = n.Visit(s.forest, &n.storage, 0, visitor)
```
HerbertJordan, line 172: Is it important that the `AccountNode` is visited first before visiting the storage or is this irrelevant for the intended use case?

go/database/mpt/live_trie_test.go
```diff
@@ -190,6 +191,100 @@ func TestLiveTrie_Fail_Read_Data(t *testing.T) {
...
+			for i := 0; i < Addresses; i++ {
+				if err := trie.SetAccountInfo(common.AddressFromNumber(i), AccountInfo{Nonce: common.Nonce{1}}); err != nil {
+					t.Fatalf("failed to set account info: %v", err)
+				}
+			}
+
+			// there are no keys in the accounts
+			for i := 0; i < Addresses; i++ {
+				addr := common.AddressFromNumber(i)
+				if err := trie.VisitAccount(addr, MakeVisitor(func(node Node, _ NodeInfo) VisitResponse {
```
HerbertJordan, line 232: This part should somehow check that an account is indeed visited.

HerbertJordan (approved):

go/database/mpt/live_trie.go
kjezek: not sure I understand this comment, I have to find the account first, before I can start to iterate its storage, right?

go/database/mpt/live_trie.go
kjezek: ok, renamed

go/database/mpt/live_trie_test.go
kjezek: added

go/database/mpt/live_trie.go
HerbertJordan: The visitor getting passed to the `VisitAccountStorage` function will be called first with the account node, followed by all the branch, extension, and value nodes of the account's state. My question was whether this initial call with the account node is essential or whether just visiting the trie forming the Storage state is sufficient.

If the `AccountNode` needs to be visited for your use case, the function should probably be renamed to `VisitAccountAndStorage`. If only the storage is relevant, the documentation should be updated and the visit process should be started with the root node of the storage instead of with the `AccountNode` referencing the root.

HerbertJordan (approved):

go/database/mpt/live_trie.go
kjezek: oh, got it. No I did not need the account node in this use case, it was just the most straightforward to implement. Changed not to return the account node.

HerbertJordan (approved):

PR #1022, opened 2024-09-12
Add diagnoses to MPT tests and start fixing memory leaks

HerbertJordan:
This PR adds optional diagnostics support to the MPT unit test. The main motivation is to identify memory leaks in the tests.

To enable diagnostic, export the `MPT_DIAGNOSTIC_PORT` environment variable before running tests, similar to this:
```
MPT_DIAGNOSTIC_PORT=6060 go test -v ./database/mpt
```
 While the tests are running, the following command can be used to obtain a heap profile from the tests:
```
go tool pprof -http "localhost:8000" http://localhost:6060/debug/pprof/heap
```

In this profile, tests failing to free all their memory allocations (in particular failing to close all their MPT instances) can be identified.

### DIFF
--- a/go/database/mpt/archive_trie_test.go
+++ b/go/database/mpt/archive_trie_test.go
@@ -804,6 +804,7 @@ func TestArchiveTrie_Add_DuplicatedBlock(t *testing.T) {
 			if err != nil {
 				t.Fatalf("failed to create empty archive, err %v", err)
 			}
+			defer archive.Close()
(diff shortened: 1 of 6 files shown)


kjezek (approved):
I think it can be merged, I would like to see checking of errors from `Close` calls as well, but I can fix it as a follow up.

kjezek (approved):

PR #1024, opened 2024-09-13
check resource consumption archive+live tries tests

kjezek:
This PR resolves problem with non-release resources when running tests of Live and Archive Trie.

### DIFF
--- a/go/database/mpt/archive_trie_test.go
+++ b/go/database/mpt/archive_trie_test.go
@@ -2609,20 +2609,25 @@ func TestArchiveTrie_RestoreBlockHeight(t *testing.T) {
 		return nil
 	}
 
-	tests := map[string]func(*ArchiveTrie) error{
-		"clean_close": func(archive *ArchiveTrie) error {
+	tests := map[string]func(*testing.T, *ArchiveTrie) error{
+		"clean_close": func(t *testing.T, archive *ArchiveTrie) error {
 			return archive.Close()
 		},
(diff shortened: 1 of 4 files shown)


go/database/mpt/archive_trie_test.go
```diff
@@ -2586,8 +2620,9 @@ func TestArchiveTrie_RestoreBlockHeight(t *testing.T) {
 			)
 		},
 		"no_close": func(archive *ArchiveTrie) error {
+			// close the inner forest to release resources
 			// In order to allow the recovery to access the directory, the lock needs to be released
-			return archive.head.(*MptState).lock.Release()
+			return errors.Join(archive.head.(*MptState).lock.Release(), archive.forest.Close())
```
HerbertJordan, line 2625: This test is actually aiming to recover an archive that has not been properly close. By closing it now, the test setup is changed.

To fix this, you could pass `t *testing.T` to the this test setup function and register the close operation as a cleanup step.

HerbertJordan (changes requested):
Thanks for fixing those issues. Fixes look good except in one test case where I think an archive is getting closed to early. Please have a look.

go/database/mpt/archive_trie_test.go
kjezek: ok, updated. I think this version was also correct, because it closes only the inner `forest`, and the archive stays updated, but right, using the test clean-up phase makes it cleaner.

HerbertJordan (approved):
Thanks!

PR #1026, opened 2024-09-16
uses async MPT reading to speed up live db export

kjezek:
This PR uses async reading of MPT to speed up export of liveDB.

Experiment was taken to export liveDB from a block height beyond 80M blocks, a cloud VM with local SSD.

Originally, the export took approx. **1 day**, after the change, it takes approx. **2 hours**.

### DIFF
--- a/go/database/mpt/config.go
+++ b/go/database/mpt/config.go
@@ -10,6 +10,8 @@
 
 package mpt
 
+import "github.com/Fantom-foundation/Carmen/go/backend/stock"
+
 // MptConfig defines a set of configuration options for customizing the MPT
 // implementation. It is mainly intended to facilitate the accurate modeling
 // of Ethereum's MPT implementation (see schema 5) but may also be used for
@@ -82,6 +84,17 @@ func GetConfigByName(name string) (MptConfig, bool) {
 		}
 	}
 	return MptConfig{}, false
+}
+
+// GetEncoders returns the value encoders for the given MPT configuration.
+func (c MptConfig) GetEncoders() (
+	stock.ValueEncoder[AccountNode],
+	stock.ValueEncoder[BranchNode],
+	stock.ValueEncoder[ExtensionNode],
+	stock.ValueEncoder[ValueNode],
+) {
+
+	return getEncoder(c)
 }
 
 type HashStorageLocation bool
--- a/go/database/mpt/io/parallel_visit.go
+++ b/go/database/mpt/io/parallel_visit.go
@@ -33,11 +33,12 @@ import (
 // before calling this function to avoid reading out-dated or inconsistent data.
 func visitAll(
 	directory string,
+	config mpt.MptConfig,
 	root mpt.NodeId,
 	visitor noResponseNodeVisitor,
 	pruneStorage bool,
 ) error {
-	return visitAllWithSources(&nodeSourceHashWithNodesFactory{directory}, root, visitor, pruneStorage)
+	return visitAllWithSources(&stockNodeSourceFactory{directory, config}, root, visitor, pruneStorage)
 }
 
 // visitAllWithSources is an internal implementation of visitAll that allows to provide a custom node source factory.
@@ -374,14 +375,13 @@ type nodeSource interface {
 	get(mpt.NodeId) (mpt.Node, error)
 }
 
-// nodeSourceHashWithNodesFactory is a nodeSourceFactory implementation;
-// that creates stock backed sources to access nodes, using the node encoders
-// storing hashes with nodes.
-type nodeSourceHashWithNodesFactory struct {
+// stockNodeSourceFactory is a nodeSourceFactory implementation that creates stock backed sources to access nodes.
+type stockNodeSourceFactory struct {
 	directory string
+	config    mpt.MptConfig
 }
 
-func (f *nodeSourceHashWithNodesFactory) open() (nodeSource, error) {
+func (f *stockNodeSourceFactory) open() (nodeSource, error) {
 	var toClose []io.Closer
 	closeWithErr := func(err error) error {
 		for _, s := range toClose {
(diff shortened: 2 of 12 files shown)


go/database/mpt/io/parallel_visit.go
```diff
@@ -374,12 +374,14 @@ type nodeSource interface {
 	get(mpt.NodeId) (mpt.Node, error)
 }
 
-// stockNodeSourceFactory is a nodeSourceFactory implementation that creates stock backed sources to access nodes.
-type stockNodeSourceFactory struct {
+// nodeSourceHashWithNodesFactory is a nodeSourceFactory implementation;
```
HerbertJordan, line 377:
```suggestion
// nodeSourceHashWithNodesFactory is a nodeSourceFactory implementation
```

go/database/mpt/io/parallel_visit.go
```diff
@@ -420,6 +422,54 @@ func (f *stockNodeSourceFactory) open() (nodeSource, error) {
...
 }
 
+// nodeSourceHashWithChildNodesFactory is a nodeSourceFactory implementation;
+// that creates stock backed sources to access nodes, using the node encoders
+// storing hashes of children with parent nodes.
+type nodeSourceHashWithChildNodesFactory struct {
+	directory string
+}
+
+func (f *nodeSourceHashWithChildNodesFactory) open() (nodeSource, error) {
```
HerbertJordan, line 432: Instead of duplicating the code you could have parameterized the open code with an MptConfig and reuse the [getEncoder](https://github.com/Fantom-foundation/Carmen/blob/a7cc9b3cbcfef26ebd69d5ac3076258f3fcdac6d/go/database/mpt/forest.go#L1048) function to geth the encoder needed for the individual stocks. I think this would also make the `open` function itself more readable.

go/database/mpt/io/parallel_visit_test.go
```diff
@@ -219,45 +219,68 @@ func TestVisit_Nodes_CannotOpenFiles(t *testing.T) {
...
+			trie := createArchive(t, dir)
+			root, err := trie.GetBlockRoot(0)
+			if err != nil {
+				t.Fatalf("failed to get block root: %v", err)
+			}
+			if err := trie.Close(); err != nil {
+				t.Fatalf("failed to close archive: %v", err)
+			}
+			return root
+		}},
```
HerbertJordan, line 237: nit: please try to keep the line-width low (e.g. below 120 characters) to facilitate reviews also on laptops if needed. Here it should be doable by doing a line break for each parameter of the struct and the function bodies.

```suggestion
		{
			"hashWithNodes",
			func(dir string) nodeSourceFactory {
				return &nodeSourceHashWithNodesFactory{directory: dir}
			},
			func(t *testing.T, dir string) (root mpt.NodeId) {
				trie := createArchive(t, dir)
				root, err := trie.GetBlockRoot(0)
				if err != nil {
					t.Fatalf("failed to get block root: %v", err)
				}
				if err := trie.Close(); err != nil {
					t.Fatalf("failed to close archive: %v", err)
				}
				return root
			},
		},
```

go/database/mpt/io/parallel_visit_test.go
```diff
@@ -219,45 +219,68 @@ func TestVisit_Nodes_CannotOpenFiles(t *testing.T) {
...
-	trie := createArchive(t)
-
-	nodeId, err := trie.GetBlockRoot(0)
-	if err != nil {
-		t.Fatalf("failed to get block root: %v", err)
-	}
-	if err := trie.Close(); err != nil {
-		t.Fatalf("failed to close archive: %v", err)
+	sourceFactories := []struct {
+		name       string
```
HerbertJordan, line 223: nit: when your test cases have a `name` a nice trick is also to use a `map[string]struct{ ... }` and use the `name` as a key instead of a field. This gives a nice syntax for enumerating cases (`for name, test := range tests {`) and avoid name collisions (probably not that much of a problem).

HerbertJordan (changes requested):
Looks good, however, I would favor a single `nodeSourceFactory` with an `MptConfig` parameter over two distinct types for the two use cases. This should be less code and it scales better if we would go for ever supporting S4 or other configurations one day.

go/database/mpt/io/parallel_visit.go
kjezek: ah, good idea!

go/database/mpt/io/parallel_visit_test.go
kjezek: sorry, it must have been some auto-formating which I have missed.

go/database/mpt/io/parallel_visit_test.go
kjezek: updated

go/database/mpt/config.go
```diff
@@ -84,6 +86,17 @@ func GetConfigByName(name string) (MptConfig, bool) {
...
 
+// GetEncoders returns the value encoders for the given MPT configuration.
+func (c MptConfig) GetEncoders() (
+	stock.ValueEncoder[AccountNode],
+	stock.ValueEncoder[BranchNode],
+	stock.ValueEncoder[ExtensionNode],
+	stock.ValueEncoder[ValueNode],
+) {
+
+	return getEncoder(c)
```
HerbertJordan, line 97: nit: In this case I think we should replace `getEncoder(c)` with `c.GetEncoders()` everywhere ... wdyt?

HerbertJordan (approved):
Looks good, thanks for parameterizing the source factory. I like that the "visitAll" now gets a configuration as a parameter.

I did not go through all the restructured tests ... I trust that those have been correctly updated. It is unfortunately always difficult to check for changes if the indentation of code changes ...

go/database/mpt/config.go
kjezek: It could be, yes, I will update in a follow up

PR #1027, opened 2024-09-17
makes use of GetEncoders, public function of config

kjezek:
This PR updates the code to use the public function of the config, GetEncoders, at every place where original getEncoder() was use.  This is a code clean-up.


HerbertJordan (approved):
Thanks for this code quality improvement!

PR #1029, opened 2024-09-18
fixes computation of roots slice capacity

kjezek:
This PR fixes a wrong calculation of capacity of the roots slice. It caused a high memory consumption as an unnecessarily big slice was pre-allocated.


HerbertJordan (approved):
Thanks for fixing this!

PR #1030, opened 2024-09-18
Add throttling to node prefetchers

HerbertJordan:
This PR introduces two improvements to the parallel visit code:

- **throttling** of node pre-fetchers: workers stop prefetching nodes beyond a configurable threshold of prefetched nodes; however, since at the time workers are throttled not all needed nodes may have been fetched, the main thread is extended to contribute to the prefetching as needed, this puts a constant upper limit on the memory usage of the `visitAll` operation;

- **lockstep prefetching** of workers: workers will synchronize their activities to avoid prefetching large shares of far-future nodes and thereby exceeding the throttling limit for a long time. By synchronizing prefetch workers to load nodes in blocks, the focus of what is to be prefetched is forced to remain on the near future window.

For the latter, a `barrier` utility was introduced facilitating the synchronization of workers.

- [x] complete a full export run for block 88M

### DIFF
--- a/go/database/mpt/io/parallel_visit_test.go
+++ b/go/database/mpt/io/parallel_visit_test.go
@@ -166,6 +166,11 @@ func TestVisit_CanHandleSlowConsumer(t *testing.T) {
 	if err != nil {
 		t.Fatalf("failed to open live db: %v", err)
 	}
+	defer func() {
+		if err := live.Close(); err != nil {
+			t.Fatalf("failed to close live db: %v", err)
+		}
+	}()
 
 	addr := common.Address{}
 	err = errors.Join(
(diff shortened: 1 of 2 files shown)


go/database/mpt/io/parallel_visit_test.go
```diff
@@ -154,6 +158,94 @@ func TestNodeSource_CanRead_Nodes(t *testing.T) {
 	}
 }
 
+func TestVisit_CanHandleSlowConsumer(t *testing.T) {
+	// Create a reasonable large trie.
+	config := mpt.S5LiveConfig
+	dir := t.TempDir()
+	live, err := mpt.OpenGoFileState(dir, config, mpt.NodeCacheConfig{Capacity: 1024})
```
kjezek, line 165: closing the trie is missing

go/database/mpt/io/parallel_visit_test.go
```diff
@@ -154,6 +158,94 @@ func TestNodeSource_CanRead_Nodes(t *testing.T) {
...
+		t.Fatalf("failed to visit trie: %v", err)
+	}
+
+	root := live.GetRootId()
+	if err := live.Close(); err != nil {
+		t.Fatalf("failed to close live db: %v", err)
+	}
+
+	// This visitor is stalling from time to time providing the pre-fetcher
+	// workers from rushing ahead and filling up the prefetch buffer.
```
kjezek, line 209:
```suggestion
	// workers room to rush ahead and filling up the prefetch buffer.
```

kjezek (approved):
nice idea of how to enable the throttling, thanks

go/database/mpt/io/parallel_visit_test.go
HerbertJordan: It was closed in line 204 but you are right, it should be closed in a `defer` statement just in case the test fails. Also, it is better aligned with the rest of our test codes.

kjezek (approved):

PR #1031, opened 2024-09-20
Fix memory leaking tests.

cabrador:
This PR fixes leaking tests in `state` package.
`s.Flush()` sometimes failed (mainly for `sql` impelementations) which caused `live.Close()` not being called - this caused that the `LruCache` remained in the memory.

Here is the result:

Both experiments ran with `diagnostic_tool` from `mpt` package and this test to deadlock the tests after they are complete.
```
func TestZZZZ(t *testing.T)  {
    time.Sleep(1 * time.Hour)
    
}
```

It is clear that the cache prevailed even after the tests were closed.

Before:
Screenshot 2024-09-23 at 17 42 21
Screenshot 2024-09-23 at 18 09 27

After:
Screenshot 2024-09-23 at 17 14 46
Screenshot 2024-09-23 at 17 12 08

There is still tiny bit of memory leak which prevails even after fixes for which I cannot find the root cause.

This PR also reduces the size of `LruCache` in tests from 100 MiB to 10 MiB which cuts the total memory usage approximately by 2/3.

### DIFF
--- a/go/state/state_db.go
+++ b/go/state/state_db.go
@@ -14,12 +14,11 @@ package state
 import (
 	"errors"
 	"fmt"
+	"github.com/Fantom-foundation/Carmen/go/common/witness"
 	"maps"
 	"math/big"
 	"sync"
 	"unsafe"
-
-	"github.com/Fantom-foundation/Carmen/go/common/witness"
 
 	"github.com/Fantom-foundation/Carmen/go/common"
 	"github.com/Fantom-foundation/Carmen/go/common/amount"
--- a/go/state/state_db.go
+++ b/go/state/state_db.go
@@ -14,11 +14,12 @@ package state
 import (
 	"errors"
 	"fmt"
-	"github.com/Fantom-foundation/Carmen/go/common/witness"
 	"maps"
 	"math/big"
 	"sync"
 	"unsafe"
+
+	"github.com/Fantom-foundation/Carmen/go/common/witness"
 
 	"github.com/Fantom-foundation/Carmen/go/common"
 	"github.com/Fantom-foundation/Carmen/go/common/amount"
--- a/go/state/state_db_and_state_test.go
+++ b/go/state/state_db_and_state_test.go
@@ -226,7 +226,7 @@ func testCarmenStateDbHashAfterModification(t *testing.T, mod func(s state.State
 		if err != nil {
 			t.Fatalf("failed to create reference state: %v", err)
 		}
-		ref := state.CreateStateDBUsing(ref_state)
+		ref := state.CreateCustomStateDBUsing(ref_state, 100000)
 		defer ref.Close()
 		mod(ref)
 		ref.EndTransaction()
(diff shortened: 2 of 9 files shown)


go/state/state_db_test.go
HerbertJordan, line 1: All tests in this file target the `StateDB` instance that manages transaction scopes and sits between the VM and the actual database. To be database-independent, all those tests use (deliberately) a mocked version of a database, that can not leak memory or resources. Thus, I think it is unnecessary clutter to add close calls to all those mock-based `StateDB` instances, which just leads to the need of additional `Flush` and `Close` expectations.

Unless there is evidence that the previous version actually caused resource issues, I would argue that this is adding complexity and clutter that should be avoided.

HerbertJordan (changes requested):
Thanks for cleaning up the `state` package tests. 

While I appreciate the effort, I think that in particular for the `StateDB` tests (`state_db_test.go`) that are all based on mocks of the underlying state DB, the extra boiler plate code is not warranted. It  adds extra clutter distracting from the core operations under test. Unless this is fixing an actual resource issue, I would favor the previous code.

Have you identified the memory issues using heap-dumps like in the MPT package? Is this PR eliminating all of them, or is more work required in the `state` package?

kjezek (changes requested):
I would give about the same feedback as @HerbertJordan 
I do not see how this PR can "fix memory leaking tests"?

I did some measurement, and it shows that majority of memory is consumed by` LruCache` used in schemas 1-3. I.e. it seems that this memory is not released when closing a state. This should be targeted mainly. 

pprof002

go/state/state_db.go
```diff
@@ -14,12 +14,13 @@ package state
...
 	"errors"
 	"fmt"
-	"github.com/Fantom-foundation/Carmen/go/common/witness"
 	"maps"
 	"math/big"
 	"sync"
 	"unsafe"
 
+	"github.com/Fantom-foundation/Carmen/go/common/witness"
+
```
kjezek, line 23: this change is unrelated, please revert.

I understand that it was done by IDE, and for this reason we accept such changes when there are real functional changes in the file. But in this case nothing has changed

go/state/state_db_and_state_test.go
```diff
@@ -584,7 +584,9 @@ func TestStateDBSupportsConcurrentAccesses(t *testing.T) {
 			if err != nil {
 				t.Fatalf("failed to initialize state %s; %s", config.name(), err)
 			}
-			defer s.Close()
+			defer func() {
```
kjezek, line 587: this update changes nothing

kjezek (dismissed):
Thanks for fixing this. Can you give an example why `Flush` was failing?

Also I wonder if this change fixes all leaks related to `LruCache`.  Most notably, the chart I provided, and the chart you provided, look quite different. In particular, in my chart, there leaked 9GB and was many more `LruCache` instances, from many more tests.

cabrador:
Moved to #1034

PR #1032, opened 2024-09-23
sends only the number of responses coming from workers

kjezek:
This PR resolves failures of tests, where the number of data produced by async workers in parallel trie visit is measured. 

The original code measured the total number of responses in the queue. It could easily exceed the expected limit, and the test failed. 

It seems that it was not originally intended by this test.  The test was supposed to measure only the volume of data produced by the workers, not to exceed memory limits by threads possibly rushing ahead in computation. 

During an execution, it can happen that all workers correctly stop producing data realising that too many responses were  produced. On the other hand, not to stall the execution, the main thread continues and still produces data, which piles up to the already existing number of responses. This is, correct behaviour as the execution cannot stop and the main thread must produce more data until it collects the data it needs.  It can lead to a situation, where the number of responses exceeds the tested limit. 

This PR added a `bool`  flag marking each response as either coming from a worker, or from the main thread.  It is then validated that only the amount of responses coming from workers is not exceeded.  The number of responses generated by the main thread is kept unbound (unchecked).

### DIFF
--- a/go/database/mpt/io/parallel_visit.go
+++ b/go/database/mpt/io/parallel_visit.go
@@ -99,6 +99,7 @@ func visitAllWithConfig(
 	type response struct {
 		node mpt.Node
 		err  error
+		main bool // response comes from the main thread
 	}
 	responses := map[mpt.NodeId]response{}
 	responsesMutex := sync.Mutex{}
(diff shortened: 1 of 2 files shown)


HerbertJordan (approved):
Thanks for fixing the test!

PR #1033, opened 2024-09-24
Add stringer to GetMemoryFootprint

cabrador:
This PR adds `fmt.Stringer` which explicitly forces `String()` method to be implemented by `MemoryFootprint`.
This is necessary because `carmen.MemoryFootprint` needs to be passed as a `fmt.Stringer`

### DIFF
--- a/go/carmen/memory_footprint.go
+++ b/go/carmen/memory_footprint.go
@@ -10,9 +10,14 @@
 
 package carmen
 
-import "github.com/Fantom-foundation/Carmen/go/common"
+import (
+	"fmt"
+
+	"github.com/Fantom-foundation/Carmen/go/common"
+)
 
 type MemoryFootprint interface {
+	fmt.Stringer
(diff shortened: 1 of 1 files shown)


kjezek (approved):
Can you please better explain use case for this change?

cabrador:
> Can you please better explain use case for this change?

@wsodsong explicitly asked for adding `String()` method to `carmen.MemoryFootprint`. Although this method is already implemented, adding `fmt.Stringer` to the interface is necessary because it actually is not a `fmt.Stringer`.

Up until now the implementation of `String()` was optional. Now its mandatory.
Error from Aida explaining this change is necessary:
```
state/carmen.go:445:45: cannot use usage (variable of type carmen.MemoryFootprint) as fmt.Stringer value in struct literal: carmen.MemoryFootprint does not implement fmt.Stringer (missing method String)
```

PR #1034, opened 2024-09-24
clear caches when closing resources to release memory

kjezek:
This PR releases items stored in `LruCache` when components using this cache are `Closed`. This reduces memory in tests of state package.  Heap dumps below shows the memory allocated before and after this change. The dumps were captured right after all tests passed. 

Before:
state-mem-heap-before

After:
state-mem-profile

### DIFF
--- a/go/backend/store/cache/cachedstore.go
+++ b/go/backend/store/cache/cachedstore.go
@@ -11,6 +11,7 @@
 package cache
 
 import (
+	"errors"
 	"github.com/Fantom-foundation/Carmen/go/backend"
 	"github.com/Fantom-foundation/Carmen/go/backend/store"
 	"github.com/Fantom-foundation/Carmen/go/common"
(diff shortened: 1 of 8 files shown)


cabrador (approved):

go/backend/store/cache/cachedstore.go
```diff
@@ -75,10 +76,8 @@ func (m *Store[I, V]) Flush() error {
 }
 
 func (m *Store[I, V]) Close() error {
-	if err := m.Flush(); err != nil {
-		return err
-	}
-	return m.store.Close()
+	m.cache.Clear()
+	return errors.Join(m.Flush(), m.store.Close())
```
cabrador, line 80: Shouldnt we flush before clearing the cache?

cabrador (approved):

go/backend/store/cache/cachedstore.go
kjezek: in this case, flush does not use the cache. But you are right, considering this as a black-box, it is correct to flush first. Fixed.

cabrador (approved):

Issue #981, opened 2024-07-31
The Test TestNodeFlusher_TriggersFlushesPeriodically is flaky


HerbertJordan:
The test [TestNodeFlusher_TriggersFlushesPeriodically](https://github.com/Fantom-foundation/Carmen/blob/81e342a258101a641c66783deb657c3cf62965bc/go/database/mpt/node_flusher_test.go#L46) is flaky, sometimes failing the nightly race-detection tests (example [here](https://scala.fantom.network/job/Carmen/job/Race-Detection/123/)).

kjezek:
fixed by #1035

PR #1035, opened 2024-09-25
introduces mockable ticker

kjezek:
This PR introduces abstraction of `time.Ticker` which may be mocked. 

This new type is used in the `nodeFlusher`, and fixes test #981 .


go/common/ticker/ticker.go
```diff
@@ -0,0 +1,51 @@
...
+//go:generate mockgen -source ticker.go -destination ticker_mocks.go -package ticker
+
+// Ticker is an abstraction of a ticker from standard time package.
+// It contains a channel which produces ticks at certain intervals
+// defined by implementations. For example, TimeTicker wraps
+// the standard time.Ticker based on time.Duration intervals.
+// Furthermore, the ticker includes the method Stop to terminate
+// the ticker. When the ticker is stopped, no more ticks will be
+// sent via the channel.
+type Ticker interface {
```
HerbertJordan, line 24: nit: you should add to the documentation that all tickers need to be stopped -- only since `time.Ticker` Go 1.23 this is no longer needed; but I think we might not have set this yet as our minimum Go requirement.

go/database/mpt/node_flusher_test.go
```diff
@@ -44,32 +45,43 @@ func TestNodeFlusher_StartAndStopWithDisabledFlusher(t *testing.T) {
 }
 
 func TestNodeFlusher_TriggersFlushesPeriodically(t *testing.T) {
```
HerbertJordan, line 47: Given the new test code, the test should probably be renamed into something like
`TestNodeFlusher_UsesTickerToTriggerFlushes`

HerbertJordan (approved):
Nice.

PR #1042, opened 2024-10-16
exposes method to assess empty storage of an account into State

kjezek:
There has been already a method to assess if a storage root of an account is empty or not.  This PR exposes this method to `State` and `StateDB`.

### DIFF
--- a/go/state/gostate/go_state_test.go
+++ b/go/state/gostate/go_state_test.go
@@ -495,6 +495,24 @@ func TestGetMemoryFootprint(t *testing.T) {
 	}
 }
 
+func TestGoState_HasEmptyStorage(t *testing.T) {
+	addr := common.Address{0x1}
+	ctrl := gomock.NewController(t)
+	live := state.NewMockLiveDB(ctrl)
+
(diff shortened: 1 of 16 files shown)


go/state/gostate/go_state_test.go
```diff
@@ -773,6 +792,10 @@ func TestState_All_Live_Operations_May_Cause_Failure(t *testing.T) {
 				if _, err := db.GetHash(); shouldFail && !errors.Is(err, injectedErr) {
 					t.Errorf("operation should fail")
 				}
+				shouldFail = shouldFail || errors.Is(results[7], injectedErr)
```
HerbertJordan, line 795: If I read this test setup correctly, this one should use `results[8]` and all subsequent ones should have the index increased by 1. At least in the expectation, the return value set is `results[8]`.

HerbertJordan (approved):

go/state/gostate/go_state_test.go
kjezek: you are right, thanks! This test should be probably improved to initialise data in a loop

HerbertJordan (approved):

cabrador (approved):

PR #1049, opened 2024-11-19
reflects that accounts may be desctructed in block when assessing storage

kjezek:
This PR fixes error in assessing if the storage is empty in cases where an account was desctructed in the same block.

To check if the method is empty, method `HasEmptyStorage` is called.  The result of this method is known only for sealed block. It may happend though that an acocunt is desctructed and re-created in the same block. In this case, the method must correctly reflect intermediate state of the account. 

Previous version caused state validation error for Ethereum mainet data. Somwhat surrprisingly such an error did not happen for Opera mainet data.

### DIFF
--- a/go/state/state_db_test.go
+++ b/go/state/state_db_test.go
@@ -4372,7 +4372,7 @@ func TestStateDB_HasEmptyStorage(t *testing.T) {
 	}
 }
 
-func TestStateDB_HasEmptyStorage_ReportEmpty_Destructed_Contract(t *testing.T) {
+func TestStateDB_HasEmptyStorage_DestructedContract_ReportEmpty(t *testing.T) {
 	addr := common.Address{0x1}
 	ctrl := gomock.NewController(t)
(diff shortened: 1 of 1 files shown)


go/state/state_db_test.go
```diff
@@ -4372,6 +4372,33 @@ func TestStateDB_HasEmptyStorage(t *testing.T) {
 	}
 }
 
+func TestStateDB_HasEmptyStorage_ReportEmpty_Destructed_Contract(t *testing.T) {
```
HerbertJordan, line 4375:
```suggestion
func TestStateDB_HasEmptyStorage_DestructedContract_ReportEmpty(t *testing.T) {
```

HerbertJordan (dismissed):
Thanks for digging into this and identifying the cause of the problem.

Before we can merge this and updated it in Sonic we need to run all tests we have to make sure it is not introducing new issues.

HerbertJordan (approved):

PR #1052, opened 2024-11-25
Lower export LiveDB cache capacity

cabrador:
This PR changes the cache capacity for import to 200k and export to 2k.

This is necessary due to high memory consumption in Sonic integration tests. It lowers it by ~2 GB.

Measurements done on 20mil database:
Current main took `26:21`
This branch took `25:36`


kjezek:
Can you elabarote more on which use-cases did you test?

kjezek (approved):
looks good, thanks

PR #1054, opened 2024-11-28
Witness proof returns empty RLP node encoding for empty hash root

kjezek:
This PR updates witness proof extraction to return a single-item proof with an RLP encoding of an empty node when the input query is for an empty node hash. 

This can happen in two cases: 
* query to an empty MPT (cannot happen for non-empty database)
* query to an empty sub-tree (practical case for empty storage of an account)

Furthermore, this PR adds tests checking this invariant. 

The empty node itself is not usually stored in the witness proof - i.e. when one obtains RLP encoded nodes from the proof, there will not be an empty node within the returned elements since all the time when there is a reference to a child, which is empty, this reference will be a hash on an RLP encoded empty node.  It would be redundant to store the empty node itself in otherwise a non-empty proof. 

The only case the empty node is present in the proof is when the input already refers to an empty trie (sub-trie), and in this case the proof must contain only one item, the empty node


go/database/mpt/proof_test.go
```diff
@@ -175,6 +175,101 @@ func TestCreateWitnessProof_CanCreateProof(t *testing.T) {
...
+						storage: &Empty{}},
+				},
+			},
+		},
+		"path incomplete - extension": {
+			desc: &Branch{
+				children: Children{
+					addressNibbles[0]: &Extension{}},
+			},
+		},
```
HerbertJordan, line 217: Both of these are invalid MPTs -- there must not be a Child node with only one successor, and every extension must have a non-empty next.

go/database/mpt/proof_test.go
```diff
@@ -175,6 +175,101 @@ func TestCreateWitnessProof_CanCreateProof(t *testing.T) {
...
+				if got, want := n.hashes[nibble], EmptyNodeEthereumHash; got != want {
+					t.Errorf("unexpected hash: got %v, want %v (empty RLP node)", got, want)
+				}
+			case *decodedAccountNode:
+				if got, want := n.storageHash, EmptyNodeEthereumHash; got != want {
+					t.Errorf("unexpected storage hash: got %v, want %v (empty RLP node)", got, want)
+				}
+			default:
+				t.Errorf("unexpected terminal node: %T", n)
+			}
```
HerbertJordan, line 268: This is less testing anything about the visitor but more that the test infrastructure as assembled the nodes correctly ...

go/database/mpt/proof_test.go
```diff
@@ -175,6 +175,101 @@ func TestCreateWitnessProof_CanCreateProof(t *testing.T) {
 	}
 }
 
+func TestCreateWitnessProof_ProofContainsEmptyNodes(t *testing.T) {
```
HerbertJordan, line 178: I am not sure what this test is supposed to test. It does not check that proofs contain empty nodes (which it should not) nor does it check that it does not contain empty nodes.

HerbertJordan (changes requested):

go/database/mpt/proof_test.go
kjezek: ok, I removed this test completely, when two out of three tests mimics an unrealistic MPT, it is not worth fixing it. Test of empty storage and empty trie is covered by other tests.

The other new test was extended to cover also a non-existing account - it is not passing at the moment, but will get fixed by the other PR - while the fix is obvious, I did not do it to prevent merge conflicts.

HerbertJordan (approved):

HerbertJordan (approved):

PR #1055, opened 2024-11-29
List witness proof elements in order

HerbertJordan:
This PR applies the following changes to the Witness proof interface:
- the function `GetStorageElements` can only be called with exactly one `key` anymore; except in one test, the multi-key version was never used, and by restricting it to one key element, the following modification is possible
- the elements returned by `GetStorageElements` are now listed in navigation order, from the storage root to the terminal node that certifies the stored value
- the function `GetStorageElements` no longer returns the storage root hash
- an additional `GetAccountElements` function is added, allowing to extract the proof steps from the state-root to the terminal certifying the account state in-order; this function now returns the storage root hash

These modifications have been made to facilitate the extraction of proofs listing proof steps in the order as they are needed by the `eth_getProof` RPC method.


go/carmen/proof.go
```diff
@@ -48,15 +48,18 @@ type WitnessProof interface {
 	// GetElements returns serialised elements of the witness proof.
 	GetElements() []Bytes
 
-	// GetStorageElements returns serialised elements of the witness proof for a given account
-	// and selected storage locations from this proof.
+	// GetAccountElements returns serialised elements of the witness proof for a selected account.
+	GetAccountElements(root Hash, address Address) ([]Bytes, bool)
```
kjezek, line 52: would be probably more intuitive to return `storageRoot` in this method, and not in the `GetStorageElememts` method - but it can be done as a further improvement later.  Maybe put in a TODO

kjezek (approved):
it looks good, thanks

go/carmen/proof.go
HerbertJordan: Would yo get this done by tomorrow or should I do it now?

HerbertJordan: I fixed it right now.

kjezek (approved):
looks good, thanks

PR #1056, opened 2024-12-02
tests path navigation order in witness proof

kjezek:
This PR extends tests of WitnessProof methods `GetAccountElements` and `GetStorageElements`;
it tests that nodes returned from this methods are ordered according to navigation path in the MPT.

Furthermore, this PR extends code coverage, except two unimplemented methods, the code coverage is 100%.

### DIFF
--- a/go/database/mpt/proof_test.go
+++ b/go/database/mpt/proof_test.go
@@ -962,14 +962,13 @@ func TestCreateWitnessProof_Elements_Path_Sorted_By_Trie_Navigation(t *testing.T
 
 	// check that the nodes are in the correct order, starting with account nodes, followed by storage nodes
 	proofNodes := append(accountElements, storageElements...)
+	if got, want := len(proofNodes), len(mptNodes); got != want {
+		t.Fatalf("unexpected number of nodes: got %v, want %v", got, want)
+	}
 	for i, node := range proofNodes {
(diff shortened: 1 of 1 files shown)


go/database/mpt/proof_test.go
```diff
@@ -854,6 +855,126 @@ func TestCreateWitnessProof_GetAccountElements(t *testing.T) {
...
+
+	// check that the nodes are in the correct order, starting with account nodes, followed by storage nodes
+	proofNodes := append(accountElements, storageElements...)
+	for i, node := range proofNodes {
+		if got, want := node, mptNodes[i]; !bytes.Equal(got.ToBytes(), want.ToBytes()) {
+			t.Errorf("unexpected node: got %v, want %v", got, want)
+		}
+	}
+
+	if got, want := len(proofNodes), len(mptNodes); got != want {
```
HerbertJordan, line 971: nit: this length check should be done before accessing `mptNodes[i]` in line 966 to avoid a crash in case the length of the lists do not match.

HerbertJordan (approved):
Looks good, thanks!

go/database/mpt/proof_test.go
kjezek: right, updated

HerbertJordan (approved):

